PR #648, opened 2015-08-04
When not using AWS default instance the script release-ec2.py will fail

PerGon:
Need to specify in the AWS commands what "--region" to use, otherwise it will use the default one (us-east-1). And if the user had choose a different one ( `./scripts/release-ec2.sh (...) --region eu-west-1` ) the script will fail


wkozaczuk:
Again we do not accept pull requests. Plus I do not think we maintain those EC2 scripts and we have rewritten some of the since. So I am closing it.

PR #674, opened 2015-11-03
Update the Documentation and Licenses to Modern Eras

nhubbard:


nyh:
Hi, moving my review to the mailing list, so everyone case see it (that's
the problem with github's pull request - it makes reviews invisible to the
rest of the crowd).

I don't think we should modify the documentation/LICENSE-* files, as these
are supposed to be verbatim copies of the licenses of those different
projects.

If anything, I would move these documentation/LICENSE-* files to licences/
(where we already have some of them).

The other changes, of documentation/*.txt into *.md, I'm good with. Sad we
don't have more meaningful documentation there - this is like a tiny random
subset of what OSv offers, written almost 3 years ago and I'm not even sure
is still relevant....


--
Nadav Har'El
nyh@scylladb.com

nhubbard:
Sounds good. Should I revert the licenses back to plain text?

PR #682, opened 2015-11-20
Added Fedora 23 dist to the setup.py script

sandlbn:


nyh:
Please send this patch to the mailing list, so it can be reviewed.

In particular, I have a question - what is this "--best", why is it needed
to add it in Fedora 23, and whether the option existed in earlier
incarnations of yum (nominally, we also support Fedora 19, although I doubt
anyone is actually testing that).


--
Nadav Har'El
nyh@scylladb.com

On Fri, Nov 20, 2015 at 1:30 PM, Marcin Spoczynski <notifications@github.com

nyh:
We've done this long ago. Closing.

PR #727, opened 2016-02-14
Bigfile

Meeuwisse:
When uploading a file larger than the image, the file name would silently corrupt. Attempting to upload without any disk space left would crash the httpserver. These patches cause an exception to be thrown for the former, and exceptions in general to be caught & returned to the user, including message parsing errors.


nyh:
I've committed patches from the mailing list (note that not all of them), so closing this "pull request".

PR #736, opened 2016-04-04
[Patch] Check only for Ubuntu Major Releases

VincentS:
I changed the setup.py script to check only for Ubuntu major releases (13,14,15).
Before it I had to adjust the setup.py for every minor release of Ubuntu where it is unlikely that something breaks / is not compatible.


nyh:
Thanks for the patches. Can you please send them (as two separate patches) to the OSv mailing list, for review, instead of using a pull request? Thanks!

VincentS:
Sure, will do it today :)

Issue #742, opened 2016-05-05
One more package need to be install on ubuntu 14.04


guyj93:
I followed the steps in README.md to build osv image on ubuntu 14.04.
But at the step of 
```
scripts/build
```
I got this error:
```
global_server.cc:26:27: fatal error: yaml-cpp/yaml.h: No such file or directory
 #include "yaml-cpp/yaml.h"
                           ^
compilation terminated.
make: *** [obj/global_server.o] Error 1
Traceback (most recent call last):
  File "scripts/module.py", line 243, in <module>
    args.func(args)
  File "scripts/module.py", line 200, in build
    make_modules(modules, args)
  File "scripts/module.py", line 104, in make_modules
    raise Exception('make failed for ' + module.name)
Exception: make failed for httpserver
scripts/build failed: jdkbase=$jdkbase ARCH=$arch mode=$mode OSV_BASE=$SRC OSV_BUILD_PATH=$OSV_BUILD_PATH scripts/module.py $j_arg build -c $modules
```
It seems the package libyaml-cpp-dev is needed.
After I run
```
sudo apt-get install libyaml-cpp-dev
```
the build script works.

I think libyaml-cpp-dev may be added to scripts/setup.py.

bamboojsr:
Only need to  update README.md

PR #744, opened 2016-05-12
fix the package name of libyaml on Ubuntu

guyj93:
Fix on [issue#742](https://github.com/cloudius-systems/osv/issues/742)
There isn't a package names "libyaml-cpp" in Ubuntu official apt-get repo.
The correct name of the required package is "libyaml-cpp-dev".


nyh:
I've committed this patch, thanks. Next time, please send a patch to the mailing list instead of using pull requests. Thanks.

PR #785, opened 2016-08-10
Add Ubuntu 16.04 support in scripts/setup.py

miha-plesko:
Was unable to build OSv on my 16.04 just because
setup.py said "Your distribution is not supported by this script."
So I've copied configuration from 15.04. Also had to add apt repo
for openjdk-7-jdk to become apt-get installable.

### DIFF
--- a/scripts/setup.py
+++ b/scripts/setup.py
@@ -39,6 +39,7 @@ class Fedora(object):
         packages = ['java-1.7.0-openjdk', 'python-requests']
         ec2_packages = []
         test_packages = []
+        preinstall = []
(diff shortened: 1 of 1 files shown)


nyh:
Can you please send this patch to the mailing list instead of doing a pull request? It will be easier to review there.  I'm closing this pull-request.
I already have a question: Why do we need openjdk-7 specifically? In Fedora, we settle well with openjdk 8 (which is the openjdk package in recent distributions). Why do we need to insist on openjdk 7 for Ubuntu?

Issue #792, opened 2016-08-29
netperf benchmark on EXSI hypervisor


ufo19890607:
Hi, 
I ran the netperf app on Exsi hypervisor, I followed the tutor and converted the image to vmdk files(scripts/convert vmdk). However, when I run the netperf apps, the client show the fault messages.
[backtrace]
0x000000000000002257f8 <__assert_fail+24>
0x00000000000000360402 <VMW::VMXnet3_rxqueue::receive()+1042>
0x000000000000003605b1 <VMW::VMXnet3_rxqueue::receive_work()+353>
0x0000000000000036284b <std::_Function_handler<void (), VMW;;VMXnet3_rxqueue::VMXnet
3_rxqueue()::{laMbda()#1}>::_M_invoke(std::_Any_data const&)+11>
0x000000000000003e1f86 <thread_main_c+38>
0x00000000000000388492 <???+3703954>

Is it a bug?

DerangedMonkeyNinja:
Probably.  The vmxnet3 driver appears to have state issues.  The change in #793 works for me.

wkozaczuk:
Can we check if this is fixed by #793?

nyh:
I just realized that we never committed this patch! I don't know why that pull request was ever closed.
The patch is sitting on the mailing list, titled "drivers/vmxnet3: Fix driver state issues".
I don't know how to test this patch myself (I don't have vmware)... But I believe @DerangedMonkeyNinja tested it well so I think I should commit it. I have no idea if it fixes this bug, though, nor how to reproduce it.

wkozaczuk:
I have access to VMware so I will try to test it.

PR #793, opened 2016-08-30
drivers/vmxnet3.cc: Fix driver state issues

DerangedMonkeyNinja:
* Handle both missing SOP and EOP descriptors in the rx loop.
* Always leave at least one Tx descriptor available.  The driver
  appears to hang when it runs out of them.


DerangedMonkeyNinja:
Possible fix for #792.

benoit-canet:
Annoying pre review disclaimer:

Could you please follow:
https://github.com/cloudius-systems/osv/blob/master/CONTRIBUTING

we don't take git pr.

PR #825, opened 2016-12-27
update for Ubuntu 16.10, changed openjdk 7 in 8 version

robo3945:
I updateded the setup.py to work with Ubuntu 16.10. I have changed the version of Java to 8.


nyh:
The problem with this patch is that you changed the version of Java also for older versions of Ubuntu, which may not have Java 8. You can look at the Fedora stuff in setup.py to see how we solved exactly the same problem for Fedora.
Also, please send your patch (with git send-email) to the OSv mailing list instead of using a pull request. It better encourages public review of the patches.

robo3945:
ok thanks

PR #1002, opened 2018-09-27
Add Ubuntu Bionic(18.04)

pshem:
Based on Mint 19, simplifies setup on Ubuntu 18.04


wkozaczuk:
Unfortunately we do not accept pull requests. Instead we ask for git patches as described here - https://github.com/cloudius-systems/osv/wiki/Formatting-and-sending-patches

pshem:
Took me a while due to other commitments, but here you are: https://groups.google.com/forum/#!topic/osv-dev/BemDWfPPK3g

PR #1214, opened 2023-01-20
Fix pointer cast

gohar94:
On `g++` 11.3.0 (Ubuntu), the following error was being thrown:
`error: pointer of type ‘void *’ used in arithmetic`


wkozaczuk:
Hi,

Could you please provide the full build command which fails? I can run `./scripts/build -j$(nproc) image=native-example` on Ubuntu with GCC 11.3.0 and I do not get this error. The OSv makefile explicitly adds `-Wno-pointer-arith` to make a pointer to `void` and treat it as a pointer to `char`.

PR #1217, opened 2023-02-08
Disable BARs decoding when computing size

FedeParola:
When computing the size of a BAR the device should not try to decode the content of the BAR as an address. For example, this can cause a crash in QEMU trying to map memory at a nonexistent address.


wkozaczuk:
It looks good and the unit tests pass so I am merging it.

PR #1219, opened 2023-03-15
pci: avoid execution of msix_enable() function if already enabled

FedeParola:
When using the `interrupt_manager::easy_register()` function multiple times on a device (e.g., to register interrupts for multiple queue pairs in a virtio-net device), `msi(x)_enable()` is called multiple times, causing interrupts that have previously been registered to be masked and never unmasked. Avoid this problem by preventing  the execution of `msi[x]_enable()` if msi[x] is already enabled.

### DIFF
--- a/drivers/pci-function.cc
+++ b/drivers/pci-function.cc
@@ -680,7 +680,7 @@ namespace pci {
 
     void function::msix_enable()
     {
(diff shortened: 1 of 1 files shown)


wkozaczuk:
Good catch! It feels like a very important bug to fix.

PR #1220, opened 2023-03-16
rofs/rofs_cache: use both inode_no and sb as key in the cache hashmap

valighita:
The ROFS cache map is global and not unique per mount and the entries are currently indexed only by inode_no.  If there are multiple ROFS mounts, reading from the cache will produce wrong results when there are files having the same `inode_no` in different mounts.

To address this issue, use both the `inode_no` and the superblock pointer as the key in the cache map.

### DIFF
--- a/fs/rofs/rofs_cache.cc
+++ b/fs/rofs/rofs_cache.cc
@@ -41,6 +41,29 @@ struct file_cache {
 };
 
 //
+// Structure used as a key in the global file cache.
+// The entries must be indexed using both inode_no and sb pointer, because
+// different ROFS mounts can contain files with same inode_no
+struct rofs_cache_key {
+    uint64_t inode_no;
+    struct rofs_super_block *sb;
+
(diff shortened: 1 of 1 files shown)


wkozaczuk:
Good catch! 
I see no issues with your MR and all tests pass (`./scripts/build check`) so I am merging it.

nyh:
@wkozaczuk this commit got the committer name `Commit: GitHub <noreply@github.com>` (you can see this with `git log --pretty=full`). There is no mention of your name, or this PR. I wonder if we can do something better for a more reasonable git history... Or this is unfortunately the best that github knows to do with its "merge" button :-(

wkozaczuk:
It is unfortunate. I have merged the last two PRs with "merge + squash".
But I looked at the other one where I did not do squash and the same
thing: Commit:
GitHub ***@***.***>.

I wonder if there is some setting we can apply that would fix it. Or maybe
something has to specifically toggled when applying the "Merge".

Issue #386, opened 2014-07-14
Support some POSIX-RT APIs to run rt-tests (cyclictest) on Osv


hw-claudio:
I have tried building and running cyclictest on OSv:

https://rt.wiki.kernel.org/index.php/Cyclictest

git clone git://git.kernel.org/pub/scm/linux/kernel/git/clrkwllms/rt-tests.git
or tarballs:

https://www.kernel.org/pub/linux/kernel/people/clrkwllms/

there seem to be these missing functions required to be able to do a minimal cyclictest run:

sched_getscheduler
sched_setscheduler

timer_create
timer_settime

do you think it's possible to implement these POSIX rt functions?

Ciao,

Claudio

bamboojsr:
@nyh  @hw-claudio , Today I met same problem, But I had add pthread into OSv lib. 

Adding /libenviron.so...
Adding /zpool.so...
Adding /libzfs.so...
Adding /libuutil.so...
Adding /zfs.so...
Adding /tools/mkfs.so...
Adding /tools/cpiod.so...
Adding /tools/mount-nfs.so...
Adding /tools/umount.so...
Adding /usr/lib/libgcc_s.so.1...
Adding /usr/lib/libpthread.so.0...
Adding /usr/lib/libboost_filesystem.so.1.55.0...
Adding /usr/lib/libboost_regex.so.1.55.0...
Adding /etc/hosts...
Link /etc/mnttab to /proc/mounts ...
Adding /etc/fstab...
Adding /dev...
Adding /proc...
Adding /tmp...
Adding /tools/cyclictest.so...

nyh:
libpthread.so.0 is on the list of objects supplied by the OSv kernel itself (see the full list in program::program in elf.cc), so adding it to the image should not make a difference (did it?), and even if it did, it is quite unlikely that anything from libpthread.so (which is part of glibc) will actually work unmodified on OSv.

If you're interested in adding real-time scheduling to OSv (can you please say something about your use case?) you might be happy to know that we actually have working patches for many (but not all) of these features: I have a patch (which I continue to maintain off-tree) adding OSv-specific methods set_realtime_priority(), and set_realtime_slice(), and Claudio had patches to use those to implement some of the Posix API for scheduling classes.

My patch involves tinkering inside the OSv scheduler - it is not something as simple as strapping on an additional library like libthread.so.

bamboojsr:
[...]

PR #1223, opened 2023-03-29
sched: low-level support for POSIX-compatible real-time scheduling

nyh:
This patch adds "real-time priority" to OSv threads, which can be used to implement Posix's SCHED_FIFO and SCHED_RR scheduling policies.

Refs #386. It doesn't yet "Fixes" it because this patch does not yet add the Posix front-end to this implementation (sched_setscheduler() et al.) - this will be added in a separate patch.

We add to each thread a _realtime._priority, an unsigned integer. Normal (non-real-time) threads have _realtime._priority of 0, and are time-shared fairly as before, while realtime threads have _realtime._priority > 0, and preempt any thread with a lower realtime priority, including all normal threads.

We add a new API, t->set_realtime_priority(priority), to make a thread real-time scheduled. The "priority" is an unsigned int as explained above (0 means normal thread). The resulting scheduling policy is fully compatible with POSIX's SCHED_FIFO. This includes following the intricate rules on what happens to the running thread's place in the queue when there are several real-time threads with the same priority and the running thread yields, waits, or is preempted by a higher-priority thread. This patch also adds an exercise for these cases in tests/misc-scheduler.cc.

This patch does not yet implement time slices as needed for supporting POSIX's SCHED_RR policy. This patch already includes a time-slice setting API, t->set_realtime_time_slice(duration), but if used it produces a warning message, and the scheduler currently ignores this setting. Adding support for real-time time slices is left for a future patch: To implement time slices, we will need to keep in _realtime the total time this thread ran since starting this slice (only zeroed when the thread is removed from the run queue), and each time we switch to a realtime thread with timeslice >= 0, we'll set an appropriate timer (for timeslice minus already used time).

I tested that although this patch adds code into the scheduler it does not noticably slow down context switches of normal threads. E.g., tests/misc-ctxsw.so measured colocated context switch time of 308ns before this patch, and exactly the same after it.

While developing this patch I tested it using added workloads in tests/misc-scheduler.cc, included in this patch, but unfortunately this is not a real regression test - it has no assertions and cannot automatically detect failure (and, like all the "misc-*" tests, it doesn't run automatically in "make check").

### DIFF
--- a/core/sched.cc
+++ b/core/sched.cc
@@ -26,6 +26,7 @@
 #include <osv/preempt-lock.hh>
 #include <osv/app.hh>
 #include <osv/symbols.hh>
+#include <osv/stubbing.hh>
 
 MAKE_SYMBOL(sched::thread::current);
 MAKE_SYMBOL(sched::cpu::current);
@@ -290,7 +291,16 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
 #endif
         } else if (!called_from_yield) {
             auto &t = *runqueue.begin();
-            if (p->_runtime.get_local() < t._runtime.get_local()) {
+            if (p->_realtime._priority > 0) {
+                // Only switch to a higher-priority realtime thread
+                if (t._realtime._priority <= p->_realtime._priority) {
+#ifdef __aarch64__
+                    return switch_data;
+#else
+                    return;
+#endif
+                }
+            } else if (t._realtime._priority == 0 && p->_runtime.get_local() < t._runtime.get_local()) {
                 preemption_timer.cancel();
                 auto delta = p->_runtime.time_until(t._runtime.get_local());
                 if (delta > 0) {
@@ -302,6 +312,20 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
                 return;
 #endif
             }
+        } else { /* called_from_yield */
+            if (p->_realtime._priority > 0) {
+                auto &t = *runqueue.begin();
+                // When yielding, only switch if the next thread has a higher
+                // or same priority as p (i.e., don't switch if t it has a
+                // lesser priority than p).
+                if (t._realtime._priority < p->_realtime._priority) {
+#ifdef __aarch64__
+                    return switch_data;
+#else
+                    return;
+#endif
+                }
+            }
         }
         // If we're here, p no longer has the lowest runtime. Before queuing
         // p, return the runtime it borrowed for hysteresis.
@@ -309,7 +333,10 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         p->_detached_state->st.store(thread::status::queued);
 
         if (!called_from_yield) {
-            enqueue(*p);
+            // POSIX requires that if a real-time thread doesn't yield but
+            // rather is preempted by a higher-priority thread, it be
+            // reinserted into the runqueue first, not last, among its equals.
+            enqueue_first_equal(*p);
         }
 
         trace_sched_preempt();
@@ -350,16 +377,18 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         n->_runtime.add_context_switch_penalty();
     }
     preemption_timer.cancel();
-    if (!called_from_yield) {
-        if (!runqueue.empty()) {
-            auto& t = *runqueue.begin();
-            auto delta = n->_runtime.time_until(t._runtime.get_local());
-            if (delta > 0) {
-                preemption_timer.set_with_irq_disabled(now + delta);
+    if (n->_realtime._priority == 0) {
+        if (!called_from_yield) {
+            if (!runqueue.empty()) {
+                auto& t = *runqueue.begin();
+                auto delta = n->_runtime.time_until(t._runtime.get_local());
+                if (delta > 0) {
+                    preemption_timer.set_with_irq_disabled(now + delta);
+                }
             }
+        } else {
+            preemption_timer.set_with_irq_disabled(now + preempt_after);
         }
-    } else {
-        preemption_timer.set_with_irq_disabled(now + preempt_after);
     }
 
     if (app_thread.load(std::memory_order_relaxed) != n->_app) { // don't write into a cache line if it can be avoided
@@ -528,6 +557,16 @@ void cpu::enqueue(thread& t)
 {
     trace_sched_queue(&t);
     runqueue.insert_equal(t);
+}
+
+// When the run queue has several threads with equal thread_runtime_compare,
+// enqueue() puts a thread after its equals, while enqueue_first_equal()
+// puts it before its equals. The distinction is mostly interesting for real-
+// time priority threads.
+void cpu::enqueue_first_equal(thread& t)
+{
+    trace_sched_queue(&t);
+    runqueue.insert_before(runqueue.lower_bound(t), t);
 }
 
 void cpu::init_on_cpu()
@@ -864,6 +903,29 @@ void thread::set_priority(float priority)
 float thread::priority() const
 {
     return _runtime.priority();
+}
+
+void thread::set_realtime_priority(unsigned priority)
+{
+    _realtime._priority = priority;
+}
+
+unsigned thread::realtime_priority() const
+{
+    return _realtime._priority;
+}
+
+void thread::set_realtime_time_slice(thread_realtime::duration time_slice)
+{
+    if (time_slice > 0) {
+        WARN_ONCE("set_realtime_time_slice() used but real-time time slices"
+                " not yet supported");
+    }
+    _realtime._time_slice = time_slice;
+}
+
+thread_realtime::duration thread::realtime_time_slice() const {
+    return _realtime._time_slice;
 }
 
 sched::thread::status thread::get_status() const
--- a/include/osv/sched.hh
+++ b/include/osv/sched.hh
@@ -313,6 +313,24 @@ private:
     int _renormalize_count;
 };
 
+// "Normal" threads are fairly time-shared according to the thread_runtime
+// data above. Such normal threads all have _realtime._priority == 0.
+// A "real-time" thread is one where _realtime._priority > 0. The scheduler
+// always picks the thread with the highest _realtime_priority to run next;
+// In particular normal threads run only when no real-time thread wants to
+// run. When several real-time threads with equal _realtime._priority want to
+// run, each one is run for _realtime._time_slice before switching to the
+// next one; If _realtime_time_slice is == 0, there is no limit on the amount
+// of time the thread may run - it will only be preempted when it waits,
+// yields, or a higher-priority thread comes along.
+// The realtime scheduling policy matches the POSIX SCHED_RR and SCHED_FIFO.
+class thread_realtime {
+public:
+    using duration = thread_runtime::duration;
+    unsigned _priority = 0;
+    duration _time_slice = duration::zero();
+};
+
 // "tau" controls the length of the history we consider for scheduling,
 // or more accurately the rate of decay of an exponential moving average.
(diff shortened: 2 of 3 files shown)


core/sched.cc
```diff
@@ -289,8 +290,26 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
             return;
 #endif
         } else if (!called_from_yield) {
-            auto &t = *runqueue.begin();
-            if (p->_runtime.get_local() < t._runtime.get_local()) {
+           auto &t = *runqueue.begin();
+            if (p->_realtime._priority > 0) {
+                // Only switch to a higher-priority realtime thread, or to
+                // the next one with equal priority if yielding.
+                if (t._realtime._priority < p->_realtime._priority) {
```
wkozaczuk, line 297: This means that if somehow the current thread `p` has a high enough priority (for example 99) and all other threads in the run queue have a lower priority, this thread `p` will keep running forever, not even kernel threads will get a chance to do anything. Am I wrong in my analysis? Is this by design?

core/sched.cc
```diff
@@ -289,8 +290,26 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
...
+#ifdef __aarch64__
+                    return switch_data;
+#else
+                    return;
+#endif
+                }
+                if (t._realtime._priority == p->_realtime._priority &&
```
wkozaczuk, line 304: Is this condition right? If the priority of `t` is equal to `p`, then we keep `p` running. Is this correct?

core/sched.cc
```diff
@@ -530,6 +554,16 @@ void cpu::enqueue(thread& t)
...
 }
 
+// When the run queue has several threads with equal thread_runtime_compare,
+// enqueue() puts a thread after its equals, while enqueue_first_equal()
+// puts it before its equals. The distinction is mostly interesting for real-
+// time priority threads.
+void cpu::enqueue_first_equal(thread& t)
+{
+    trace_sched_queue(&t);
+    runqueue.insert_before(runqueue.lower_bound(t), t);
```
wkozaczuk, line 564: I assume this behaves the way we want it - "1st among the equals ... in terms of priority" given the modified `thread_runtime_compare`? And if all threads have priority 0, the `enqueue_first_equal` would behave like `enqueue`?

core/sched.cc
```diff
@@ -350,16 +372,18 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         n->_runtime.add_context_switch_penalty();
     }
     preemption_timer.cancel();
-    if (!called_from_yield) {
-        if (!runqueue.empty()) {
-            auto& t = *runqueue.begin();
-            auto delta = n->_runtime.time_until(t._runtime.get_local());
-            if (delta > 0) {
-                preemption_timer.set_with_irq_disabled(now + delta);
+    if (n->_realtime._priority == 0) {
```
wkozaczuk, line 375: It looks like I am completely ignorant here, but why execute the same logic - the `if (!called_from_yield) {` block and it's else` body ONLY if the priority of the next thread scheduled `n` is equal to 0? Is it because by design (POSIX) the runtime does not matter for real-time threads (aka priority > 0)?

include/osv/sched.hh
```diff
@@ -313,6 +313,24 @@ private:
...
+// "Normal" threads are fairly time-shared according to the thread_runtime
+// data above. Such normal threads all have _realtime._priority == 0.
+// A "real-time" thread is one where _realtime._priority > 0. The scheduler
+// always picks the thread with the highest _realtime_priority to run next;
+// In particular normal threads run only when no real-time thread wants to
+// run. When several real-time threads with equal _realtime._priority want to
+// run, each one is run for _realtime._time_slice before switching to the
+// next one; If _realtime_time_slice is == 0, there is no limit on the amount
+// of time the thread may run - it will only be preempted when it waits,
+// yields, or a higher-priority thread comes along.
```
wkozaczuk, line 325: I think this comment helps me answer some of my other questions above and I understand how the real-time thread may be preempted by yield (the `called_from_yield` related ifs) or if a higher-priority thread comes along. But I am not sure how it would be preempted when it waits - if it has the highest priority it would stay running until it changes or yields, no?

wkozaczuk (changes requested):
Not sure what specific changes but two nitpicks are required.

core/sched.cc
```diff
@@ -289,8 +290,26 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
             return;
 #endif
         } else if (!called_from_yield) {
-            auto &t = *runqueue.begin();
-            if (p->_runtime.get_local() < t._runtime.get_local()) {
+           auto &t = *runqueue.begin();
```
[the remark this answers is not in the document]
nyh: I don't know how this happens, I didn't need to change that line at all ;-) Thanks.

core/sched.cc
nyh: Yes, it is by design that a real-time thread that never yields, is never preempted by any thread with a lower priority or non-real-time priority. 
However, the question about kernel threads is a good one, as in Linux/Unix the things that happen in the kernel weren't called "threads" and the priorities didn't apply to them. In particular cpu::load_balance() which can't even move the poor threads to other CPUs. 

I think the idea is that real-time threads should be short-running - and if a real-time thread just starts an endless loop that never returns, it will cause problems also on Linux.

Maybe the right thing to set some sort of special priority to cpu::load_balance (and anything else?) to allow it to run also when a real-time thread is running. I don't know. I think we can improve this later.

core/sched.cc
```diff
@@ -289,8 +290,26 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
...
+                // the next one with equal priority if yielding.
+                if (t._realtime._priority < p->_realtime._priority) {
+#ifdef __aarch64__
+                    return switch_data;
+#else
+                    return;
+#endif
+                }
+                if (t._realtime._priority == p->_realtime._priority &&
+                        !called_from_yield) {
```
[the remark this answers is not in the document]
nyh: You're right, and it's even worse - I think the if of line 291/292 is not the right thing. I think I have a bug. According to https://man7.org/linux/man-pages/man2/sched_yield.2.html, "If the calling thread is the only thread in the highest priority list at that time, it will continue to run after a call to sched_yield().". So the code wrongly runs some other thread even in non-real-time priorities. I'll need to fix this.

core/sched.cc
```diff
@@ -289,8 +290,26 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
...
+#ifdef __aarch64__
+                    return switch_data;
+#else
+                    return;
+#endif
+                }
+            } else if (p->_runtime.get_local() < t._runtime.get_local()) {
```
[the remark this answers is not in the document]
nyh: Good catch. I should && t._realtime._priority == 0 in this if (see also thread_runtime_compare()). We only stay in this thread - which we know is not real time, if the next thread is not real time AND it has a higher runtime.

core/sched.cc
nyh: The code below deals with the preemption timer. Currently (only SCHED_FIFO implemented, not SCHED_RR) only non-realtime (priority==0) have preemption timer. Such threads need to calculate how much time will pass until the "runtime" of this thread will reach the runtime of the next thread - and set the preemption timer to that time. None of this is needed for real-time threads which just run in FIFO order, with no preemptions and no concept of "runtime" (how much time this thread recently ran).

core/sched.cc
nyh: enqueue_first_equal() is exactly the same as the old enqueue(), it just says that if there are already several threads with exactly the same "runtime_compare", the thread gets put *before* those equal items in the queue, instead of after them as usual. 
Note that the only real case of exact equality of runtime_compare is in real-time threads which have the same real-time priority (for non-realtime threads, the chance that the specific runtime is identical is very small and not interesting).
We need to use it in exactly one place: When a thread of realtime priority 1 is preempted by a thread of realtime priority 2 (for example), we need to put the thread with priority 1 back into the runqueue, and we need to put it **first** (not last) so that when the thread of priority 2 is done, we go back to the original thread with priority 1 - not the next one. This is a requirement of POSIX.

include/osv/sched.hh
nyh: If the realtime-priority thread **waits**, it goes out of the run queue, and not considered for running. Only later, when that thread is done waiting, it is re-inserted into the run queue. At that point it will preempt anything with a lower priority, but if something is running with exactly the same realtime priority, it will be put - as usual (enqueue()) - last among its equals - the thread running at that point will continue to run.

include/osv/sched.hh
```diff
@@ -862,7 +929,12 @@ osv::clock::uptime::duration process_cputime();
 class thread_runtime_compare {
 public:
     bool operator()(const thread& t1, const thread& t2) const {
-        return t1._runtime.get_local() < t2._runtime.get_local();
+        if (t1._realtime._priority > t2._realtime._priority)
```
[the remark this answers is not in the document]
nyh: Right. Before this patch, threads with **lower** "runtime" ran first (the thread with higher runtime recently ran more so we want to give the other one a chance).
After this patch, a thread with **higher** real-time priority runs first.
When threads have real-time priority, the old "runtime" is ignored (the 'else' is reached only if both realtime priorities are zero). There will be "ties" in thread_runtime_compare, but this is fine - the queue is kept in a specific order that doesn't change, so if we have three tied threads that somehow got the order A, B, C, this order will be kept as it until one of them is removed from the queue. This is why the enqueue()/enqueue_first_equal() is important - let's say we have in the queue B, C with tied priority and we want to enter A with the same priority. enque_first_equal() will create the order A, B, C while enqueue() will create the order B, C, A.

nyh:
Pushed a new version with the bugs that @wkozaczuk noticed (thanks!)  fixed.

While fixing these bugs, I decided the scheduler is written in a too complicated way - e.g., it's ugly and confusing that we have two separate places in the context switch code which re-queue the outgoing thread.  But unfortunately, after doing this cleanup the resulting scheduler's context switches wre much slower - and I couldn't figure out why even though I spent several hours trying to fix it :-( So the version I pushed now just fixes the bugs that Waldek found, without trying to improve the understandability of the code.

nyh:
By the way, @wkozaczuk, the Travis CI fails saying that "Error: Cannot perform an interactive login from a non TTY device" when trying to pass docker a password from a TTY. Do you have any idea where/how to fix it?

wkozaczuk:
@nyh I believe I have fixed the issue with the Travis build.

wkozaczuk:
Though, honestly, I am not sure if it is actually building and testing your pull request changes or what is on master.

wkozaczuk (approved):

wkozaczuk:
I think your changes look good and the unit tests pass on `x64`. But I would like to run more tests involving various apps (see https://github.com/cloudius-systems/osv/wiki/Automated-Testing-Framework) before I merge it.

nyh:
Definitely, no hurry, it's been waiting for half a decade already 😂 
We can also wait for the followup patches and benchmarking that started this renewed interest.

wkozaczuk:
I have just committed and pushed this MR into the master branch. I meant to do it properly using the github UI but somehow accidentally did it by manually applying the patch version of it (the side effect of local testing :-)). So I am closing it manually.

Anyhow I have run quite extensive tests including the ones driven by https://github.com/cloudius-systems/osv/wiki/Automated-Testing-Framework and others from `tests/misc-*`. Everything seems to work fine including aarch64. In my case, I did see a little bit of a bump in the length of the colocated context switch time (2-3 ns with ROFS images and 5-10ns with ZFS images - more threads?). But I think it is negligible. 

Closed with 894ceb2c035cba71c9b6bdc83e8d0613850e5dc6

PR #1229, opened 2023-04-15
Correct resolution of liblua53.so

maximdeclercq:
Host `liblua53.so` might not contain all symbols (e.g. `luaL_newstate`). This makes `lddtree` resolve the correct `liblua53.so`

### DIFF
--- a/modules/cli/Makefile
+++ b/modules/cli/Makefile
@@ -1,6 +1,7 @@
 SRC = $(shell readlink -f ../..)
 
 LUA_DIR = $(SRC)/modules/lua/upstream/lua5.3
+LD_LIBRARY_PATH = $(LUA_DIR)
 
 CC=gcc
 CFLAGS=-O2 -g -Wall -std=gnu99 -I $(LUA_DIR)/include


maximdeclercq:
Ok, so I've discovered that this is part of a bigger issue. Dependent files don't seem to be included in the resulting package.

nyh:
I think you're right - modules/cli/Make compiles cli.c with "L$(LUA_DIR) -llua53", where lua53 is a shared library that we deliberately built from source, and then calls scripts/manifest_from_host.sh which wrongly copies the lua53 shared library from the host, using lddtree. However, looking at the manifest_from_host.sh I see that if the library is missing ("not found") on the host, it is simply ignored: On my machine, where lua53.so doesn't even exist (I have lua 5.4 installed), so I get:
```
$ lddtree /home/nyh/osv.tmp2/osv/modules/cli/cli
cli => /home/nyh/osv.tmp2/osv/modules/cli/cli (interpreter => /lib64/ld-linux-x86-64.so.2)
    libedit.so.0 => /lib64/libedit.so.0
    libtinfo.so.6 => /lib64/libtinfo.so.6
    liblua53.so => not found
    libc.so.6 => /lib64/libc.so.6
        ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
```
So liblua53.so is not copied from the host, and instead copied by the lua submodule (modules/lua/usr.manifest) from the locally built one.
But if a lua53.so did exist on the host, we wrongly copy it from the host.

I think we need a way to explicitly **exclude** manifest_from_host.sh taking lua53.so from the host, knowing it will be copied from the locally built one.
Maybe the way done in this patch will also work - lddtree will find the locally-built lua, and will copy it, and it will be copied a second time by modules/lua/usr.manifest but I guess that's fine (?).

wkozaczuk:
I think your patch is logically correct. But I am not convinced the `LD_LIBRARY_PATH` will be passed correctly to the `manifest_from_host.sh`. As a matter of fact, I did a quick experiment by adding an echo and it is not.

I think you need to change to:
```make
export LD_LIBRARY_PATH = $(LUA_DIR)
```

modules/cli/Makefile
```diff
@@ -1,6 +1,7 @@
 SRC = $(shell readlink -f ../..)
 
 LUA_DIR = $(SRC)/modules/lua/upstream/lua5.3
+LD_LIBRARY_PATH = $(LUA_DIR)
```
wkozaczuk, line 4: Needs `export LD_LIBRARY_PATH = $(LUA_DIR)`

wkozaczuk (changes requested):

wkozaczuk (approved):

PR #1232, opened 2023-04-26
Implemented named semaphores

### LINKED ISSUE #1025: Implement named semaphores
When testing various python3 modules, I triggered an unimplemented symbol in `scikit-learn`(aka. `sklearn`):
selectionshot_2019-02-12_03 22 28

This symbol was stubbed as unimplemented in Issue #77. If anyone has time to investigate, `scikit-learn` depends on `numpy` and `scipy` being present on the unikernel. I'll be publishing a PR for osv-apps adding a nice way to add selected python modules to the `python3x` osv-app shortly.

StonewallJohnson:
Closes #1025

### DIFF
--- a/libc/sem.cc
+++ b/libc/sem.cc
@@ -6,17 +6,45 @@
  */
 
 #include <semaphore.h>
+#include <fcntl.h>
 #include <osv/semaphore.hh>
 #include <osv/sched.hh>
 #include <memory>
 #include "libc.hh"
+#include <unordered_map>
 
 // FIXME: smp safety
 
-struct indirect_semaphore : std::unique_ptr<semaphore> {
-    explicit indirect_semaphore(unsigned units)
-        : std::unique_ptr<semaphore>(new semaphore(units)) {}
+struct posix_semaphore : semaphore {
+    private:
+        int references;
+        bool named;
+    public:
+        posix_semaphore(int units, int refs, bool named)
+        : semaphore(units), references(refs), named(named) {}
+
+        void add_reference(){
+            references++;
+        }
+
+        void remove_reference(){
+            references--;
+        }
+
+        bool not_referenced(){
+            return references <= 0;
+        }
+
+        void unlink(){
+            named = false;
+        }
+
+        bool linked(){
+            return named;
+        }
 };
+
+using indirect_semaphore = std::unique_ptr<posix_semaphore>;
 
 indirect_semaphore& from_libc(sem_t* p)
 {
@@ -27,7 +55,8 @@ OSV_LIBC_API
 int sem_init(sem_t* s, int pshared, unsigned val)
 {
     static_assert(sizeof(indirect_semaphore) <= sizeof(*s), "sem_t overflow");
-    new (s) indirect_semaphore(val);
+    posix_semaphore *sem = new posix_semaphore(val, 1, false); 
+    new (s) indirect_semaphore(sem);
     return 0;
 }
 
@@ -74,5 +103,69 @@ int sem_trywait(sem_t* s)
 {
     if (!from_libc(s)->trywait())
         return libc_error(EAGAIN);
+    return 0;
+}
+
+static std::unordered_map<std::string, indirect_semaphore> named_semaphores;
+static mutex named_semaphores_mutex;
+
+OSV_LIBC_API
+sem_t *sem_open(const char *name, int oflag, mode_t mode, unsigned int value)
+{
+    SCOPE_LOCK(named_semaphores_mutex);
+    auto iter = named_semaphores.find(std::string(name));
+    
+    if (iter != named_semaphores.end()) {
+        //opening already named semaphore
+        if (oflag & O_EXCL && oflag & O_CREAT) {
+            errno = EEXIST;
+            return SEM_FAILED;
+        }
+
+        iter->second->add_reference();
+        return reinterpret_cast<sem_t*>(&(iter->second));
+    }
+    else if (oflag & O_CREAT) {
+        //creating new semaphore
+        if (value > SEM_VALUE_MAX) {
+            errno = EINVAL;
+            return SEM_FAILED;
+        }
+        
+        named_semaphores.emplace(std::string(name), std::make_unique<posix_semaphore>(value, 1, true)); 
+        return reinterpret_cast<sem_t *>(&named_semaphores[std::string(name)]);
+    }
+    
+    errno = ENOENT;
+    return SEM_FAILED;
+}
+
+OSV_LIBC_API
+int sem_unlink(const char *name)
+{
+    SCOPE_LOCK(named_semaphores_mutex);
+    auto iter = named_semaphores.find(std::string(name));
+    if (iter != named_semaphores.end()) {
+        iter->second->unlink();
+        if (iter->second->not_referenced()) {
+            sem_destroy(reinterpret_cast<sem_t *>(&iter->second));
+        }
+        named_semaphores.erase(iter);
+        return 0;
+    }
+    
+    errno = ENOENT;
+    return -1;
+}
+
+OSV_LIBC_API
+int sem_close(sem_t *sem)
+{
+    SCOPE_LOCK(named_semaphores_mutex);
+    indirect_semaphore &named_sem = from_libc(sem);
+    named_sem->remove_reference();
+    if (!named_sem->linked() && named_sem->not_referenced()) {
+        sem_destroy(sem);
+    }
     return 0;
 }
--- a/tests/tst-semaphore.c
+++ b/tests/tst-semaphore.c
@@ -11,6 +11,7 @@
 #include <unistd.h>
 #include <semaphore.h>
 #include <assert.h>
+#include <fcntl.h>
 
 #define THREAD_NUMBER 10
 
@@ -54,6 +55,47 @@ int main(void) {
 
     assert(sem_destroy(&sem_sync) == 0);
     assert(sem_destroy(&sem_done) == 0);
+
+    ///Named sempahore test
+
+    //Create and open two handles to a named semaphore
+    sem_t *named_sem1 = sem_open("name", O_CREAT, 0777, 1);
+    assert(named_sem1 != SEM_FAILED);
+    sem_t *named_sem2 = sem_open("name", O_EXCL, 0, 0);
+    assert(named_sem1 == named_sem2);
+
+    //Can't create a new named semaphore without O_CREAT
+    assert(sem_open("other", 0, 0777, 1) == SEM_FAILED);
+    assert(sem_open("other", O_EXCL | O_SYNC, 0777, 1) == SEM_FAILED); 
+
+    //Any other flags should have no effect if the named semaphore does not exist
+    sem_t *named_sem3 = sem_open("other", O_EXCL | O_CREAT | O_SYNC, 0777, 1);
+    assert(named_sem3 != SEM_FAILED);
+    assert(sem_unlink("other") == 0);
+    assert(sem_close(named_sem3) == 0);
+
+    //Close both handles to the semaphore without removing the name
+    assert(sem_close(named_sem1) == 0);
+    assert(sem_close(named_sem2) == 0);
+
+    //Open two more handles to the named sempahore
+    named_sem1 = sem_open("name", 0);
+    assert(named_sem1 != SEM_FAILED);
+    named_sem2 = sem_open("name", 0);
+    assert(named_sem1 == named_sem2);
+
+    //Can't open existing semaphore with O_CREAT and O_EXCL set
+    assert(sem_open("name", O_CREAT | O_EXCL, 0777, 1) == SEM_FAILED);
+    assert(sem_open("name", O_CREAT | O_EXCL | O_SYNC, 0777, 1) == SEM_FAILED);
+
+    //Unlink the named semaphore. Can't open more handles.
+    assert(sem_unlink("name") == 0);
+    assert(sem_unlink("name") == -1);
+    assert(sem_open("name", 0) == SEM_FAILED);
+
+    //Close handles
+    assert(sem_close(named_sem1) == 0);
+    assert(sem_close(named_sem2) == 0);
 
     return 0;
 }


libc/sem.cc
```diff
@@ -76,3 +105,81 @@ int sem_trywait(sem_t* s)
         return libc_error(EAGAIN);
     return 0;
 }
+
+//TODO: when are these initialized?
```
wkozaczuk, line 109: These would be initialized as part of the kernel ELF init which is very early on (look at premain() and inittab in loader.cc).

libc/sem.cc
```diff
@@ -76,3 +105,81 @@ int sem_trywait(sem_t* s)
...
+    errno = ENOENT;
+    return -1;
+}
+
+OSV_LIBC_API
+int sem_close(sem_t *sem){
+    indirect_semaphore *ind_sem = &from_libc(sem);
+    ind_sem->remove_reference();
+
+    if(!ind_sem->linked() && ind_sem->not_referenced()){
```
wkozaczuk, line 181: Should it be "if(ind_sem-linked() ..." instead.

libc/sem.cc
```diff
@@ -76,3 +105,81 @@ int sem_trywait(sem_t* s)
...
+    }
+    
+    errno = ENOENT;
+    return -1;
+}
+
+OSV_LIBC_API
+int sem_close(sem_t *sem){
+    indirect_semaphore *ind_sem = &from_libc(sem);
+    ind_sem->remove_reference();
```
wkozaczuk, line 179: Should we also use scope lock mutex for this method? I do not believe this is thread-safe otherwise.

libc/sem.cc
```diff
@@ -6,16 +6,45 @@
...
+#include <unordered_map>
 
 // FIXME: smp safety
 
 struct indirect_semaphore : std::unique_ptr<semaphore> {
-    explicit indirect_semaphore(unsigned units)
-        : std::unique_ptr<semaphore>(new semaphore(units)) {}
+    private:
+        int references;
+        bool linked;
```
wkozaczuk, line 21: Should we call it named instead of linked and reverse the boolean values? Otherwise I find it a bit confusing.

libc/sem.cc
```diff
@@ -76,3 +105,81 @@ int sem_trywait(sem_t* s)
...
+            iter->second.add_reference();
+            return reinterpret_cast<sem_t*>(&(iter->second));
+        }
+        
+        if(value > SEM_VALUE_MAX){
+            errno = EINVAL;
+            return SEM_FAILED;
+        }
+        /*
+         * Ignoring mode bits bits for now. Do we actually want to worry about
```
wkozaczuk, line 149: I do not think we need to worry about it.

wkozaczuk (changes requested):
I am sorry for the delay and thanks for your PR.

I think it is logically correct but I have found some issues I would like you to fix/clarify. 

Also I think you are missing the unit test to verify your implementation is correct. I suggest you expand the tst-semaphore.cc to cover named semaphores in addition.

libc/sem.cc
StonewallJohnson: Yes, I missed that. That would be necessary to ensure that the reference count can not be decremented simultaneously. Fixed this in most recent push.

libc/sem.cc
StonewallJohnson: Would you like me to remove this comment, or leave/modify it as a design choice?

libc/sem.cc
StonewallJohnson: From what I can tell, it should be checking ```!ind_sem->linked()``` as opposed to ```ind_sem->linked()```. This way, the underlying semaphore will only be destroyed when it is no longer possible to create links to it (after having been unlinked).

I have been working based off of these descriptions of [sem_unlink](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sem_unlink.html) and [sem_close](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sem_close.html) from the IEEE standard. Particularly 

>"If the semaphore has not been removed with a successful call to [sem_unlink()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sem_unlink.html), then sem_close() has no effect on the state of the semaphore. If the [sem_unlink()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sem_unlink.html) function has been successfully invoked for name after the most recent call to [sem_open()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sem_open.html) with O_CREAT for this semaphore, then when all processes that have opened the semaphore close it, the semaphore is no longer accessible."

would seem to me to suggest the current implementation.

libc/sem.cc
StonewallJohnson: Ah, I figured it must be somewhere!

nyh (changes requested):

nyh:
Something in the commits after the first one confused me (a github bug? maybe you repushed in the middle of my review?) so I'll re-review after you address my current crop of comments.

Anyway, nice patch and close to what we need, so thanks! I suggest (maybe you already did that?) that you write the unit test in a way it can be run on Linux as well as OSv, which will allow you to experiment what these functions do in various esoteric cases in Linux that are perhaps difficult to understand from the documentation - and then you can do the same things in Osv.

libc/sem.cc
```diff
@@ -18,11 +18,11 @@
 struct indirect_semaphore : std::unique_ptr<semaphore> {
     private:
         int references;
-        bool linked;
+        bool named;
```
nyh, line 21: If you decided to rename this member, please "squash" this patch with the one that introduced the previous name.

libc/sem.cc
```diff
@@ -76,3 +105,79 @@ int sem_trywait(sem_t* s)
...
+    }
+    
+    errno = ENOENT;
+    return SEM_FAILED;
+}
+
+OSV_LIBC_API
+sem_t *sem_open(const char *name, int oflag, mode_t mode, unsigned int value){
+    SCOPE_LOCK(named_semaphores_mutex);
+    std::unordered_map<std::string, indirect_semaphore>::iterator iter =
```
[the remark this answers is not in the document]
StonewallJohnson: This is very nice to know!

libc/sem.cc
```diff
@@ -76,3 +105,79 @@ int sem_trywait(sem_t* s)
...
+
+OSV_LIBC_API
+int sem_unlink(const char *name){
+    SCOPE_LOCK(named_semaphores_mutex);
+    std::unordered_map<std::string, indirect_semaphore>::iterator iter = 
+                                named_semaphores.find(std::string(name));
+    if(iter != named_semaphores.end()){
+        iter->second.unlink();
+        named_semaphores.erase(std::string(name));
+        return 0;
```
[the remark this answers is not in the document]
StonewallJohnson: Yes, we should check the reference count after the semaphore is unlinked and free it if it has no references after the unlink. I added this in the most recent push.

libc/sem.cc
```diff
@@ -76,3 +105,79 @@ int sem_trywait(sem_t* s)
...
+    return SEM_FAILED;
+}
+
+OSV_LIBC_API
+int sem_unlink(const char *name){
+    SCOPE_LOCK(named_semaphores_mutex);
+    std::unordered_map<std::string, indirect_semaphore>::iterator iter = 
+                                named_semaphores.find(std::string(name));
+    if(iter != named_semaphores.end()){
+        iter->second.unlink();
```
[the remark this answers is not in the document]
StonewallJohnson: Since `sem_unlink` takes a name, rather than a `sem_t *`, I don't think it can called on an unnamed semaphore and so we don't have to worry about trying to call `unlink` on a type that doesn't have that method. I believe that when a name is passed that is already unlinked (i.e. not in the named_semaphores map), it should fail with `ENOENT` as the error number. I took your advice and created some tests in a Linux environment and the results seem consistent with the explanations above.

libc/sem.cc
```diff
@@ -76,3 +105,79 @@ int sem_trywait(sem_t* s)
...
+}
+
+OSV_LIBC_API
+sem_t *sem_open(const char *name, int oflag, mode_t mode, unsigned int value){
+    SCOPE_LOCK(named_semaphores_mutex);
+    std::unordered_map<std::string, indirect_semaphore>::iterator iter = 
+                                named_semaphores.find(std::string(name));
+    
+    if(oflag & O_CREAT){
+        if(iter != named_semaphores.end()){
```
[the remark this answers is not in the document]
StonewallJohnson: Yes, I adjusted this with my most recent push when eliminating the extra `sem_open` function.

libc/sem.cc
```diff
@@ -6,16 +6,45 @@
...
+#include <fcntl.h>
 #include <osv/semaphore.hh>
 #include <osv/sched.hh>
 #include <memory>
 #include "libc.hh"
+#include <unordered_map>
 
 // FIXME: smp safety
 
 struct indirect_semaphore : std::unique_ptr<semaphore> {
```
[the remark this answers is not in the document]
StonewallJohnson: I puzzled over this for a while trying to find an elegant way to solve this problem. I ended up adding a new constructor to the `semaphore` object and relocating the `named` and `references` member variables into this object as well. This way, I could still keep using the `indirect_semaphore` object (now the way it was before I began, just with an adjusted call to the semaphore constructor). I tried adding an additional object in the style of an `indirect_semaphore` (think `indirect_named_semaphore`) and working with that, but when doing that I ran into an issue in `sem_close` trying to convert a `sem_t *` into, and distinguish between, `indirect` and `indirect_named_semaphores`. There may be a way to still do this, but I was trying for a while and I kept pushing the problem somewhere else only for to continue to cause trouble. Let me know if what I have is okay. If not, I can keep working on an alternative.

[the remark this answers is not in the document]
nyh: I'm not sure this is a good idea... The "semaphore" from <osv/semaphore.hh> is supposed to be something generic used in various places in the kernel (it's not used much, but it is used in a few places) and I think it doesn't make sense to enlarge it just to implement the pthread-specific concept of a semaphore.

Why doesn't the idea I proposed above of my_semaphore, an extension of semaphore, work? You'll use this my_semaphore in libc/sem.cc everywhere - both for named and unnamed semaphores (you won't have two separate indirect_*_semaphore types, just one), and outside this file libc/sem.cc the regular semaphore would be used. Does something not work if you try that?

nyh:
Thanks for working on this. I didn't like the new direction of the modified "semaphore" - I'd prefer we only change libc/sem.cc and not stuff used all over OSv - I commented in more detail about this below. I think it should be fairly easy to fix without modifying "semaphore" but maybe I'm missing something?

Please "squash" the patches either to one patch  or to several patches divided by functionality, not historically - because it's hard to review when the first patch adds "wrong" code and the second patch "fixes" it so when I read code in my review I have no idea if it's relevant or will be gone in the next patch (I do the review one patch at a time).

StonewallJohnson:
Maybe I misunderstood you the first time I read your comment, but I think I hopefully understand what you meant now. Is this more in line with what you want?

include/osv/semaphore.hh
```diff
@@ -15,6 +15,7 @@
 class semaphore {
 public:
     explicit semaphore(unsigned val);
+    semaphore(int units, int refs, bool named);
```
nyh, line 18: If I understand correctly, in this version of the patch you don't need to change include/osv/semaphore.hh. And also not semaphore.cc by the way.

libc/sem.cc
```diff
@@ -6,18 +6,46 @@
...
+        void unlink(){
+            named = false;
+        }
+
+        bool linked(){
+            return named;
+        }
 };
 
+using indirect_semaphore = std::unique_ptr<posix_semaphore>;
```
nyh, line 47: Excellent. That's what I meant (I think ;-)).

libc/sem.cc
```diff
@@ -27,7 +55,7 @@ OSV_LIBC_API
 int sem_init(sem_t* s, int pshared, unsigned val)
 {
     static_assert(sizeof(indirect_semaphore) <= sizeof(*s), "sem_t overflow");
-    new (s) indirect_semaphore(val);
+    from_libc(s) = std::make_unique<posix_semaphore>(val, 1, false);
```
nyh, line 58: Why did you change the new(s) to a regular assignment here? I think this change is wrong (but maybe I'm wrong?). The regular assignment operator assumes that the left-hand side is already a valid object and needs to be destructed before being assigned over, but this is not what we want to do here - we want to create a new object.

libc/sem.cc
```diff
@@ -76,3 +104,72 @@ int sem_trywait(sem_t* s)
...
+    auto iter = named_semaphores.find(std::string(name));
+    
+    if (iter != named_semaphores.end()) {
+        //opening already named semaphore
+        if (oflag & O_EXCL) {
+            errno = EEXIST;
+            return SEM_FAILED;
+        }
+
+        iter->second.get()->add_reference();
```
nyh, line 124: I don't think you need to do second.get()->add_reference() - you can simply do the shorter second->add_reference(). Doesn't that work?

libc/sem.cc
```diff
@@ -76,3 +104,72 @@ int sem_trywait(sem_t* s)
...
+
+OSV_LIBC_API
+int sem_unlink(const char *name)
+{
+    SCOPE_LOCK(named_semaphores_mutex);
+    auto iter = named_semaphores.find(std::string(name));
+    if (iter != named_semaphores.end()) {
+        iter->second.get()->unlink();
+        named_semaphores.erase(std::string(name));
+        if (iter->second.get()->not_referenced()) {
```
nyh, line 154: It's not a good idea (and might even not work?) to use the iterator after having removed the item from the map. Please move this not_referenced check before the erase().

libc/sem.cc
```diff
@@ -76,3 +104,72 @@ int sem_trywait(sem_t* s)
...
+}
+
+OSV_LIBC_API
+int sem_unlink(const char *name)
+{
+    SCOPE_LOCK(named_semaphores_mutex);
+    auto iter = named_semaphores.find(std::string(name));
+    if (iter != named_semaphores.end()) {
+        iter->second.get()->unlink();
+        named_semaphores.erase(std::string(name));
```
nyh, line 153: I think there is a way to erase() the iter which you already have and no need to pass it the string name again?

tests/tst-semaphore.c
```diff
@@ -55,5 +56,32 @@ int main(void) {
     assert(sem_destroy(&sem_sync) == 0);
     assert(sem_destroy(&sem_done) == 0);
 
+    ///Named sempahore test
+
+    //Create and open two handles to a named semaphore
+    sem_t *named_sem1 = sem_open("name", O_CREAT, 0777, 1);
```
nyh, line 62: Did you try this test on Linux too? If I read the sem_overview(7) manual page correctly, the name passed to sem_open() must start with a slash and have no other slashes. But what happens when it doesn't start with a slash? Please try to run this test on Linux.

nyh (changes requested):
Thanks. I like this version better. I think there are still a few errors left in the code, so please see my comments.

libc/sem.cc
StonewallJohnson: As it was, that line was causing a compiler error because it was not matching the constructor for the `indirect_semaphore` alias. I opted for the `make_unique` constructor here, but I can see why that might be an issue if it requires the the equal operator for assignment. I have changed this here to construct a new `posix_semaphore`, then use the new assignment for `s` with that new semaphore.

tests/tst-semaphore.c
StonewallJohnson: From the [documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sem_open.html) I have read, it appears that the functionality of `sem_open` when given a name that does not begin with a slash is implementation dependent, as is the meaning of any non-leading slash. In the Linux environment that I am running this in, it passes. However, it does fail when adding any extra non-leading slash, which currently would pass with my implementation. Admittedly, the documentation does say that the name argument should adhere to the rules of a standard pathname, which (to me) would imply that leading slash. But, then it says that a name given without a leading slash causes implementation defined functionality. That extra bit seems to give free reign to do whatever when there is no leading slash, so I guess it is a little unclear. If you would like, I could add in this slash checking (leading, non-leading, both or either) functionality to align more with the Linux documentation.

StonewallJohnson:
I also identified a bug in the way I was checking for the O_CREAT and O_EXCL flag. `sem_open` should fail when both are set and the semaphore already exists. I had previously just been checking if the O_EXCL flag was set, but this would lead to incorrect behavior if the semaphore was found and the O_EXCL flag was set without the O_CREAT flag, which should work normally, as tested in a Linux environment. I also added a corresponding test for this functionality.

tests/tst-semaphore.c
nyh: You're right that the error-case is less important than correctly supporting the correct case (with a leading slash which is the only slash), and no reasonable application would depend on the error case.

core/semaphore.cc
```diff
@@ -69,8 +69,8 @@ bool semaphore::trywait(unsigned units)
     }
 
     return ok;
-}
 
 
 
 
+}
```
nyh, line 76: Please remove this part of the patch (you aren't touching semaphore.cc code at all, so no need to move the brace)

libc/sem.cc
```diff
@@ -76,3 +105,72 @@ int sem_trywait(sem_t* s)
...
+
+OSV_LIBC_API
+sem_t *sem_open(const char *name, int oflag, mode_t mode, unsigned int value)
+{
+    SCOPE_LOCK(named_semaphores_mutex);
+    auto iter = named_semaphores.find(std::string(name));
+    
+    if (iter != named_semaphores.end()) {
+        //opening already named semaphore
+        if (oflag & (O_EXCL | O_CREAT)) {
```
nyh, line 120: Are you sure about O_CREAT here? The Linux manual page suggests that "if  O_CREAT  is specified in oflag, then the semaphore is created if it does not already exist." so it shouldn't be an error to use O_CREAT. Moreover the documentation says "If  both O_CREAT and O_EXCL are specified in oflag, then an error is returned if a semaphore with the  given  name  already exists."

So maybe the right if() should be

```c++
if (oflag & O_EXCL && oflag & O_CREATE)
```
or alternatively
```c++
if (oflag & (O_EXCL | O_CREATE)) == (O_EXCL | O_CREAT))
```
Do you have a test for this?

libc/sem.cc
```diff
@@ -76,3 +105,72 @@ int sem_trywait(sem_t* s)
...
+    return -1;
+}
+
+OSV_LIBC_API
+int sem_close(sem_t *sem)
+{
+    SCOPE_LOCK(named_semaphores_mutex);
+    indirect_semaphore *named_sem = &from_libc(sem);                                  
+    
+    named_sem->get()->remove_reference();
```
nyh, line 171: Nitpick: here and below, you don't need to use get() just to use a smart pointer. Unless I'm missing something, you can write `named_sem->remove_reference()` instead `named_sem->get()->remove_reference()`.

nyh (changes requested):
Thanks. Looks very good but I think you still have a bug with EXCL. I also left a couple of other tiny comments.

nyh:
> I also identified a bug in the way I was checking for the O_CREAT and O_EXCL flag. `sem_open` should fail when both are set and the semaphore already exists. I had previously just been checking if the O_EXCL flag was set, but this would lead to incorrect behavior if the semaphore was found and the O_EXCL flag was set without the O_CREAT flag, which should work normally, as tested in a Linux environment. I also added a corresponding test for this functionality.

I just reviewed your PR and found exactly the same bug :-) I think it's even worse than what you said because it also mishandles O_CREATE without O_EXCL too.
Maybe you forgot to push your latest version?

libc/sem.cc
StonewallJohnson: The get was necessary the way I was using the `named_sem` object, which is a pointer to an `indirect_semaphore`. It isn't possible to use `indirect_semaphore named_sem = from_libc(sem);`, which I initially tried in order to avoid recasting the object, because the copy constructor for std::unique_ptr is deleted. An alternative option is to use `from_libc(sem)->` in place of all `named_sem->get()->`. Any preference?

libc/sem.cc
StonewallJohnson: This was more insidious than I thought. I believe your first check would be the correct method. The second could lead to incorrect behavior if the O_CREAT and O_EXCL flags are both set while any other bit is also set. Because of the exact equality check, any other bit set would mean that we would not be testing just for the failure case of O_CREAT and O_EXCL. I adopted the first check and added tests for this check, as well as some new tests for the following portion where new semaphores are created.

nyh: Thanks. Please push the new version when you're ready.

libc/sem.cc
nyh: Oh, I missed that. Another thing that could work is, I think (didn't try) is to use `&` instead of `*`:
```c++
indirect_semaphore& named_sem = from_libc(sem);
named_sem->remove_reference();
```

StonewallJohnson: This does work and seems like the most concise option.

nyh (approved):

nyh:
@wkozaczuk I see you have "requested changes", can you please re-review? Thanks.

libc/sem.cc
wkozaczuk: Please remove this comment.

libc/sem.cc
wkozaczuk: @nyh Can we resolve it as my original comment was in the context of the outdate code now? And @StonewallJohnson fixed it the way you wanted it?

libc/sem.cc
wkozaczuk: Yeah. Maybe my suggestion was wrong in the context of my newer comments.

.vscode/settings.json
```diff
@@ -0,0 +1,77 @@
+{
+    "files.associations": {
```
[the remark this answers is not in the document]
wkozaczuk: Shall we resolve it?

libc/sem.cc
```diff
@@ -76,3 +105,79 @@ int sem_trywait(sem_t* s)
...
+        iter->second.add_reference();
+        return reinterpret_cast<sem_t*>(&(iter->second));
+    }
+    
+    errno = ENOENT;
+    return SEM_FAILED;
+}
+
+OSV_LIBC_API
+sem_t *sem_open(const char *name, int oflag, mode_t mode, unsigned int value){
```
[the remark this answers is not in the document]
wkozaczuk: Shall we resolve it?

wkozaczuk:
@nyh I saw you have approved this PR. I did try to ask if many older comments of yours should be resolved but I gave up at some point so maybe disregard those. 

I think all my comments have been resolved and the code is so different that they do not apply.

The only thing left is my question about `named` vs `linked` which I may have originally led @StonewallJohnson to believe we should rename `linked` to `named`. Now I think we should reverse my original suggestion and call it `linked`. What do you think?

And one minor thing was about the obsolete permission comment.

wkozaczuk:
Shall we merge this? @ny I think I may be lost on what is still left here to fix or is it ready?

nyh:
> Shall we merge this? @ny I think I may be lost on what is still left here to fix or is it ready?

Sorry, merged now. I already "approved" it earlier, but thought that you had some requests (you are listed on top as "requested changes").

wkozaczuk:
When I try to build this on my Fedora 37 I get this error:
```
libc/sem.cc: In function ‘sem_t* sem_open(const char*, int, mode_t, unsigned int)’:
libc/sem.cc:135:58: error: ‘make_unique’ is not a member of ‘std’
  135 |         named_semaphores.emplace(std::string(name), std::make_unique<posix_semaphore>(value, 1, true));
      |                                                          ^~~~~~~~~~~
libc/sem.cc:135:58: note: ‘std::make_unique’ is only available from C++14 onwards
libc/sem.cc:135:85: error: expected primary-expression before ‘>’ token
  135 |         named_semaphores.emplace(std::string(name), std::make_unique<posix_semaphore>(value, 1, true));
      |                                                                                     ^
libc/sem.cc:135:87: error: left operand of comma operator has no effect [-Werror=unused-value]
  135 |         named_semaphores.emplace(std::string(name), std::make_unique<posix_semaphore>(value, 1, true));
      |                                                                                       ^~~~~
libc/sem.cc:135:97: error: right operand of comma operator has no effect [-Werror=unused-value]
  135 |         named_semaphores.emplace(std::string(name), std::make_unique<posix_semaphore>(value, 1, true));
      |                                                                                                 ^~~~
cc1plus: all warnings being treated as errors
```

I wonder if ` note: ‘std::make_unique’ is only available from C++14 onwards` this means we cannot use it because we build OSv with these flags:
```
CXXFLAGS = -std=gnu++11 $(COMMON) $(cxx-hide-flags)
```

wkozaczuk:
@nyh I have not had a chance to look at your newest PR. But even if I fixed the compilation errors with this patch:
```
diff --git a/libc/sem.cc b/libc/sem.cc
index a698f3df..45e7f666 100644
--- a/libc/sem.cc
+++ b/libc/sem.cc
@@ -109,8 +109,18 @@ int sem_trywait(sem_t* s)
 static std::unordered_map<std::string, indirect_semaphore> named_semaphores;
 static mutex named_semaphores_mutex;
 
+#if __cplusplus < 201402L
+template<class T> struct _Unique_if {
+    typedef std::unique_ptr<T> _Single;
+};
+
+template<class T, class... Args> typename _Unique_if<T>::_Single make_unique(Args&&... args) {
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+#endif
+
 OSV_LIBC_API
-sem_t *sem_open(const char *name, int oflag, mode_t mode, unsigned int value)
+sem_t *sem_open(const char *name, int oflag, ...)
 {
     SCOPE_LOCK(named_semaphores_mutex);
     auto iter = named_semaphores.find(std::string(name));
@@ -127,12 +137,21 @@ sem_t *sem_open(const char *name, int oflag, mode_t mode, unsigned int value)
     }
     else if (oflag & O_CREAT) {
         //creating new semaphore
+        va_list ap;
+        va_start(ap, oflag);
+        va_arg(ap, mode_t);
+        unsigned value = va_arg(ap, unsigned);
+        va_end(ap);
         if (value > SEM_VALUE_MAX) {
             errno = EINVAL;
             return SEM_FAILED;
         }
         
+#if __cplusplus < 201402L
+        named_semaphores.emplace(std::string(name), make_unique<posix_semaphore>(value, 1, true)); 
+#else
         named_semaphores.emplace(std::string(name), std::make_unique<posix_semaphore>(value, 1, true)); 
+#endif
         return reinterpret_cast<sem_t *>(&named_semaphores[std::string(name)]);
     }
     
@@ -168,4 +187,4 @@ int sem_close(sem_t *sem)
         sem_destroy(sem);
     }
     return 0;
-}
\ No newline at end of file
+}
```

and run the unit test I get this crash:
```
Thread 1: Incremented 10th
Thread 9: Incremented 10th
Thread 10: Incremented 10th
Thread 3: Incremented 10th
Thread 8: Incremented 9th
Thread 8: Incremented 10th
Thread 4: Incremented 10th
Thread 6: Incremented 10th
Thread 7: Incremented 10th
Thread 5: Incremented 10th
Aborted

[backtrace]
0x0000000040216ca5 <???+1075932325>
0x00000000403d205c <osv::handle_mmap_fault(unsigned long, int, exception_frame*)+28>
0x00000000402aae9b <mmu::vm_fault(unsigned long, exception_frame*)+315>
0x00000000402faa2b <page_fault+139>
0x00000000402f99b6 <???+1076861366>
0x00001000000012c7 <???+4807>
0x909d3eebf6bfcd61 <???+-155202207>
0x0000000000000000 <???+0>
0x00000000403d50df <???+1077760223>
0xef0f661974f83947 <???+1962424647>
qemu failed.

(gdb) bt
#0  0x0000000040301362 in processor::cli_hlt () at arch/x64/processor.hh:247
#3  0x00000000402370f0 in abort (fmt=fmt@entry=0x405c3b13 "Aborted\n") at runtime.cc:145
#4  0x0000000040202988 in abort () at runtime.cc:106
#5  0x0000000040216ca6 in osv::generate_signal (siginfo=..., ef=ef@entry=0x400000f5a078) at libc/signal.cc:130
#6  0x00000000403d205d in osv::handle_mmap_fault (addr=addr@entry=35184380485632, sig=sig@entry=11, ef=ef@entry=0x400000f5a078) at libc/signal.cc:145
#7  0x00000000402aae9c in mmu::vm_sigsegv (ef=0x400000f5a078, addr=35184380485632) at core/mmu.cc:1428
#8  mmu::vm_fault (addr=addr@entry=35184380485632, ef=ef@entry=0x400000f5a078) at core/mmu.cc:1456
#9  0x00000000402faa2c in page_fault (ef=0x400000f5a078) at arch/x64/mmu.cc:42
#10 <signal handler called>
#11 0x0000200000802000 in ?? ()
#12 0x00000000403d49c6 in std::default_delete<posix_semaphore>::operator() (__ptr=<optimized out>, this=<optimized out>) at /usr/include/c++/12/bits/unique_ptr.h:89
#13 std::unique_ptr<posix_semaphore, std::default_delete<posix_semaphore> >::~unique_ptr (this=0x600000f2c668, __in_chrg=<optimized out>) at /usr/include/c++/12/bits/unique_ptr.h:396
#14 sem_destroy (s=0x600000f2c668) at libc/sem.cc:66
#15 sem_close (sem=sem@entry=0x600000f2c668) at libc/sem.cc:188
#16 0x00001000000012c8 in main () at /home/wkozaczuk/projects/osv-master/tests/tst-semaphore.c:75
#17 0x0000000040397e6d in osv::application::run_main (this=0x600000bf9c10) at core/app.cc:426
#18 0x0000000040398059 in operator() (app=<optimized out>, __closure=0x0) at core/app.cc:236
#19 _FUN () at core/app.cc:238
#20 0x00000000403cdde6 in operator() (__closure=0x600000d76500) at libc/pthread.cc:116
#21 std::__invoke_impl<void, pthread_private::pthread::pthread(void* (*)(void*), void*, sigset_t, const pthread_private::thread_attr*)::<lambda()>&> (__f=...)
    at /usr/include/c++/12/bits/invoke.h:61
#22 std::__invoke_r<void, pthread_private::pthread::pthread(void* (*)(void*), void*, sigset_t, const pthread_private::thread_attr*)::<lambda()>&> (__fn=...)
    at /usr/include/c++/12/bits/invoke.h:154
#23 std::_Function_handler<void(), pthread_private::pthread::pthread(void* (*)(void*), void*, sigset_t, const pthread_private::thread_attr*)::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...) at /usr/include/c++/12/bits/std_function.h:290
#24 0x00000000403684d8 in sched::thread::main (this=0x400000f55040) at core/sched.cc:1409
#25 sched::thread_main_c (t=0x400000f55040) at arch/x64/arch-switch.hh:350
#26 0x00000000402fa933 in thread_main () at arch/x64/entry.S:116
(gdb) quit
```

So I either have a bug in my patch (around `make_unique` maybe) or there is a bug in the original implementation.

wkozaczuk:
I used this construct from your patch instead:

```
named_semaphores.emplace(std::string(name),
    std::unique_ptr<posix_semaphore>(new posix_semaphore(value, 1, true)));
```

and I still get the same crash.

wkozaczuk:
I have opened new issue to track this problem - #1258

PR #1236, opened 2023-06-07
brk/sbrk: implement program break

### LINKED ISSUE #1138: Implement brk syscall to support statically linked executables
The very naive implementation using `mmap()` might look like this:
```
diff --git a/linux.cc b/linux.cc
index 20da2146..f48a3107 100644
--- a/linux.cc
+++ b/linux.cc
@@ -37,6 +37,8 @@
 #include <sys/file.h>
 #include <sys/unistd.h>
 #include <sys/random.h>
 
 #include <unordered_map>
 
@@ -371,6 +373,38 @@ static int pselect6(int nfds, fd_set *readfds, fd_set *writefds,
     return pselect(nfds, readfds, writefds, exceptfds, timeout_ts, NULL);
 }
 
+#define ARBITRARY_PROGRAM_BREAK_START 0x300000000000ul
+#define ARBITRARY_PROGRAM_BREAK_SIZE    0x100000
+extern void *program_break;
+
+static u64 long_brk(void *addr)
+{
+    if (!addr && !program_break) {
+        program_break = mmap((void*)ARBITRARY_PROGRAM_BREAK_START, ARBITRARY_PROGRAM_BREAK_SIZE, PROT_READ|PROT_WRITE, MAP_ANONYMOUS|MAP_PRIVATE, -1, 0);
+        addr = program_break;
+    }
+    return (u64)addr;
+}
+#define __NR_long_brk __NR_brk
 long syscall(long number, ...)
 {
     // Save FPU state and restore it at the end of this function
@@ -451,6 +485,10 @@ long syscall(long number, ...)
     SYSCALL2(mkdir, char*, mode_t);
 #endif
     SYSCALL3(mkdirat, int, char*, mode_t);
+    SYSCALL1(long_brk,void*);
```

To make it possible to de-allocate memory we would need to do something more interesting than that and truly track the `addr` argument that normally increases and decreases.

wkozaczuk:
This patch implements the program break and the related functions `brk()` and `sbrk()` and the corresponding syscall. The program counter implementation is necessary to support statically linked executables as well as applications launched by the Linux dynamic linker (aka `ld-linux-x86-64.so.2`) on OSv.

Please note that both the functions and the syscall are considered legacy API however we need to support it for the reasons stated above. Glibc and some memory allocators like **jemalloc** use `brk` for some type of memory allocations. 

In essence, the program break implementation is based on `mmap()`. For details please look into the relevant code comments.

This patch also adds a new unit test to verify if the program break works correctly.

This is one of the very first patches of the series implementing the missing functionality to support statically linked executables as well as applications launched by the Linux dynamic linker.

Fixes #1138

### DIFF
--- a/libc/mman.cc
+++ b/libc/mman.cc
@@ -16,6 +16,7 @@
 #include <osv/stubbing.hh>
 #include "libc/libc.hh"
 #include <safe-ptr.hh>
+#include <atomic>
 
 #ifndef MAP_UNINITIALIZED
 #define MAP_UNINITIALIZED 0x4000000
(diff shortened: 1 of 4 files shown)


wkozaczuk:
@nyh What do you think about this PR? Would you have a bit of time to review it?

PR #1242, opened 2023-08-10
pselect6: implement the case when sigmask is not NULL

wkozaczuk:
This patch completes the implementation of the pselect6 syscall to handle the case when sigmask is not NULL. It does by pthread_sigmask() and select() as explained by the manual page https://linux.die.net/man/2/pselect6.

### DIFF
--- a/linux.cc
+++ b/linux.cc
@@ -423,21 +423,44 @@ static int sys_ioctl(unsigned int fd, unsigned int command, unsigned long arg)
     }
 }
 
+struct sys_sigset {
+    const sigset_t *ss;     /* Pointer to signal set */
+    size_t          ss_len; /* Size (in bytes) of object pointed to by 'ss' */
+};
+
 static int pselect6(int nfds, fd_set *readfds, fd_set *writefds,
-                   fd_set *exceptfds, const struct timespec *timeout_ts,
-                   void *sig)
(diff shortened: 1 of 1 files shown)


linux.cc
```diff
@@ -504,7 +502,7 @@ OSV_LIBC_API long syscall(long number, ...)
     SYSCALL2(flock, int, int);
     SYSCALL4(pwrite64, int, const void *, size_t, off_t);
     SYSCALL1(fdatasync, int);
-    SYSCALL6(pselect6, int, fd_set *, fd_set *, fd_set *, const struct timespec *, void *);
+    SYSCALL6(pselect6, int, fd_set *, fd_set *, fd_set *, const struct timespec *, const sigset_t*);
```
nyh, line 505: Are you sure about this? The pselect(2) manual page says that `The final argument of the pselect6() system call is not a sigset_t * pointer, but is instead a structure of the form:`.

You even had a comment about this, which you removed in this PR.

How did you test this new implementation?

wkozaczuk: Yeah, I have messed it up. I implemented this part to support one of the apps I was testing the static executables. But clearly, my tests have missed this bug.

wkozaczuk: Addressed in the latest version of the PR.

PR #1243, opened 2023-08-10
ioctl: translate kernel version of termios when TCGETS

wkozaczuk:
As this glibc implementation of tcgetattr illustrates - https://github.com/bminor/glibc/blob/ae612c45efb5e34713859a5facf92368307efb6e/sysdeps/unix/sysv/linux/tcgetattr.c, the kernel version of the termios structure differs slightly in the length of the c_cc field. Therefore this patch changes the ioctl syscall to manually translate between two versions of this structure.

### DIFF
--- a/linux.cc
+++ b/linux.cc
@@ -39,6 +39,7 @@
 #include <sys/unistd.h>
 #include <sys/random.h>
 #include <sys/vfs.h>
+#include <termios.h>
 
 #include <unordered_map>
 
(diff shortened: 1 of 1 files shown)


nyh:
For posterity, here is how I explained this issue on the mailing list (I should have opened an issue with the description of the problem, but apparently didn't):

OSv's "struct termios" from include/api/termios.h is identical to that which gcc defines in /usr/include/bits/termios-struct.h, but it turns out that glibc does NOT assume that the kernel uses this termios structure, but something else called
__kernel_termios

So although our tcgetattr() function should return our usual termios structure as-is, TCGETS should do something different - it should write a __kernel_termios structure.
I think _kernel_termios is ktermios that you have in  /usr/include/asm-generic/termbits.h - you can see there that NCCS (the number of control characters) is lower, just 19 instead of 32, which explains the overflow you noticed.

I think the fix is simple - just copy a part of the termios struct - only up to the 19th c_cc member, not the whole thing.

linux.cc
```diff
@@ -386,9 +387,40 @@ static long sys_getcwd(char *buf, unsigned long size)
 // to Linux signature of this system call. The underlying ioctl function which we delegate to
 // is variadic and takes slightly different paremeters and therefore cannot be used directly
 // as other system call implementations can.
+#define __NCCS 19
```
nyh, line 390: nitpick: maybe call this KERNEL_NCCS to make it clearer it's related to this kernel structure.
In general using double underscores in names is not a good idea - it suggest some builtin macro, which this isn't.

linux.cc
```diff
@@ -386,9 +387,40 @@ static long sys_getcwd(char *buf, unsigned long size)
...
+struct __kernel_termios {
+    tcflag_t c_iflag;
+    tcflag_t c_oflag;
+    tcflag_t c_cflag;
+    tcflag_t c_lflag;
+    cc_t c_line;
+    cc_t c_cc[__NCCS];
+    speed_t c_ispeed;
+    speed_t c_ospeed;
+};
```
nyh, line 400: Please add a comment where you got this particular definition. 

Is this ktermios from /usr/include/asm-generic/termbits.h? (by the way, I have no idea  what are "termios" and "termios2" in the same header file...).

I looked in the glibc source code, and it assumes (sysdeps/unix/sysv/linux/kernel_termios.h) a slightly different definition.
```c
/* The following corresponds to the values from the Linux 2.1.20 kernel.  */

#define __KERNEL_NCCS 19

struct __kernel_termios
  {
    tcflag_t c_iflag;           /* input mode flags */
    tcflag_t c_oflag;           /* output mode flags */
    tcflag_t c_cflag;           /* control mode flags */
    tcflag_t c_lflag;           /* local mode flags */
    cc_t c_line;                /* line discipline */
    cc_t c_cc[__KERNEL_NCCS];   /* control characters */
  };
```
Without the speed members, but then glibc's source code sysdeps/unix/sysv/linux/tcgetattr.c assumes a c_ispeed and c_ospeed members in this structure. So I'm completely confused now.
Maybe worth checking what the actual kernel has.

nyh (changes requested):

linux.cc
wkozaczuk: Great idea!

linux.cc
wkozaczuk: You are right about the `__kernel_termios` - we should make it like in `kernel_termios.h`. 

Regarding `c_ispeed` and `c_ospeed`, I think it is all OK given this code from glibc `tcgetattr.c`:
```c
      termios_p->c_line = k_termios.c_line;
#if _HAVE_STRUCT_TERMIOS_C_ISPEED
# if _HAVE_C_ISPEED
      termios_p->c_ispeed = k_termios.c_ispeed;
# else
      termios_p->c_ispeed = k_termios.c_cflag & (CBAUD | CBAUDEX);
# endif
#endif
```

The `_HAVE_STRUCT_TERMIOS_C_ISPEED` is 1 in the generic header but `_HAVE_C_ISPEED` is 0 so in actuality, glibc would take the information from `c_cflag` instead. The code accounts for `c_ispeed` that is present for mips and powerpc architectures.

PR #1245, opened 2023-08-13
dynamic linker: handle *IRELATIVE relocations

wkozaczuk:
The OSv dynamic linker handles X_*_RELATIVE relocations which are resolved as base + addend. The X_*_IRELATIVE relocations on the other hand need to be resolved by calling a function located at "base + addend".

So this patch enhances the dynamic linker to resolve the IRELATIVE relocations when processing both PLT and RELA sections according to this formula:

```c++
*static_cast<void**>(addr) = reinterpret_cast<void *(*)()>(_base + addend)()
```

Sometimes the function invoked to calculate the address when resolving an IRELATIVE relocation has to be relocated before. For that reason, we change the order in relocate() and relocate PLT before RELA and also set elf_resolve_pltgot pointer before processing the relocations in relocate_pltgot().

This patch also adds a unit test which is built as pie and contains IRELATIVE relocations:

```
Relocation section '.rela.dyn' at offset 0x570 contains 9 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
...
0000000000004010  0000000000000025 R_X86_64_IRELATIVE                        11d0

Relocation section '.rela.plt' at offset 0x648 contains 3 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
...
0000000000003fd0  0000000000000025 R_X86_64_IRELATIVE                        11d0
```

The unit test is based on the example from here - https://sourceware.org/bugzilla/show_bug.cgi?id=28218 - which explains some edge cases. There is also this one - https://sourceware.org/bugzilla/show_bug.cgi?id=13302.

Finally, please note we are adding IRELATIVE handling logic primarily in anticipation of supporting statically linked executables. But also one of the users tried to run a dynamically linked executable that had IRELATIVE relocation.


tests/tst-reloc.c
```diff
@@ -0,0 +1,28 @@
...
+
+int func() __attribute__((ifunc("func_resolver")));
+
+// .rela.dyn.rel => R_X86_64_64 referencing STT_GNU_IFUNC in .rela.dyn
+int (*fptr_func)() = func;
+
+static void test_indirect_relocations()
+{
+#ifdef __x86_64__
+    printf("%d\n", func());
```
nyh, line 21: nitpick: you could have tested that func() returns 42.
But I admit that any reasonable failure mode would have this test crashing so even without testing anything, it's a good-enough test.

nyh (approved):

nyh:
@wkozaczuk by the way I merged this PR with the scripts/pull_github_pr.sh script (see https://github.com/cloudius-systems/osv/pull/1247) you can see out it was added as a single patch (not a merge commit) and has the "Closes" tag that closed this PR.

PR #1246, opened 2023-08-14
auxv: expose AT_MINSIGSTKSZ and AT_RANDOM

wkozaczuk:
This patch adds AT_MINSIGSTKSZ and AT_RANDOM to the auxiliary vector. 
These two values are needed to support static executables.

### DIFF
--- a/core/app.cc
+++ b/core/app.cc
@@ -20,6 +20,7 @@
 #include <boost/range/algorithm/transform.hpp>
 #include <osv/wait_record.hh>
 #include "libc/pthread.hh"
+#include <sys/random.h>
 
 using namespace boost::range;
 
@@ -333,6 +334,8 @@ void application::main()
     // _entry_point() doesn't return
 }
 
+static u64 random_bytes[2];
+
 void application::prepare_argv(elf::program *program)
 {
     // Prepare program_* variable used by the libc
@@ -360,11 +363,21 @@ void application::prepare_argv(elf::program *program)
     }
 
     // Load vdso library if available
-    int auxv_parameters_count = 2;
+    int auxv_parameters_count = 4;
     _libvdso = program->get_library("libvdso.so");
     if (!_libvdso) {
         auxv_parameters_count--;
         WARN_ONCE("application::prepare_argv(): missing libvdso.so -> may prevent shared libraries specifically Golang ones from functioning\n");
+    }
+
+    // Initialize random bytes array so it can be passed as AT_RANDOM auxv vector
+    if (getrandom(random_bytes, sizeof(random_bytes), 0) != sizeof(random_bytes)) {
+        // Fall back to rand_r()
+        auto d = osv::clock::wall::now().time_since_epoch();
+        unsigned seed = std::chrono::duration_cast<std::chrono::nanoseconds>(d).count() % 1000000000;
+        for (unsigned idx = 0; idx < sizeof(random_bytes)/(sizeof(int)); idx++) {
+            reinterpret_cast<int*>(random_bytes)[idx] = rand_r(&seed);
+        }
     }
 
     // Allocate the continuous buffer for argv[] and envp[]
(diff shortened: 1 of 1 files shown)


core/app.cc
```diff
@@ -401,6 +402,14 @@ void application::prepare_argv(elf::program *program)
     _auxv[auxv_idx].a_type = AT_PAGESZ;
     _auxv[auxv_idx++].a_un.a_val = sysconf(_SC_PAGESIZE);
 
+    _auxv[auxv_idx].a_type = AT_MINSIGSTKSZ;
+    _auxv[auxv_idx++].a_un.a_val = sysconf(_SC_MINSIGSTKSZ);
+
+    random_bytes[0] = rand();
+    random_bytes[1] = rand();
```
nyh, line 409: I have no idea what's the purpose of these random bytes, but calling rand() might not be the right thing to do... It's not thread-safe (although I guess the risk in practice is near zero) and uses some seed that the user might have set for a different purpose. Using your own seed (like the current time) and then rand_r() is probably better, although I wonder if you need to get higher quality random numbers from /dev/urandom or something?

More seriously, rand() only returns an "int", i.e., 4 random bytes, not 8. So half of your "random bytes" are zero, unless I'm missing something.

nyh (changes requested):

core/app.cc
wkozaczuk: I think a better solution is to use `getrandom()` and if it fails fall back to `rand_r()`.

core/app.cc
```diff
@@ -360,13 +363,25 @@ void application::prepare_argv(elf::program *program)
...
-    int auxv_parameters_count = 2;
+    int auxv_parameters_count = 4;
     _libvdso = program->get_library("libvdso.so");
     if (!_libvdso) {
         auxv_parameters_count--;
         WARN_ONCE("application::prepare_argv(): missing libvdso.so -> may prevent shared libraries specifically Golang ones from functioning\n");
     }
 
+    // Initialize random bytes array so it can be passed as AT_RANDOM auxv vector
+    if (getrandom(random_bytes, sizeof(random_bytes), GRND_RANDOM) != sizeof(random_bytes)) {
```
nyh, line 374: Nitpick: since you anyway intend to fall back to pseudo-random number generation, I think you could have just used urandom - i.e., not pass GRND_RANDOM. Then getrandom() is guaranteed to succeed (if number of bytes is less than 256). 
But there's nothing wrong your code either.

core/app.cc
```diff
@@ -360,13 +363,25 @@ void application::prepare_argv(elf::program *program)
...
+        // Fall back to rand_r()
+        u64 time_now = std::chrono::duration_cast<std::chrono::nanoseconds>(
+            osv::clock::wall::now().time_since_epoch()).count();
+        for (unsigned idx = 0; idx < sizeof(random_bytes)/(sizeof(int)); idx ++) {
+            // Point to the least significant part of time time as seed
+            reinterpret_cast<int*>(random_bytes)[idx] = rand_r(reinterpret_cast<unsigned int*>(&time_now) + 1);
+            time_now++; // Make sure each time different seed is passed to rand_r()
```
nyh, line 381: I don't understand this line. rand_r takes the seed you point it to, and also writes the next seed back to it. You don't need to increment the seed...

nyh (changes requested):

core/app.cc
wkozaczuk: I think it is an excellent idea as it will simplify the code.

core/app.cc
wkozaczuk: Once I follow you suggestion above this become irrelevant.

nyh (changes requested):

core/app.cc
```diff
@@ -360,13 +363,16 @@ void application::prepare_argv(elf::program *program)
...
-    int auxv_parameters_count = 2;
+    int auxv_parameters_count = 4;
     _libvdso = program->get_library("libvdso.so");
     if (!_libvdso) {
         auxv_parameters_count--;
         WARN_ONCE("application::prepare_argv(): missing libvdso.so -> may prevent shared libraries specifically Golang ones from functioning\n");
     }
 
+    // Initialize random bytes array so it can be passed as AT_RANDOM auxv vector
+    assert(getrandom(random_bytes, sizeof(random_bytes), 0) == sizeof(random_bytes));
```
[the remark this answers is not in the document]
wkozaczuk: I fixed it by partially reverting to the previous version of the code and addressing some concerns of yours.

PR #1249, opened 2023-08-16
syscalls: make exit terminate current thread

wkozaczuk:
Unlike the libc `exit()` function, the system call exit in Linux is supposed to terminate the current thread. This patch fixes the current incorrect implementation that delegates to `exit()`.

### DIFF
--- a/linux.cc
+++ b/linux.cc
@@ -355,7 +355,7 @@ int rt_sigprocmask(int how, sigset_t * nset, sigset_t * oset, size_t sigsetsize)
 
 static int sys_exit(int ret)
 {
-    exit(ret);
+    sched::thread::current()->exit();
     return 0;
 }
 


linux.cc
```diff
@@ -355,7 +355,7 @@ int rt_sigprocmask(int how, sigset_t * nset, sigset_t * oset, size_t sigsetsize)
 
 static int sys_exit(int ret)
 {
-    exit(ret);
+    sched::thread::current()->exit();
```
nyh, line 358: What happens in Linux when you syscall "exit" on the last or only thread? Doesn't the entire process exit?
Maybe we need to add to sys_exit() a mechanism where if the last application thread (or something...) exits, it calls exit() to power off the VM?

wkozaczuk: Here is what the [Linux manual](https://man7.org/linux/man-pages/man2/exit.2.html) states:

"By contrast with the C library wrapper function, the raw Linux `_exit()` system call terminates only the calling thread, and actions such as reparenting child processes or sending SIGCHLD to the parent process are performed only if this is the last thread in the thread group.

Up to glibc 2.3, the `_exit()` wrapper function invoked the kernel system call of the same name.  Since glibc 2.3, the wrapper function invokes [exit_group(2)](https://man7.org/linux/man-pages/man2/exit_group.2.html), in order to terminate all of the threads in a process."

From the 1st paragraph, we would need to see how "reparenting child processes or sending SIGCHLD to the parent process" applies to OSv.

Even if we keep it as `sched::thread::current()->exit()` without doing anything special if the last thread, I believe that once all app threads terminate, the VM would terminate as well just based on how we join on starting in `do_main_thread`. BTW for a long time (since 2002) glibc called `exit_group` in the implementation of C function `_exit()`  and OSv `exit_group` syscall would terminate VM. 

BTW relatedly the `exit/exit_group` syscalls from `strace`-ing this example:
```c
#include <stdio.h>
#include <unistd.h>
#include <pthread.h>

void* secondary(void *ignore)
{
    printf("start\n");
    sleep(1);
    printf("end\n");
}

void main() {
    pthread_t threads[10];
    for (int i = 0; i < 10; i++) 
       pthread_create(&threads[i], NULL, secondary, NULL);

    printf("Created 10 threads\n");

    void *status;
    for (int i = 0; i < 10; i++) 
       pthread_join(threads[i], &status);
}
```

```
[pid 159125] exit(0 <unfinished ...>
[pid 159123] exit(0 <unfinished ...>
[pid 159124] exit(0 <unfinished ...>
[pid 159126] exit(0 <unfinished ...>
[pid 159122] exit(0 <unfinished ...>
[pid 159127] exit(0 <unfinished ...>
[pid 159128] exit(0 <unfinished ...>
[pid 159129] exit(0 <unfinished ...>
[pid 159130] exit(0 <unfinished ...>
[pid 159131] exit(0 <unfinished ...>
exit_group(0)
```

PR #1250, opened 2023-08-16
syscalls: expose statx

wkozaczuk:
Nadav Har'EL added a basic implementation of `statx` at VFS layer. 
This patch exposes it as a syscall.


nyh (approved):
Strictly speaking, there shouldn't be a <sys/statx.h> - it should be included in <sys/stat.h>. But I think we can let that slide. I'll commit this.

PR #1252, opened 2023-08-23
libvdso: load it from kernel instead of filesystem

wkozaczuk:
This patch changes how libvdso (Virtual Dynamic Shared Object) is constructed and loaded.

Originally, libvdso support was added as part of the commit e60339de2676b186c1ec957102a5a384506d127e to support Golang apps that use libvdso library local functions to speed up execution over corresponding syscalls. During discovery work to support statically linked executables or executables loaded by Linux dynamic linker on OSv, I noticed that glibc would use libvdso as well (see https://man7.org/linux/man-pages/man7/vdso.7.html), however it would fail to work with our current libvdso. More specifically the issue had to do with the fact our current libvdso contains 4 PT_LOAD segments and OSv loads them in memory from filesystem in a way that is not compatible with what glibc expects. The fragment from the linux vdso linker script (https://elixir.bootlin.com/linux/v4.1.31/source/arch/x86/vdso/vdso-layout.lds.S) has this comment next to the PHDRS tag:

"We must supply the ELF program headers explicitly to get just one
 PT_LOAD segment, and set the flags explicitly to make segments read-only."

So we want to make our libvdso.so look similar to what the Linux one would look like.

To that extent we define new libvdso linker script that enforces single segment layout and exports all symbols. Relatedly, we modify the main makefile to build libvdso using new linker script and with egerly resolvable symbols. On top of this we put the content of libvdso.so into an libvdso.o object that becomes part of the kernel ELF in a similar way bootfs content is. We also remove libvdso.so from all manifest files.

The beginning of the vdso in the kernel image - libvdso_start - is where we load the vdso ELF object from as a object::memory_image at the end of the kernel ELF initialization in the elf::program() constructor. The mechanism of loading and processing of libvdso.so is handled by new method program::initialize_libvdso().

The advantage of this approach is that loading of vdso is no longer dependant on the filesystem, it takes less memory (the object is loaded only once) and is smaller - 2656 vs 13544 before and is probably slightly faster (all is in memory already). The disadvantage is that it makes the kernel 4K larger.


Makefile
```diff
@@ -2180,12 +2180,18 @@ libgcc_s_dir := ../../$(aarch64_gccbase)/lib64
...
 	$(call quiet, olddir=`pwd`; cd $(out); "$$olddir"/scripts/mkbootfs.py -o bootfs.bin -d bootfs.bin.d -m "$$olddir"/$(bootfs_manifest), MKBOOTFS $@)
 
 $(out)/bootfs.o: $(out)/bootfs.bin
 $(out)/bootfs.o: ASFLAGS += -I$(out)
 
+$(out)/libvdso-stripped.so: $(out)/libvdso.so
+	$(call quiet, $(STRIP) $(out)/libvdso.so -o $(out)/libvdso-stripped.so, STRIP libvdso.so -> libvdso-stripped.so)
+
+$(out)/libvdso.o: $(out)/libvdso-stripped.so
+$(out)/libvdso.o: ASFLAGS += -I$(out)
```
nyh, line 2193: Nice black magic! :-)
I would consider using a different name for libvdso.o - I think it can really confuse some future reader libvdso.o isn't build from libvdso.cc (as one would assume) but rather goes libvdso.cc -> libvdso.so -> libvdso-stripped.so -> libvdso.S -> libvdso.o
I have an idea - how about renaming libvdso.S to libvdso-content.S (or something) and calling the result libvdso-content.o instead of libvdso.o?

nyh (approved):

core/elf.cc
```diff
@@ -791,9 +792,9 @@ void object::relocate_pltgot()
     pltgot[1] = this;
     pltgot[2] = reinterpret_cast<void*>(__elf_resolve_pltgot);
 
-    bool bind_now = dynamic_exists(DT_BIND_NOW) || mlocked() ||
+    bool bind_now = dynamic_exists(DT_BIND_NOW) ||
```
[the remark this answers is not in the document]
wkozaczuk: The new `libvdso.so` is built with `BIND_NOW` so we can have symbols resolved eagerly. However, there are many old and new ways GCC marks it as you can see in this elaborate `if` expression which I think you expanded on at some point (see cc6e8f59425bc32bc2944856324b81b287d50390). And because of that the `mlocked()` would be evaluated second (before the remaining ones) and would try to call `sections()` which would eventually call `read()` which throws an exception for `memory_image`. So I could have fixed it differently, but changing the order was the easiest and possibly the right thing to do as `mlocked()` is really the OSv-specific "last-resort" type of scenario so it should be treated as such.

Makefile
wkozaczuk: Agree, I will make this change and push it ASAP.

wkozaczuk: I have updated the code with you latest suggestions.

PR #1253, opened 2023-08-30
Dynamic linker: support loading and processing statically linked executables

wkozaczuk:
The two commits provide necessary modifications to the OSv dynamic linker to support loading and processing statically linked executables.

Please note these changes are NOT enough to make OSv run statically linked executables.

### DIFF
--- a/arch/aarch64/arch-elf.cc
+++ b/arch/aarch64/arch-elf.cc
@@ -66,7 +66,7 @@ bool object::arch_relocate_rela(u32 type, u32 sym, void *addr,
         if (sym) {
             auto sm = symbol(sym);
             ulong tls_offset;
-            if (sm.obj->is_executable()) {
+            if (sm.obj->is_dynamically_linked_executable()) {
                 // If this is an executable (pie or position-dependant one)
                 // then the variable is located in the reserved slot of the TLS
                 // right where the kernel TLS lives
@@ -119,7 +119,7 @@ void object::arch_relocate_tls_desc(u32 sym, void *addr, Elf64_Sxword addend)
     ulong tls_offset;
     if (sym) {
         auto sm = symbol(sym);
-        if (sm.obj->is_executable() || sm.obj->is_core()) {
+        if (sm.obj->is_dynamically_linked_executable() || sm.obj->is_core()) {
             // If this is an executable (pie or position-dependant one)
             // then the variable is located in the reserved slot of the TLS
             // right where the kernel TLS lives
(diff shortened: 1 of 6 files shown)


nyh (approved):
I'll merge this because I think you know where you're headed, but I want to comment that I found some of the new names confusing. It's not exactly clear what "is_interpretable" means, and whereas "PIE" has a very specific meaning, it's not clear to me what "pic" is testing (one can compile any sort of object file with "-fpic").
I don't know if the solution should be better names or a comment on each of these functions saying exactly which kind of executables each one allows (that "is_interperable" is true for ordinary dynamiocally-linked executables, for PIE,  but not for static executables or whatever).

wkozaczuk:
Maybe `is_dynamically_linked_executable` is a better name than `is_interpretable`? Yes `is_pic` simply means if given ELF is a position-independent code (by detecting the presence of the `DYN`).

wkozaczuk:
Regarding your 2nd question, OSv would crash with this error if one tried to run statically linked non-PIE (aka position dependent):
```
Cmdline: /hello-static-non-pie
Statically linked executables are not supported!

[backtrace]
0x00000000402bbf7b <elf::object::process_headers()+1147>
0x00000000402c53d7 <elf::program::load_object(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, std::vector<std::shared_ptr<elf::object>, std::allocator<std::shared_ptr<elf::object> > >&)+2263>
...
```

but would stop and refuse to run a static PIE:
```
Cmdline: /hello-static-pie
Failed looking up main. Powering off.
```

After this patch, it consistently crashes with the 1st message.

Please note I am taking an incremental approach so even if one commented out this code in `process_headers`:
```c++
if (!is_core() && is_statically_linked()) {
        abort("Statically linked executables are not supported yet!\n");
    }
```

OSv would not be able to run those. I will remove these 3 lines if enough support of the statically linked ELF is in (entry_point, `arch_prctl`, TLS, more syscalls, etc).

However, all these somewhat minimal changes to the dynamic linker are necessary to support loading and processing statically linked executables once they are enabled all the way in.

wkozaczuk:
So I will update this MR with the name change and add some more comments.

wkozaczuk:
I have force pushed latest version.

nyh (approved):

PR #1254, opened 2023-09-05
core: support launching statically linked executables

### LINKED ISSUE #1140: Support loading and launching statically linked executables
That should be relatively easy and roughly match for these commits on the static-elf branch do:
- https://github.com/cloudius-systems/osv/commit/a2de1cada1dc97a6ef95ee10aa8a91867cb3e393
- https://github.com/cloudius-systems/osv/commit/fa19f90a851423d9b58a560657f6e789bb096b43
- https://github.com/cloudius-systems/osv/commit/642158a50b495cbb2657e4b86539d504339f5b76 (I think that most of it are implemented as we support position dependant dynamically-linked ELFs)
- https://github.com/cloudius-systems/osv/commit/463ca6528d6a31a4823a65fb0ba8b7e51f1f8cce
- https://github.com/cloudius-systems/osv/commit/cdc758b168355f492f951d82d9d15c339bfe21b9

Please note that implementing this is not enough to fully support the execution of statically linked executables. For that we need to implement other issues:
- #1137
- #1138 
- #1139

wkozaczuk:
This patch enhances the dynamic linker to support launching statically linked executables.

The dynamically linked executables or shared libraries are typically launched by calling a "main" function which is exported and can be resolved using the ELF symbol table. The statically linked executables do not export
such symbols and instead must be launched by jumping to the ELF entry point specified in the header. To that end this patch implements new functions - `run_entry_point()` - specific for each architecture - x86_64 and aarch64 which fundamentally do a similar thing - put `argc`, `argv`, environment variables, and auxiliary vector on the stack and jump to the elf entry point.

In addition, this patch enhances other parts of the dynamic linker logic to do things a little bit differently for statically linked executables:
- do not relocate as the executable comes with its own code that does this
- do not call INIT and FINI function for the same reason as above
- do not try to load dependent libraries as they would be none

This patch also adds augment_auxv() to add the extra auxiliary vector entries needed by statically linked executables to bootstrap themselves

In addition, this 2nd version of the PR renames the method `is_statically_linked()` to `is_statically_linked_executable()` and fixes it by identifying if an ELF is not a shared library. So it should fix the "liblua.so" problem reported by Nadav.

Closes #1140

### DIFF
--- a/arch/aarch64/arch-elf.hh
+++ b/arch/aarch64/arch-elf.hh
@@ -23,4 +23,38 @@ enum {
 
 #define ELF_KERNEL_MACHINE_TYPE 183
 
+static constexpr unsigned SAFETY_BUFFER = 256;
+#include <osv/align.hh>
+
+inline void run_entry_point(void* ep, int argc, char** argv, int argv_size)
+{
+    //The layout of the stack and state of all relevant registers is similar
+    //to how it looks for x86_64. The main difference (possibly for now)
+    //is the inlined assembly
+    int argc_plus_argv_stack_size = argv_size + 1;
+
+    //Capture current stack pointer
+    void *stack;
+    asm volatile ("mov %0, sp" : "=r"(stack));
+
+    //The code below puts argv and auxv vector onto the stack but it may
+    //also end up using some of the stack. To make sure there is no collision
+    //let us leave some space - SAFETY_BUFFER - between current stack pointer
+    //and the position on the stack we will be writing to.
+    stack -= (SAFETY_BUFFER + argc_plus_argv_stack_size * sizeof(char*));
+
+    //According to the document above the stack pointer should be 16-bytes aligned
+    stack = align_down(stack, 16);
+
+    *reinterpret_cast<u64*>(stack) = argc;
+    memcpy(stack + sizeof(char*), argv, argv_size * sizeof(char*));
+
+    //Set stack pointer and jump to the ELF entry point
+    asm volatile (
+        "mov sp, %1\n\t" //set stack
+        "blr %0\n\t"
+        :
+        : "r"(ep), "r"(stack));
+}
+
 #endif /* ARCH_ELF_HH */
--- a/arch/x64/arch-elf.hh
+++ b/arch/x64/arch-elf.hh
@@ -42,4 +42,45 @@ enum {
 
 #define ELF_KERNEL_MACHINE_TYPE 62
 
+static constexpr unsigned SAFETY_BUFFER = 256;
+#include <osv/align.hh>
+
+inline void run_entry_point(void* ep, int argc, char** argv, int argv_size)
+{
+    //The layout of the stack and state of all relevant registers is described
+    //in detail in the section 3.4 (Process Initialization) of the System V Application
+    //Binary Interface AMD64 Architecture Processor Supplement Draft Version 0.95
+    //(see https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.95.pdf)
+    int argc_plus_argv_stack_size = argv_size + 1;
+
+    //Capture current stack pointer
+    void *stack;
+    asm volatile ("movq %%rsp, %0" : "=r"(stack));
+
+    //The code below puts argv and auxv vector onto the stack but it may
+    //also end up using some of the stack. To make sure there is no collision
+    //let us leave some space - SAFETY_BUFFER - between current stack pointer
+    //and the position on the stack we will be writing to.
+    stack -= (SAFETY_BUFFER + argc_plus_argv_stack_size * sizeof(char*));
+
+    //According to the document above the stack pointer should be 16-bytes aligned
+    stack = align_down(stack, 16);
+
+    //... and it should start with argc, followed by argv, environment pointers and
+    //auxiliary vector entries. For details look at application::prepare_argv()
+    *reinterpret_cast<u64*>(stack) = argc;
+    memcpy(stack + sizeof(char*), argv, argv_size * sizeof(char*));
+
+    //TODO: Reset SSE2 and floating point registers and RFLAGS as the "Special Registers"
+    //      paragraph of the section of 3.4 (Process Initialization) of the "System V Application
+    //      Binary Interface" document states
+    //Set stack pointer, reset rdx and jump to the ELF entry point
+    asm volatile (
+        "movq %1, %%rsp\n\t" //set stack
+        "movq $0, %%rdx\n\t" //fini should be 0 for now (TODO: Eventually point to atexit())
+        "jmpq *%0\n\t"
+        :
+        : "r"(ep), "r"(stack));
+}
+
 #endif /* ARCH_ELF_HH */
--- a/core/app.cc
+++ b/core/app.cc
@@ -217,12 +217,19 @@ application::application(const std::string& command,
         throw launch_error("Failed to load object: " + command);
     }
 
-    _main = _lib->lookup<int (int, char**)>(main_function_name.c_str());
-    if (!_main) {
-        _entry_point = reinterpret_cast<void(*)()>(_lib->entry_point());
-    }
-    if (!_entry_point && !_main) {
-        throw launch_error("Failed looking up main");
+    if (_lib->is_statically_linked_executable()) {
+        //Augment auxiliary vector with extra entries like AT_PHDR, AT_ENTRY, etc
+        //that are necessary by a static executable to bootstrap itself
+        augment_auxv();
+    } else {
+        _main = _lib->lookup<int (int, char**)>(main_function_name.c_str());
+
+        if (!_main) {
+            _entry_point = reinterpret_cast<void(*)()>(_lib->entry_point());
+        }
+        if (!_entry_point && !_main) {
+            throw launch_error("Failed looking up main");
+        }
     }
 }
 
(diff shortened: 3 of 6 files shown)


arch/aarch64/arch-elf.hh
```diff
@@ -23,4 +23,31 @@ enum {
 
 #define ELF_KERNEL_MACHINE_TYPE 183
 
+#define SAFETY_BUFFER 256
+#include <osv/align.hh>
+
+inline void elf_entry_point(void* ep, int argc, char** argv, int argv_size)
```
nyh, line 29: Maybe we should call it run_entry_point? or jump_to_entry_point? Or something? The name "elf_entry_point" makes it sound like it calculates the entry point - which it doesn't - it takes one.

arch/x64/arch-elf.hh
```diff
@@ -42,4 +42,37 @@ enum {
 
 #define ELF_KERNEL_MACHINE_TYPE 62
 
+#define SAFETY_BUFFER 256
```
nyh, line 45: If this macro isn't for **users** of elf.hh but just for the internal implementation below, please make it a constexpr inside the function instead of a macro, so it doesn't "leak" outside this implementation.

arch/x64/arch-elf.hh
```diff
@@ -42,4 +42,37 @@ enum {
...
+    //Binary Interface AMD64 Architecture Processor Supplement Draft Version 0.95
+    //(see https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.95.pdf)
+    int argc_plus_argv_stack_size = argv_size + 1;
+
+    void *stack;
+    asm volatile ("movq %%rsp, %0" : "=r"(stack));
+    //
+    //According to the document above the stack pointer should be 16-bytes aligned
+    stack -= (SAFETY_BUFFER + argc_plus_argv_stack_size * sizeof(char*));
+    stack = align_down(stack, 16);
```
nyh, line 61: What is this "SAFETY BUFFER" thing? is this the red zone? The "red zone" is only 128 bytes, not 256.
Check out arch/x64/signal.cc:
```c++
    if (!rsp) {
        rsp = reinterpret_cast<void*>(ef->rsp);
        rsp -= 128;                 // skip red zone
    }
    rsp -= sizeof(signal_frame);
    // the Linux x86_64 calling conventions want 16-byte aligned rsp.
    // signal_frame may want even stricter alignment (but probably won't).
    rsp = align_down(rsp, std::max(16UL, alignof(signal_frame)));
```

arch/x64/arch-elf.hh
```diff
@@ -42,4 +42,37 @@ enum {
...
+
+    //... and it should start with argc, followed by argv, environment pointers and
+    //auxiliary vector entries. For details look at application::prepare_argv()
+    *reinterpret_cast<u64*>(stack) = argc;
+    memcpy(stack + sizeof(char*), argv, argv_size * sizeof(char*));
+
+    //TODO: Reset SSE2 and floating point registers and RFLAGS
```
nyh, line 68: Why do we need to reset anything regarding the floating point? Won't the executable do whatever it wants with it?

core/app.cc
```diff
@@ -217,12 +217,19 @@ application::application(const std::string& command,
         throw launch_error("Failed to load object: " + command);
     }
 
-    _main = _lib->lookup<int (int, char**)>(main_function_name.c_str());
-    if (!_main) {
-        _entry_point = reinterpret_cast<void(*)()>(_lib->entry_point());
-    }
-    if (!_entry_point && !_main) {
-        throw launch_error("Failed looking up main");
+    if (_lib->is_statically_linked()) {
```
nyh, line 220: Following my observation on liblua.so that something can be both a shared library with main() (this was the classic way to build stuff for OSv...) and have a broken, silly, I wonder if it doesn't make more sense to first check if we can find a main(), and only if we can't, fall back to using the entry point.
In fact, that's what the original code below did - I presume for PIEs - it first tried to find main() and only if it couldn't it set  _entry_point. Why can't or shouldn't we do the same for statically linked executable? Isn't the _entry_point calculation the same?

core/app.cc
```diff
@@ -319,22 +326,27 @@ void application::main()
     elf::get_program()->init_library(_args.size(), _argv.get());
     sched::thread::current()->set_name(_command);
 
-    if (_main) {
-        run_main();
+    if (_lib->is_statically_linked()) {
+        elf_entry_point(_lib->entry_point(), _args.size(), _argv.get(), _argv_size);
```
nyh, line 330: Continuing my previous comment, if the code in the contructor had set _entry_point like it did for the PIE case, you could have used _entry_point here instead of _lib->entry_point().

core/app.cc
```diff
@@ -394,25 +406,42 @@ void application::prepare_argv(elf::program *program)
...
+    _auxv[_auxv_idx++].a_un.a_val = sysconf(_SC_MINSIGSTKSZ);
+
+    _auxv[_auxv_idx].a_type = AT_RANDOM;
+    _auxv[_auxv_idx++].a_un.a_val = reinterpret_cast<uint64_t>(random_bytes);
+
+    _auxv[_auxv_idx].a_type = AT_NULL;
+    _auxv[_auxv_idx].a_un.a_val = 0;
+}
+
+void application::augment_auxv()
```
nyh, line 429: Can you please add a comment somewhere when/where/why this "augment" thing is needed?

core/app.cc
```diff
@@ -394,25 +406,42 @@ void application::prepare_argv(elf::program *program)
...
+{
+    _auxv[_auxv_idx].a_type = AT_PHDR;
```
nyh, line 431: Can you maybe assert() that there is enough room for the 4 items you want to add, from _auxv_idx until max_auxv_parameters_count?
Otherwise (e.g., if this function is called twice) we can have silent corruption.

core/elf.cc
```diff
@@ -372,9 +372,11 @@ void object::set_base(void* base)
         // Override the passed in value as the base for non-PICs (Position Dependant Executables)
         // needs to be set to 0 because all the addresses in it are absolute
         _base = 0x0;
+        _headers_start = reinterpret_cast<void*>(p->p_vaddr) + _ehdr.e_phoff;
     } else {
         // Otherwise for kernel, PIEs and shared libraries set the base as requested by caller
         _base = align(base, p->p_align, p->p_vaddr & (p->p_align - 1)) - p->p_vaddr;
+        _headers_start = _base + _ehdr.e_phoff;
```
nyh, line 379: I couldn't find where you use this new _headers_start variable.
On the other hand, I see there are existing places in the code which use e_phoff, should they use this new variable, or not?

core/elf.cc
```diff
@@ -1181,6 +1169,9 @@ std::string object::pathname()
 // Run the object's static constructors or similar initialization
 void object::run_init_funcs(int argc, char** argv)
 {
+    if (is_statically_linked()) {
+        return;
+    }
```
nyh, line 1174: Nitpick: you already if'ed around the call to run_init_funcs(), so no need to add an if() also inside the function?

nyh (changes requested):

arch/aarch64/arch-elf.hh
wkozaczuk: Good idea!

arch/x64/arch-elf.hh
wkozaczuk: Because this is what the spec https://refspecs.linuxfoundation.org/elf/x86_64-abi-0.95.pdf says:
"3.4 Process Initialization
3.4.1 Initial Stack and Register State
Special Registers
The AMD64 architecture defines floating point instructions. At process startup
the two floating point units, SSE2 and x87, both have all floating-point exception
status flags cleared. The status of the control words is as defined in tables 3.3 and
3.4.
Table 3.3: x87 Floating-Point Control Word
Field Value Note
RC 0 Round to nearest
PC 11 Double extended precision
PM 1 Precision masked
UM 1 Underflow masked
OM 1 Overflow masked
ZM 1 Zero divide masked
DM 1 Denormal operand masked
IM 1 Invalid operation masked
"

core/app.cc
wkozaczuk: Yeah, I need to think it through how to best address it.

core/elf.cc
wkozaczuk: After looking at this again I do not think I "if" around the call to `run_init_funcs`, do I?

arch/aarch64/arch-elf.hh
wkozaczuk: Addressed in the updated PR.

arch/x64/arch-elf.hh
wkozaczuk: Addressed in the updated PR.

arch/x64/arch-elf.hh
wkozaczuk: Addressed by better comments in the latest version of this PR.

core/app.cc
wkozaczuk: Addressed by better names and comments and some fixes in the latest version of this PR.

core/app.cc
wkozaczuk: Addressed by better comments in the latest version of this PR.

core/app.cc
wkozaczuk: Addressed by better comments in the latest version of this PR.

core/app.cc
wkozaczuk: Addressed in the latest version of this PR.

arch/x64/arch-elf.hh
wkozaczuk: I put better comments.

PR #1260, opened 2023-09-16
Implement arch_prctl syscall to support statically linked executables

### LINKED ISSUE #1137: Implement subset of arch_prctl syscall to support local-exec TLS of statically linked executables
The Golang apps like our `golang-pie-example` built with newer Golang compiler/linker (at least as of 1.15) no longer run correctly on OSv and crash like so:
```
OSv v0.55.0-247-g1433c08b
eth0: 192.168.122.15
Booted up in 162.61 ms
Cmdline: /hello.so
syscall(): unimplemented system call 158
Aborted

[backtrace]
0x00000000402178f7 <???+1075935479>
0x000000004046369e <osv::handle_mmap_fault(unsigned long, int, exception_frame*)+30>
0x0000000040342a1a <mmu::vm_fault(unsigned long, exception_frame*)+298>
0x0000000040390d5f <page_fault+143>
0x000000004038fc16 <???+1077476374>
0x0000100000066c9f <???+421023>
0x000000004045f2a5 <???+1078325925>
0x00000000403fbd89 <thread_main_c+41>
0x0000000040390b92 <???+1077480338>
```

The unimplemented syscall 158 is `__NR_arch_prctl` which can be used to set FS_BASE for local-exec TLS. It looks like something had changed in the Golang build-chain between 1.12 and 1.15. The Golang apps built with  ` -buildmode=pie` used to work just fine on OSv. It seems that in the past the Golang produced pies used to use two ways of communicating with kernel - 1) via SYSCALL instruction and 2) via libc calls for pthreads (OSv still does not support the **clone** syscall). The newer build chain seems to produce an executable that only communicates via SYSCALL.

Here is some info about the executable:
```
file apps/golang-pie-example/hello.so 
apps/golang-pie-example/hello.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, Go BuildID=KLddGH5JkygDOUUjd9N8/Bw6msSNbQDLVNmZHbfg9/N-nQvH6Ne9pl5aExWhga/Ncb0ercFgAfBxICpTZba, not stripped

ldd apps/golang-pie-example/hello.so
	statically linked

readelf -l apps/golang-pie-example/hello.so 

Elf file type is DYN (Shared object file)
Entry point 0x465080
There are 12 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000400040 0x0000000000400040
                 0x00000000000002a0 0x00000000000002a0  R      0x1000
  INTERP         0x0000000000000fe4 0x0000000000400fe4 0x0000000000400fe4
                 0x000000000000001c 0x000000000000001c  R      0x1
      [Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]
  NOTE           0x0000000000000f80 0x0000000000400f80 0x0000000000400f80
                 0x0000000000000064 0x0000000000000064  R      0x4
  LOAD           0x0000000000000000 0x0000000000400000 0x0000000000400000
                 0x0000000000099ed0 0x0000000000099ed0  R E    0x1000
  LOAD           0x000000000009a000 0x000000000049a000 0x000000000049a000
                 0x0000000000072ce5 0x0000000000072ce5  R      0x1000
  LOAD           0x000000000010d000 0x000000000050d000 0x000000000050d000
                 0x0000000000086467 0x0000000000086467  RW     0x1000
  GNU_RELRO      0x000000000010d000 0x000000000050d000 0x000000000050d000
                 0x0000000000086467 0x0000000000086467         0x1000
  LOAD           0x0000000000194000 0x0000000000594000 0x0000000000594000
                 0x0000000000015a00 0x00000000000481a8  RW     0x1000
  DYNAMIC        0x0000000000194040 0x0000000000594040 0x0000000000594040
                 0x00000000000000b0 0x00000000000000b0  RW     0x8
  TLS            0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000008  R      0x8
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x8
  LOOS+0x5041580 0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000         0x8

readelf -s apps/golang-pie-example/hello-stripped.so 

Symbol table '.dynsym' contains 1 entry:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND 
```

The info seems to be a bit confusing as it suggests that the executable is position-dependent (the entry point) and is statically linked.

I have also experimented with basic C hello world app linked statically and here is some info:
```
 readelf -l apps/native-static-example/hello

Elf file type is EXEC (Executable file)
Entry point 0x401ce0
There are 8 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000000 0x0000000000400000 0x0000000000400000
                 0x0000000000000488 0x0000000000000488  R      0x1000
  LOAD           0x0000000000001000 0x0000000000401000 0x0000000000401000
                 0x0000000000081d5d 0x0000000000081d5d  R E    0x1000
  LOAD           0x0000000000083000 0x0000000000483000 0x0000000000483000
                 0x0000000000027038 0x0000000000027038  R      0x1000
  LOAD           0x00000000000ab000 0x00000000004ac000 0x00000000004ac000
                 0x0000000000005310 0x0000000000006b40  RW     0x1000
  NOTE           0x0000000000000200 0x0000000000400200 0x0000000000400200
                 0x0000000000000044 0x0000000000000044  R      0x4
  TLS            0x00000000000ab000 0x00000000004ac000 0x00000000004ac000
                 0x0000000000000020 0x0000000000000060  R      0x8
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x10
  GNU_RELRO      0x00000000000ab000 0x00000000004ac000 0x00000000004ac000
                 0x0000000000003000 0x0000000000003000  R      0x1

strace apps/native-static-example/hello
execve("apps/native-static-example/hello", ["apps/native-static-example/hello"], 0x7ffc5ad38b90 /* 44 vars */) = 0
arch_prctl(0x3001 /* ARCH_??? */, 0x7ffedba29df0) = -1 EINVAL (Invalid argument)
brk(NULL)                               = 0x1dfc000
brk(0x1dfcd80)                          = 0x1dfcd80
arch_prctl(ARCH_SET_FS, 0x1dfc380)      = 0
uname({sysname="Linux", nodename="fedora-mbpro", ...}) = 0
readlink("/proc/self/exe", "/home/wkozaczuk/projects/osv-mas"..., 4096) = 68
brk(0x1e1dd80)                          = 0x1e1dd80
brk(0x1e1e000)                          = 0x1e1e000
mprotect(0x4ac000, 12288, PROT_READ)    = 0
fstat(1, {st_mode=S_IFCHR|0620, st_rdev=makedev(0x88, 0x1), ...}) = 0
write(1, "Hello from static C code\n", 25Hello from static C code
) = 25
exit_group(0)                           = ?
+++ exited with 0 +++
``` 

So in each case, apps make the `arch_prctl` syscall which OSv would need to support.

It is not the first time the issue of `arch_prctl` is raised - please see this conversation - https://groups.google.com/g/osv-dev/c/qLfRBRRUIHM and this one related to one of the patches attempting to add the `arch_prctl` support - https://groups.google.com/g/osv-dev/c/FPkZzUb5uB4/m/npEg9GqDAgAJ. 

Here is what @nyh wrote in the latter one:

> In an earlier review in https://groups.google.com/forum/#!msg/osv-dev/PW3bkaVCuMg/340vWUeimdMJ
and also today in another thread, I pointed out the problems with this arch_prctl thing which makes this patch either incomplete or even counter-productive:

> 1. After a thread does ARCH_SET_FS, if it get switched out and then back in, thread::switch_to() will overwrite this setting with a fixed pointer "thread->_tcb". So to support ARCH_SET_FS we need to save and restore this value on context switch (or more efficiently, save it only on ARCH_SET_FS, not on every switch out, and restore on switch in).

> 2. After a thread does ARCH_SET_FS, it cannot call any OSv code (or any library function in general). We need to modify the system-call code to restore the standard fs_base (the _tcb) when entering the system call, and restore it back to the thread's chosen address when exiting the system call.

> 3. It's even worse if after a thread does ARCH_SET_FS is calls a function directly, not through a system call. e.g., malloc(). It is possible (???) that Go is aware of this problem and restores the FS_BASE before calling C functions, but if it doesn't, we're in big problems.

My sense is that adding support of `arch_prctl` will be a bit tricky but not impossible _if we stick to certain assumptions_:
* we would only support the syscall but not `arch_prctl` glibc wrapper on purpose
* we would only support `arch_prctl` with statically linked executables both position-dependent and independent ones only; the key thing is that the SYSCALL instruction would provide us with necessary trigger points to switch the `FS` register to point to correct TLS block for a given thread 

So here is a concrete list of things to be done:
* implement the `arch_prctl`  syscall to handle `ARCH_SET_FS` and `ARCH_GET_FS` and store the  `ARCH_SET_FS` parameter in a new field for the current thread 
* modify the syscall assembly to save FS register on entry and restore FS register on exit if there is one set by `ARCH_SET_FS` for current thread
* modify the page fault handler and possibly interrupt handler (assembly?) to save FS register on entry and restore FS register on exit if there is one set by `ARCH_SET_FS` for the current thread (only if app?)
* possibly change context switch code to save and restore FS register; this may not be necessary as this would be done by the above 2

wkozaczuk:
This PR implements the `arch_prctl` syscall and makes other changes to support switching the FS register between the application and kernel TCB when running statically linked executables.

In essence, it makes it possible to launch simple statically linked executables like "Hello World" on OSv:
```
gcc -static -o hello-static-non-pie hello.c

./scripts/run.py -e /hello-static-non-pie
OSv v0.57.0-69-g053fc914
eth0: 192.168.122.15
Booted up in 193.02 ms
Cmdline: /hello-static-pie
WARNING: Statically linked executables are only supported to limited extent!
syscall(): unimplemented system call 218
syscall(): unimplemented system call 273
syscall(): unimplemented system call 334
syscall(): unimplemented system call 302
syscall(): unimplemented system call 89
syscall(): unimplemented system call 10
Hello from C code
```

Please note, that the code changes touch some critical places of the kernel functionality - context switching, syscall handling, exception handling, and VDSO implementation. 

As far as context switching goes, this patch adds only a handful of memory read/write operations that do not seem to affect it in any measurable way based on what the `misc-ctxsw.cc` indicates.

On the other hand, one could see the syscall handling cost go up by 6 - 10 ns (6-10% of the total cost based on what `misc-syscall-perf.cc` measures) when executing statically linked executables due to the fact we need to switch the `fsbase` and turn interrupts off and on twice. The good news is that the syscall handling does not seem to be affected in any significant way when running dynamically linked executables.

Finally, I did not measure the impact of changes to the exception handling (interrupts, page faults, etc) but I think it should be lower than syscall handling given we do not need to turn interrupts on and off. Also, we should not see any impact when running dynamically linked executables.

Depends on https://github.com/cloudius-systems/osv/pull/1254
Closes #1137

### DIFF
--- a/libc/vdso/vdso.c
+++ b/libc/vdso/vdso.c
@@ -3,22 +3,25 @@
 #include <sys/time.h>
 
 #ifdef __x86_64__
+time_t __vdso_kernel_time(time_t *tloc);
 __attribute__((__visibility__("default")))
 time_t __vdso_time(time_t *tloc)
 {
-    return time(tloc);
+    return __vdso_kernel_time(tloc);
 }
 
+int __vdso_kernel_gettimeofday(struct timeval *tv, struct timezone *tz);
 __attribute__((__visibility__("default")))
 int __vdso_gettimeofday(struct timeval *tv, struct timezone *tz)
 {
-    return gettimeofday(tv, tz);
+    return __vdso_kernel_gettimeofday(tv, tz);
 }
 
+int __vdso_kernel_clock_gettime(clockid_t clk_id, struct timespec *tp);
 __attribute__((__visibility__("default")))
 int __vdso_clock_gettime(clockid_t clk_id, struct timespec *tp)
 {
-    return clock_gettime(clk_id, tp);
+    return __vdso_kernel_clock_gettime(clk_id, tp);
 }
 #endif
 
--- a/Makefile
+++ b/Makefile
@@ -1012,6 +1012,7 @@ objects += arch/x64/ioapic.o
 objects += arch/x64/apic.o
 objects += arch/x64/apic-clock.o
 objects += arch/x64/entry-xen.o
+objects += arch/x64/prctl.o
 objects += arch/x64/vmlinux.o
 objects += arch/x64/vmlinux-boot64.o
 objects += arch/x64/pvh-boot.o
@@ -1849,6 +1850,11 @@ musl += time/timegm.o
 musl += time/wcsftime.o
 musl += time/ftime.o
 $(out)/libc/time/ftime.o: CFLAGS += -Ilibc/include
+
+ifeq ($(arch),x64)
+libc += vdso/vdso_kernel.o
+libc_to_hide += vdso/vdso_kernel.o
+endif
 
 musl += termios/tcflow.o
 
--- a/arch/aarch64/arch-switch.hh
+++ b/arch/aarch64/arch-switch.hh
@@ -164,6 +164,10 @@ void thread::free_tcb()
     free(_tcb);
 }
 
+void thread::free_syscall_stack()
+{
+}
+
 void thread_main_c(thread* t)
 {
     arch::irq_enable();
--- a/arch/x64/arch-cpu.hh
+++ b/arch/x64/arch-cpu.hh
@@ -13,6 +13,8 @@
 #include "cpuid.hh"
 #include "osv/pagealloc.hh"
 #include <xmmintrin.h>
+#include "syscall.hh"
+#include "msr.hh"
 
 struct init_stack {
     char stack[4096] __attribute__((aligned(16)));
@@ -46,6 +48,7 @@ struct arch_cpu {
     u32 apic_id;
     u32 acpi_id;
     u64 gdt[nr_gdt];
+    syscall_stack _current_syscall_stack;
     void init_on_cpu();
     void set_ist_entry(unsigned ist, char* base, size_t size);
     char* get_ist_entry(unsigned ist);
@@ -181,6 +184,8 @@ inline void arch_cpu::init_on_cpu()
     processor::init_fpu();
 
     processor::init_syscall();
+
+    processor::wrmsr(msr::IA32_GS_BASE, reinterpret_cast<u64>(&_current_syscall_stack.stack_top));
 }
 
 struct exception_guard {
--- a/arch/x64/arch-cpu.hh
+++ b/arch/x64/arch-cpu.hh
@@ -49,6 +49,7 @@ struct arch_cpu {
     u32 acpi_id;
     u64 gdt[nr_gdt];
     syscall_stack _current_syscall_stack;
+    u64 kernel_tcb;
     void init_on_cpu();
     void set_ist_entry(unsigned ist, char* base, size_t size);
     char* get_ist_entry(unsigned ist);
--- a/arch/x64/arch-switch.hh
+++ b/arch/x64/arch-switch.hh
@@ -11,6 +11,7 @@
 #include "msr.hh"
 #include <osv/barrier.hh>
 #include <string.h>
+#include "tls-switch.hh"
 
 //
 // The last 16 bytes of the syscall stack are reserved for -
@@ -64,12 +65,15 @@ void set_fsbase_fsgsbase(u64 v)
     processor::wrfsbase(v);
 }
 
+bool fsgsbase_avail = false;
+
 extern "C"
 void (*resolve_set_fsbase(void))(u64 v)
 {
     // can't use processor::features, because it is not initialized
     // early enough.
     if (processor::features().fsgsbase) {
+        fsgsbase_avail = true;
         return set_fsbase_fsgsbase;
     } else {
         return set_fsbase_msr;
@@ -93,6 +97,7 @@ void thread::switch_to()
     c->arch.set_exception_stack(_state.exception_stack);
     c->arch._current_syscall_stack.caller_stack_pointer = _state._syscall_stack.caller_stack_pointer;
     c->arch._current_syscall_stack.stack_top = _state._syscall_stack.stack_top;
+    c->arch.kernel_tcb = reinterpret_cast<u64>(_tcb);
     auto fpucw = processor::fnstcw();
     auto mxcsr = processor::stmxcsr();
     asm volatile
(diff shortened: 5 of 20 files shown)


libc/vdso/vdso.c
```diff
@@ -3,22 +3,25 @@
 #include <sys/time.h>
 
 #ifdef __x86_64__
+time_t __vdso_kernel_time(time_t *tloc);
 __attribute__((__visibility__("default")))
 time_t __vdso_time(time_t *tloc)
 {
-    return time(tloc);
+    return __vdso_kernel_time(tloc);
```
nyh, line 10: Won't it be clearer to just add the `arch::tls_switch_on_app_stack` here? What's the benefit of having another source file, another set of functions? Also the extra call isn't free.

nyh (commented):
It's hard for me to "approve" or "request changes" in this - I'm pretty rusty in this area, and I'm not sure I understood all the details. I left some comments where I either thought I understood - or where I thought I didn't understand. In any case, nice work.

wkozaczuk:
Closing this one in lieu of the new one #1267 and another future one with an updated code to support switching between app and kernel TCB (the 2nd commit of this PR).

nyh:
> Closing this one in lieu of the new one #1267 and another future one with an updated code to support switching between app and kernel TCB (the 2nd commit of this PR).

Please take a look at my comments above, if they are still relevant in your new version.

wkozaczuk:
Yes, I have addressed your comments in the new version.

PR #1265, opened 2023-10-03
Add new syscalls needed to run statically linked executables

wkozaczuk:


linux.cc
```diff
@@ -396,6 +397,22 @@ static long sys_getcpu(unsigned int *cpu, unsigned int *node, void *tcache)
     return 0;
 }
 
+#define __NR_sys_prlimit __NR_prlimit64
+static int sys_prlimit(pid_t pid, int resource, const struct rlimit *new_limit,
+    struct rlimit *old_limit)
+{
+    if (new_limit) {
+        return 0;
```
nyh, line 412: Why return 0? If the new limits are ignored maybe it should return EINVAL or EPERM or something? And at least warn_once that it's unimplemented?

linux.cc
```diff
@@ -396,6 +397,22 @@ static long sys_getcpu(unsigned int *cpu, unsigned int *node, void *tcache)
     return 0;
 }
 
+#define __NR_sys_prlimit __NR_prlimit64
+static int sys_prlimit(pid_t pid, int resource, const struct rlimit *new_limit,
```
nyh, line 408: Shouldn't you call this function prlimit() and publish it? Does the Linux system call differ from the glibc function?

linux.cc
```diff
@@ -396,6 +397,22 @@ static long sys_getcpu(unsigned int *cpu, unsigned int *node, void *tcache)
...
+    struct rlimit *old_limit)
+{
+    if (new_limit) {
+        return 0;
+    }
+
+    if (old_limit) {
+        return getrlimit(resource, old_limit);
```
nyh, line 416: Probably need to verify first that "pid" is 0 or the current pid or something, and if not return an error (?)

linux.cc
wkozaczuk: Will fix it in a new PR.

linux.cc
wkozaczuk: Will fix it in a new PR.

linux.cc
wkozaczuk: Will fix it in a new PR.

wkozaczuk:
I am closing this PR and will create new one for the `prlimit` syscall and address all review issues.

PR #1284, opened 2023-12-09
Implement nvme driver

JanBraunwarth:
I implemented a nvme driver for OSv as part of my bachelor thesis and like to share the results here.
I don't think it's ready for merging, since the performance is far from the theoretical possible iops compared to an ubuntu vm. But i hope it still can serve as a base for future implementations.

The patch also fixes a simple bug that prevented addressing data with an offset greater than 2GB through the block cache. Cause: the block offset was saved in an int and later used to reconstruct the byte offset. The simple fix in buf.h should be integrated, even if the rest of the patch isn't.

I tested with fio and the io-test included in the patch; on qemu emulated and via pci-passthrough. On real hardware interrupts lead to a noticeable overhead, resulting in better iops on the qemu emulated nvme. Interrupt coalescing reduces the amount of interrupts which results in similar iops but this can only be seen as work-arround. I'm not sure what goes wrong here. 

Another factor in the disappointing performance is the block cache. Splitting every read/write into 512 byte-sized requests is  not advisable. Most modern nvmes internally use 4k blocks and splitting into 512 byte blocks will result in 8+ times worse iops. Especially since all the resulting requests get done one after another. I included a read/write that doesn't use the block cache and doesn't copy the data; therefor does not work on stack addresses (virt_to_phys fails on these).

### DIFF
--- a/drivers/nvme-queue.cc
+++ b/drivers/nvme-queue.cc
@@ -1,0 +1,464 @@
+#include <sys/cdefs.h>
+
+#include "nvme-queue.hh"
+#include <osv/bio.h>
+#include <vector>
+#include <memory>
+#include <osv/contiguous_alloc.hh>
+
+extern std::unique_ptr<nvme_sq_entry_t> alloc_cmd();
+
+TRACEPOINT(trace_nvme_io_queue_wake, "nvme%d qid=%d", int, int);
+TRACEPOINT(trace_nvme_wait_for_completion_queue_entries, "nvme%d qid=%d,have_elements=%d", int, int, bool);
+TRACEPOINT(trace_nvme_completion_queue_not_empty, "nvme%d qid=%d,not_empty=%d", int, int, bool);
+TRACEPOINT(trace_nvme_enable_interrupts, "nvme%d qid=%d", int, int);
+TRACEPOINT(trace_nvme_disable_interrupts, "nvme%d qid=%d", int, int);
+
+TRACEPOINT(trace_nvme_read, "nvme%d qid=%d cid=%d, bio data=%#x, slba=%d, nlb=%d", int, int , u16, void*, u64, u32);
+TRACEPOINT(trace_nvme_write, "nvme%d qid=%d cid=%d, bio data=%#x, slba=%d, nlb=%d", int, int , u16, void*, u64, u32);
+
+TRACEPOINT(trace_nvme_req_done_error, "nvme%d qid=%d, cid=%d, status type=%#x, status code=%#x, bio=%#x", int, int, u16, u8, u8, bio*);
+TRACEPOINT(trace_nvme_req_done_success, "nvme%d qid=%d, cid=%d, bio=%#x",int,int, u16, bio*);
+
+TRACEPOINT(trace_nvme_admin_queue_wake, "nvme%d qid=%d",int,int);
+
+TRACEPOINT(trace_nvme_admin_queue_submit, "nvme%d qid=%d, cid=%d",int, int, int);
+TRACEPOINT(trace_nvme_admin_req_done_error, "nvme%d qid=%d, cid=%d, status type=%#x, status code=%#x", int, int, u16, u8, u8);
+TRACEPOINT(trace_nvme_admin_req_done_success, "nvme%d qid=%d, cid=%d",int,int, u16);
+
+TRACEPOINT(trace_advance_sq_tail_full, "nvme%d qid=%d, sq_tail=%d, sq_head=%d", int, int, int, int);
+TRACEPOINT(trace_nvme_wait_for_entry, "nvme%d qid=%d, sq_tail=%d, sq_head=%d", int, int, int, int);
+
+nvme_queue_pair::nvme_queue_pair(
+    int did,
+    u32 id,
+    int qsize,
+    pci::device &dev,
+    nvme_sq_entry_t* sq_addr,
+    u32* sq_doorbell,
+    nvme_cq_entry_t* cq_addr,
+    u32* cq_doorbell,
+    std::map<u32, nvme_ns_t*>& ns)
+          : _id(id)
+          ,_driverid(did)
+          ,_qsize(qsize)
+          ,_dev(&dev)
+          ,_sq_addr(sq_addr)
+          ,_sq_head(0)
+          ,_sq_tail(0)
+          ,_sq_doorbell(sq_doorbell)
+          ,_sq_full(false)
+          ,_cq_addr(cq_addr)
+          ,_cq_head(0)
+          ,_cq_tail(0)
+          ,_cq_doorbell(cq_doorbell)
+          ,_cq_phase_tag(1)
+          ,_ns(ns)
+
+{
+    auto prplists = (u64**) malloc(sizeof(u64*)*qsize);
+    memset(prplists,0,sizeof(u64*)*qsize);
+    _prplists_in_use.push_back(prplists);
+
+    assert(!completion_queue_not_empty());
+}
+
+nvme_queue_pair::~nvme_queue_pair()
+{
+    memory::free_phys_contiguous_aligned(_sq_addr);
+    memory::free_phys_contiguous_aligned(_cq_addr);
+    for(auto vec: _prplists_in_use)
+        memory::free_phys_contiguous_aligned(vec);
+}
+
+inline void nvme_queue_pair::advance_sq_tail()
+{
+    _sq_tail = (_sq_tail + 1) % _qsize;
+    if(_sq_tail == _sq_head) {
+        _sq_full = true; 
+        trace_advance_sq_tail_full(_driverid,_id,_sq_tail,_sq_head);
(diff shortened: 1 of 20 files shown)


drivers/nvme-queue.cc
```diff
@@ -0,0 +1,464 @@
...
+    return cqe;
+}
+
+int nvme_io_queue_pair::make_request(bio* bio, u32 nsid=1)
+{
+    u64 slba = bio->bio_offset;
+    u32 nlb = bio->bio_bcount; //do the blockshift in nvme_driver
+    u16 cid;
+    
+    _lock.lock();
```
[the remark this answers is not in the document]
wkozaczuk: I assume we need to lock because there is a single pair of queues. I wonder if we can avoid locking if we use multi-queue approach and have a pair per cpu.

drivers/nvme-queue.cc
```diff
@@ -0,0 +1,464 @@
...
+        
+        /*Wake up the thread that requested the admin command*/
+        new_cq = true;
+        _req_waiter.wake_from_kernel_or_with_irq_disabled();
+    }
+}
+
+std::unique_ptr<nvme_cq_entry_t> nvme_admin_queue_pair::submit_and_return_on_completion(std::unique_ptr<nvme_sq_entry_t> cmd, void* data, unsigned int datasize)
+{
+    _lock.lock();
```
wkozaczuk, line 377: SCOPE _LOCK ?

drivers/nvme-queue.hh
```diff
@@ -0,0 +1,126 @@
...
+        u32* cq_doorbell,
+        std::map<u32, nvme_ns_t*>& ns
+    );
+
+    std::unique_ptr<nvme_cq_entry_t> _req_res;
+    volatile bool new_cq;
+    void req_done();
+    std::unique_ptr<nvme_cq_entry_t> submit_and_return_on_completion(std::unique_ptr<nvme_sq_entry_t> cmd, void* data=nullptr, unsigned int datasize=0);
+private:
+    sched::thread_handle _req_waiter;
```
wkozaczuk, line 123: Why do we need this extra waiter on top of the one in the base class?

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+
+static int
+nvme_write(struct device *dev, struct uio *uio, int ioflags)
+{
+    return bdev_write(dev, uio, ioflags);
+}
+
+static int
+nvme_direct_rw(struct device *dev, struct uio *uio, int ioflags)
+{
```
wkozaczuk, line 75: Nitpick: some indentation problem.

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+        create_io_queues_foreach_cpu();
+    }else {
+        u16 ret;
+        set_number_of_queues(1, &ret);
+        assert(ret>=1);
+        create_io_queue();
+    }
+
+    if(_identify_controller->vid != QEMU_VID) {
+        set_interrupt_coalescing(20,2);
```
wkozaczuk, line 207: What do these numbers mean? Can we have some consts for these?

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+        prv->drv = this;
+        prv->nsid = ns.first;
+        osv_dev->size = size;
+        /*
+        * IO size greater than 4096 << 9 would mean we need 
+        * more than 1 page for the prplist which is not implemented
+        */
+        osv_dev->max_io_size = 4096 << ((9 < _identify_controller->mdts)? 9 : _identify_controller->mdts );
+
+        #if CONF_drivers_io_test
```
wkozaczuk, line 241: What is the motivation behind running this test during the driver initialization stage?

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+    }
+    NVME_ERROR("timeout=%d waiting for ready %d", timeout, ready);
+    return ETIME;
+}
+
+void nvme::init_controller_config()
+{
+    nvme_controller_config_t cc;
+    cc.val = mmio_getl(&_control_reg->cc.val);
+    cc.iocqes    = 4;  // completion queue entry size 16B
```
wkozaczuk, line 330: Do these comments match the values? Can we have consts here?

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+    register_interrupt(iv,id-1);
+
+    _admin_queue->submit_and_return_on_completion(std::unique_ptr<nvme_sq_entry_t>((nvme_sq_entry_t*)cmd));
+    _admin_queue->submit_and_return_on_completion(std::unique_ptr<nvme_sq_entry_t>((nvme_sq_entry_t*)cmd_sq));
+
+    return id -1;
+}
+
+void nvme::create_io_queues_foreach_cpu()
+{
```
wkozaczuk, line 414: I see here a potential for a lot of code re-use between `create_io_queue()` and `create_io_queues_foreach_cpu()`. Can we have some helper method to implement common logic?

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+    auto cmd = alloc_cmd();
+    cmd->identify.cns = 2;
+    cmd->identify.common.nsid = start - 1;
+    cmd->identify.common.opc = NVME_ACMD_IDENTIFY;
+    auto active_namespaces = (u64*) alloc_phys_contiguous_aligned(4096, 4);
+    memset(active_namespaces, 0, 4096);
+
+    _admin_queue->submit_and_return_on_completion(std::move(cmd), (void*) mmu::virt_to_phys(active_namespaces), 4096);
+    int err;
+    for(int i=0; i < 1024; i++) {
```
wkozaczuk, line 532: Why 1024? Can we have a const for this?

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+    
+    if(bio->bio_cmd == BIO_FLUSH && (_identify_controller->vwc == 0 || !NVME_VWC_ENABLED )) {
+        biodone(bio,true);
+        return 0;
+    }
+
+    if(sched::current_cpu->id >= _io_queues.size())
+        return _io_queues[0]->make_request(bio, nsid);
+
+    return _io_queues[sched::current_cpu->id]->make_request(bio, nsid);
```
wkozaczuk, line 564: In this per-cpu queue setup do we have to lock in `make_request()`? Do we handle scenarios when we have fewer or more queues than cpus? Should we do something like modulo `sched::cpus.size()`?

drivers/nvme.cc
```diff
@@ -0,0 +1,693 @@
...
+        sched::thread::attr().name("nvme"+ std::to_string(_id)+"_aq_req_done"));
+    aq_thread->start();
+        
+    bool ok = msix_register(0, [this] { this->_admin_queue->disable_interrupts(); }, aq_thread);
+    _dev.msix_unmask_entry(0);
+    if(not ok)
+        printf("admin interrupt registration failed\n");
+}
+
+bool nvme::msix_register(unsigned iv,
```
wkozaczuk, line 579: This code seems to have a lot of overlap with `arch/x64/msi.cc`. Any reason not to re-use it?

wkozaczuk (changes requested):
I have left some comments and questions. Nothing major.

drivers/nvme-queue.cc
```diff
@@ -0,0 +1,464 @@
...
+}
+
+void nvme_queue_pair::disable_interrupts()
+{
+    _dev->msix_mask_entry(_id);
+    trace_nvme_disable_interrupts(_driverid,_id);
+}
+
+//only use with interrupts disabled
+std::unique_ptr<nvme_cq_entry_t> nvme_queue_pair::check_for_completion(u16 cid)
```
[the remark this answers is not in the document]
JanBraunwarth: The function isn't used currently. The function polls for cqe instead of relying on interrupts. I mostly used this while checking if the interrupts work correctly. 
I left it in since it might be useful to have an alternative to msix interrupts in case something goes wrong

drivers/nvme-queue.hh
JanBraunwarth: The waiter on the base class is used to wake up a thread that waits for a unused submission queue entry because the submission queue was full before.

The additional waiter in the admin_queue_pair is used to wake up the requesting thread.
https://github.com/cloudius-systems/osv/blob/2b84047aa6393bf871c6ab9ffb87bd6160cc5294/drivers/nvme-queue.cc#L375

I could have reused the base class waiter because in the current implementation the admin_queue will never have more than one outstanding request (every request waits for completion) but I thought this might have been even more confusing and it might not be optimal that the admin_queue waits for every request to finish if we implement more admin commands. In this case we might need the base class waiter.

wkozaczuk:
I am closing this PR as another one based on it - https://github.com/cloudius-systems/osv/pull/1317 - has been merged into the master branch.

PR #1288, opened 2023-12-18
signals: switch to app TLS if available

### LINKED ISSUE #1278: Running signal handler that uses thread local variables in statically linked executables crashes
This can be reproduced by running `iperf3` and then doing Ctrl-C which causes this:
```
OSv v0.57.0-91-g98c90f03
eth0: 192.168.122.15
Booted up in 126.00 ms
Cmdline: /lib64/ld-linux-x86-64.so.2 /iperf3 -s
syscall(): unimplemented system call 334
warning: this system does not seem to support IPv6 - trying IPv4
-----------------------------------------------------------
Server listening on 5201
-----------------------------------------------------------
Aborted

[backtrace]
0x000000004022133d <???+1075974973>
0x00000000403eace2 <osv::handle_mmap_fault(unsigned long, int, exception_frame*)+34>
0x00000000402b950a <mmu::vm_fault(unsigned long, exception_frame*)+298>
0x000000004030bd53 <page_fault+147>
0x000000004030aa82 <???+1076931202>
```

Connecting to gdb yields this:
```
(gdb) bt
#0  0x0000000040312942 in processor::cli_hlt () at arch/x64/processor.hh:247
#3  0x0000000040245344 in abort (fmt=fmt@entry=0x405cd413 "Aborted\n") at runtime.cc:145
#4  0x0000000040202ee5 in abort () at runtime.cc:106
#5  0x000000004022133e in osv::generate_signal (siginfo=..., ef=0x40007fe83088) at libc/signal.cc:141
#6  0x00000000403eace3 in osv::handle_mmap_fault (addr=addr@entry=0, sig=sig@entry=11, ef=ef@entry=0x40007fe83088) at libc/signal.cc:156
#7  0x00000000402b950b in mmu::vm_sigsegv (ef=0x40007fe83088, addr=0) at core/mmu.cc:1428
#8  mmu::vm_fault (addr=0, addr@entry=1688, ef=ef@entry=0x40007fe83088) at core/mmu.cc:1456
#9  0x000000004030bd54 in page_fault (ef=0x40007fe83088) at arch/x64/mmu.cc:42
#10 <signal handler called>
#11 __GI___pthread_cleanup_upto (target=target@entry=0x200000205080 <sigend_jmp_buf>, targetframe=targetframe@entry=0x40007fe93fc8 "\005\"D\177") at ./nptl/pthread_cleanup_upto.c:32
#12 0x000020007f44232c in _longjmp_unwind (env=env@entry=0x200000205080 <sigend_jmp_buf>, val=val@entry=1) at ../sysdeps/nptl/jmp-unwind.c:27
#13 0x000020007f442205 in __libc_siglongjmp (env=0x200000205080 <sigend_jmp_buf>, val=1) at ../setjmp/longjmp.c:30
#14 0x0000200000202543 in sigend_handler (sig=2) at main.c:121
#15 0x000000004037f0ee in sched::thread::main (this=0x40007fe7e040) at core/sched.cc:1415
#16 sched::thread_main_c (t=0x40007fe7e040) at arch/x64/arch-switch.hh:377
#17 0x000000004030bc52 in thread_main () at arch/x64/entry.S:161
(gdb) osv info threads
  id  address             name            cpu  status         vruntime  total_cpu_time location
   1 (0x400000019040) reclaimer       cpu0 waiting     4.52364e-25             799 condvar::wait(lockfree::mutex*, sched::timer*) at core/condvar.cc:51
   2 (0x400000056040) kvm_wall_clock_ cpu0 waiting     9.36834e-22           12151 std::_Function_handler<void(), kvmclock::kvmclock()::<lambda()> >::_M_invoke(const std::_Any_data &) at /usr/include/c++/11/bits/std_function.h:290
   3 (0x400000074040) page_pool_l2    cpu0 waiting     8.80017e-22         1135716 memory::page_pool::l2::fill_thread() at core/mempool.cc:1496
   4 (0x400000094040) itimer-real     cpu? unstarted             0               0 ?? at arch/x64/entry.S:160
   5 (0x400000111040) itimer-virt     cpu? unstarted             0               0 ?? at arch/x64/entry.S:160
   6 (0x400000146040) balancer0       cpu0 waiting     2.23752e-21          108849 ?? at arch/x64/entry.S:161
   7 (0x40000015e040) rcu0            cpu0 waiting     9.56428e-25            1723 osv::rcu::cpu_quiescent_state_thread::do_work() at core/rcu.cc:181
   8 (0x400000177040) page_pool_l1_0  cpu0 waiting     8.98783e-23           42318 memory::page_pool::l1::fill_thread() at core/mempool.cc:1381
   9 (0x40000018f040) percpu0         cpu0 waiting     3.23117e-25             584 workman::call_of_duty() at core/percpu-worker.cc:92
  10 (0x4000001a7040) async_worker0   cpu0 waiting      4.3944e-25             760 async::async_worker::run() at core/async.cc:176
  11 (0x4000001bf040) idle0           cpu0 queued              inf      1015916699 std::_Function_handler<void(), sched::cpu::init_idle_thread()::<lambda()> >::_M_invoke(const std::_Any_data &) at /usr/include/c++/11/bits/std_function.h:290
  12 (0x400000a6e040) >init           cpu0 waiting     5.84196e-24             493 osv::acpi_interrupt::process_interrupts() at drivers/acpi.cc:250
  13 (0x400000a85040) >init           cpu0 waiting     5.97121e-24             689 std::_Function_handler<void(), sched::thread::reaper::reaper()::<lambda()> >::_M_invoke(const std::_Any_data &) at /usr/include/c++/11/bits/std_function.h:290
  14 (0x400000a9c040) thread taskq    cpu0 waiting     6.16508e-24            1019 synch_port::_msleep(void*, mtx*, int, char const*, int) at bsd/porting/synch.cc:102
  15 (0x400000aa7040) callout         cpu0 waiting     2.44065e-21          102382 _callout_thread() at bsd/porting/callout.cc:138
  16 (0x400000adc040) netisr          cpu0 waiting     1.64014e-23            2852 netisr_osv_thread_wrapper(void (*)(void*), void*) at bsd/sys/net/netisr1.cc:26
  17 (0x400000c4c040) >init           cpu0 waiting     5.45383e-20        79119367 sched::thread::join() at core/sched.cc:1562
	joining on <optimized out>
  18 (0x400000c66040) virtio-net-rx   cpu0 waiting     2.44023e-28           19225 virtio::virtio_driver::wait_for_queue(virtio::vring*, bool (virtio::vring::*)() const) at drivers/virtio.cc:158
  19 (0x400000c8e040) virtio-tx-0     cpu0 waiting     8.90519e-29             851 osv::xmitter<virtio::net::txq, 4096u, std::function<bool ()>, boost::iterators::function_output_iterator<osv::xmitter_functor<virtio::net::txq> > >::poll_until() at include/osv/percpu_xmit.hh:399
  20 (0x400000da9040) virtio-blk      cpu0 waiting     2.34544e-22          239819 virtio::virtio_driver::wait_for_queue(virtio::vring*, bool (virtio::vring::*)() const) at drivers/virtio.cc:158
  21 (0x400000dc0040) virtio-rng      cpu0 waiting     2.46307e-20          803801 condvar::wait(lockfree::mutex*, sched::timer*) at core/condvar.cc:51
  22 (0x400000dda040) isa-serial-inpu cpu0 waiting      3.5655e-20          251753 console::LineDiscipline::read_poll(console::console_driver*) at drivers/line-discipline.cc:86
  23 (0x400000df1040) kbd-input       cpu0 waiting     1.45144e-23            7330 console::LineDiscipline::read_poll(console::console_driver*) at drivers/line-discipline.cc:86
  24 (0x400000e13040) rand_harvestq   cpu0 waiting     1.80029e-20          681103 synch_port::_msleep(void*, mtx*, int, char const*, int) at bsd/porting/synch.cc:102
  25 (0x400000e2d040) dhcp            cpu0 waiting     7.69433e-22          899366 dhcp::dhcp_worker::dhcp_worker_fn() at core/dhcp.cc:828
  26 (0x400000f53040) /lib64/ld-linux cpu0 waiting     1.21231e-20        12121800 do_poll(std::vector<poll_file, std::allocator<poll_file> >&, boost::optional<std::chrono::time_point<osv::clock::uptime, std::chrono::duration<long, std::ratio<1l, 1000000000l> > > >) at core/poll.cc:275
  27 (0x40007fe7e040) signal_handler  cpu0 running    -7.17013e-19               0 ?? at arch/x64/entry.S:160
Number of threads: 27
```

One can see that OSv executes the handler on the thread that does not have the same thread-local variables (see https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/nptl/pt-cleanup.c#L24C1-L38)/

wkozaczuk:
NOTE: This PR is a proposal and I am not sure we should merge it. Even though it fixes #1278 I have found that Golang apps signal handlers seem to validate that the stack used by signal handler routine was created by Golang. This means we may re-think how we handle user signal routine and possibly make it be executed on one of the application's threads (just like Linux does) instead of creating a new thread.

This patch changes logic around executing user signal handler in kill() to make new signal handler thread use app TLS of an application thread if available.

Normally, application threads share thread local storage area (TLS) with kernel or use one created by kernel. But when running statically linked executables or dynamic ones with Linux dynamic linker, the application threads create TLS on their own, and user signal handler may reference thread-local variables that do not exist in the TLS of the new signal handler thread. As a result, the app would crash like so:

```
9  0x000000004030bd54 in page_fault (ef=0x40007fe83088) at arch/x64/mmu.cc:42
10 <signal handler called>
11 __GI___pthread_cleanup_upto (target=target@entry=0x200000205080 <sigend_jmp_buf>, targetframe=targetframe@entry=0x40007fe93fc8 "\005\"D\177") at ./nptl/pthread_cleanup_upto.c:32
12 0x000020007f44232c in _longjmp_unwind (env=env@entry=0x200000205080 <sigend_jmp_buf>, val=val@entry=1) at ../sysdeps/nptl/jmp-unwind.c:27
13 0x000020007f442205 in __libc_siglongjmp (env=0x200000205080 <sigend_jmp_buf>, val=1) at ../setjmp/longjmp.c:30
14 0x0000200000202543 in sigend_handler (sig=2) at main.c:121
15 0x000000004037f0ee in sched::thread::main (this=0x40007fe7e040) at core/sched.cc:1415
16 sched::thread_main_c (t=0x40007fe7e040) at arch/x64/arch-switch.hh:377
17 0x000000004030bc52 in thread_main () at arch/x64/entry.S:161
```

This patch applies a bit of trickery (potentially dangerous if user handler tries to write to TLS) and selects an application TLS of a current thread or one of the other application threads and makes it available to the new signal handler thread. The signal handler thread in such case would switch from kernel TLS to app TLS before executing handler routine.

This patch makes following tests pass when running with Linux dynamic linker:
- tst-kill
- tst-sigwait
- tst-sigaction

Refers #1278

### DIFF
--- a/arch/x64/tls-switch.hh
+++ b/arch/x64/tls-switch.hh
@@ -48,6 +48,28 @@ public:
         }
     }
 };
+//
+//Simple RAII utility classes that implement the logic to switch
+//fsbase to the specified app address and back to the kernel one
+class user_tls_switch {
+    thread_control_block *_kernel_tcb;
+public:
+    user_tls_switch() {
+        asm volatile ( "movq %%gs:16, %0\n\t" : "=r"(_kernel_tcb));
+
+        //Switch to app tcb if app tcb present
+        if (_kernel_tcb->app_tcb) {
+            set_fsbase(reinterpret_cast<u64>(_kernel_tcb->app_tcb));
+        }
+    }
+
+    ~user_tls_switch() {
+        //Switch to kernel tcb if app tcb present
+        if (_kernel_tcb->app_tcb) {
+            set_fsbase(reinterpret_cast<u64>(_kernel_tcb->self));
+        }
+    }
+};
 
 }
 
--- a/core/sched.cc
+++ b/core/sched.cc
@@ -2088,6 +2088,16 @@ void with_thread_by_id(unsigned id, std::function<void(thread *)> f) {
     }
 }
 
+thread *find_first_app_thread(std::function<bool(thread &)> f) {
+    WITH_LOCK(thread_map_mutex) {
+        for (auto th : thread_map) {
+            if(th.second->is_app() && f(*th.second)) {
+               return th.second;
+            }
+        }
+    }
+    return nullptr;
+}
 
 }
 
--- a/include/osv/sched.hh
+++ b/include/osv/sched.hh
@@ -1552,6 +1552,8 @@ void with_all_threads(std::function<void(sched::thread &)>);
 // should return quickly.
 void with_thread_by_id(unsigned id, std::function<void(sched::thread *)>);
 
+thread *find_first_app_thread(std::function<bool(thread &)> f);
+
 }
 
 #endif /* SCHED_HH_ */
--- a/libc/signal.cc
+++ b/libc/signal.cc
@@ -21,6 +21,10 @@
 #include <osv/pid.h>
 #include <osv/export.h>
 
+#ifdef __x86_64__
+#include "tls-switch.hh"
+#endif
+
 using namespace osv::clock::literals;
 
 namespace osv {
@@ -393,6 +397,11 @@ int kill(pid_t pid, int sig)
                 signal_actions[sigidx].sa_flags = 0;
                 signal_actions[sigidx].sa_handler = SIG_DFL;
             }
+#ifdef __x86_64__
+            //In case this signal handler thread has specified app thread local storage
+            //let us switch to it before invoking the user handler routine
+            arch::user_tls_switch tls_switch;
+#endif
             if (sa.sa_flags & SA_SIGINFO) {
                 // FIXME: proper second (siginfo) and third (context) arguments (See example in call_signal_handler)
                 sa.sa_sigaction(sig, nullptr, nullptr);
@@ -401,6 +410,26 @@ int kill(pid_t pid, int sig)
             }
         }, sched::thread::attr().detached().stack(65536).name("signal_handler"),
                 false, true);
+        //If we are running statically linked executable or a dynamic one with Linux
+        //dynamic linker, its threads very likely use app thread local storage and signal
+        //routine may rely on it presence. For that reason we use app TLS of the current
+        //thread if it has one. Otherwise we find 1st app thread with non-null app TLS
+        //and assign the signal handler thread app TLS to it so it is switched to (see above).
+        //TODO: Ideally we should only run the logic below if the current app is statically
+        //linked executable or a dynamic one ran with Linux dynamic linker
+        //(what if we are handling "Ctrl-C"?)
+        u64 app_tcb = sched::thread::current()->get_app_tcb();
+        if (!app_tcb) {
+            auto first_app_thread = sched::find_first_app_thread([&](sched::thread &t) {
+                return t.get_app_tcb();
+            });
+            if (first_app_thread) {
+                app_tcb = first_app_thread->get_app_tcb();
+            }
+        }
+        if (app_tcb) {
+            t->set_app_tcb(app_tcb);
+        }
         t->start();
     }
     return 0;


gburd (commented):
The direction here is reasonable - switching fsbase to the target app thread's TCB so the signal handler runs against the app's thread-locals rather than the kernel's is the correct instinct, and the `user_tls_switch` RAII wrapper is a clean way to bound the switch. A few things that would help get this to the finish line:

1. **The TLS borrow via `find_first_app_thread` is the core fragility.** `kill()` picks the *first* app thread and runs the handler against that thread's TLS block. If two signals are delivered concurrently and both resolve to the same app thread, or if the chosen thread is itself mid-execution touching its own thread-locals, the handler and the thread stomp on each other's `errno`/TLS state. On Linux a signal runs on the target thread's own stack and TLS; borrowing one arbitrary thread's TCB is an approximation that holds only when handlers don't touch per-thread state. Worth deciding explicitly: is this an interim mitigation (document the constraint) or the path to a Linux-faithful redesign (deliver the signal on the actual target thread)?

2. **Gate it to the cases where it's needed and safe.** The switch only matters when the app has its own TLS - i.e. statically-linked apps or the Linux dynamic-linker path. For the kernel-internal case the switch is a no-op at best. Guarding on that (rather than switching unconditionally in `kill()`) narrows the blast radius and makes the intent legible.

3. **`find_first_app_thread` walks `thread_map` under `thread_map_mutex` on the signal path.** That's a lock acquired on every `kill()`; if a handler can itself raise a signal or if `kill()` is reachable from a context that already holds related locks, that ordering needs to be checked. At minimum call out the lock acquisition on the delivery path.

4. **Document the borrow at the call site.** A comment in `libc/signal.cc` stating that the handler executes against a borrowed app-thread TLS, that this is read-mostly-safe but write-unsafe for per-thread state, would save the next reader from assuming Linux semantics.

Concretely, to unblock: (a) confirm whether the intent is interim vs. full redesign; (b) add the static/Linux-linker gate; (c) note the `thread_map_mutex` acquisition and any lock-ordering constraint; (d) document the TLS-borrow contract at the call site. Happy to help test once the scope is settled.

Issue #853, opened 2017-02-12
Inefficient implementation of futex


nyh:
As comments in linux.cc explain, our current futex implementation was not optimized for performance, because the only real use we saw for it was in gcc's __cxa__guard_* functions, which are very rarely contended - and only when contended call the futex functions. So we just cared about the correctness of the futex implementation, not its performance.

However, it turns out that golang does use futexes heavily - @benoit-canet noted 10,000 per second (across a few cores) in one run. At that point, having such a bad implementation is not a good idea. Some of the bad things about the current implementation:

1. All the futexes, even unrelated ones, contend on one lock for the list of futexes.
2. If a futex is reused often, we constantly create wait queue objects and then delete them.
3. And probably other things :-)

In #23 there was one idea on how we can have a lock-free futex implementation. But it is likely that with some thought, even a locking implementation could be better.

nyh:
Some untested quick hacks we can use to determine which inefficiencies might be causing performance problems in an actual use case:

1. We can trivially replace the single queues_mutex and queues by an array of (say) 100 of those (mutexes and maps), and pick the uaddr%100'th mutex and map. This should take contention of this lock out of the equation.

2. The constant allocation and deallocation of the wait queue objects may be slow. We can comment out the queues.erase(i) line in the code to take it out of the equation.

nyh:
Another untested idea:
The 100 maps in the above idea could be protected by spinlocks (yikes!) instead of mutexes because we know they will only be held for very short durations (and our spinlock does not yet have short-wait spinning optimization).
Then, each "waitqueue" in the map could be, instead, a lock-free queue of waiters, <lockfree/queue-mpsc.hh>.

PR #1289, opened 2023-12-19
kernel: upgrade to C++ 17

wkozaczuk:
This patch makes necessary changes to the relevant source files to make it possible to build kernel at C++ 17 level. This will allow us to take advantage of new features available in C++ 17 standard.

The following changes are part of this PR:

- remove obsolete `register` keyword (mostly under `bsd/` part of the tree)

- add `const` specifier to the comparison operator definitions where necessary

- replace deprecated `std::iterator` usage with explicit definitions of traits (for example `iterator_category`, `value_type`, and 3 others) (for some background info read this - https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/)

- add `constexpr` to the `friend` declaration of std::distance<mbuf_iterator>

- make tests, httpserver-api, and some other modules compilable with non-C++ 11 level 
Please note, that all the unit tests and modules implemented in C++ are not subject to this change. On that note, the unit tests are passing.

An example of how to build with C++ 17:
```
./scripts/build image=httpserver-api fs=rofs.fg -j$(nproc) conf_cxx_level=gnu++17
```

Finally, this PR does not change the default C++ level which remains at 11, until we permanently decide to change in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,7 +292,7 @@ endif
 #     returns option if file builds with -ffoo, empty otherwise
 compiler-flag = $(shell $(CXX) $(CFLAGS_WERROR) $1 -o /dev/null -c $3  > /dev/null 2>&1 && echo $2)
 
-compiler-specific := $(call compiler-flag, -std=gnu++11, -DHAVE_ATTR_COLD_LABEL, compiler/attr/cold-label.cc)
+compiler-specific := $(call compiler-flag, -std=$(conf_cxx_level), -DHAVE_ATTR_COLD_LABEL, compiler/attr/cold-label.cc)
 
 source-dialects = -D_GNU_SOURCE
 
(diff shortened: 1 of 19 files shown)


nyh:
I'm curious, what is the motivation of this change. Which C++17 features are you planning to use?
But I admit it's kind of silly to stick to the 12 year old C++11 standard, so I won't oppose this change.

wkozaczuk:
@nyh I have some ideas on how to tackle #853 and I needed a way to re-purpose existing `waitlist` objects and instead of deleting them, keep them (your idea to comment out `erase`) and "re-key" unused entries by doing `extract` and `insert` (see https://www.fluentcpp.com/2020/05/01/how-to-change-a-key-in-a-map-or-set-in-cpp/). This is to minimize allocations.

wkozaczuk:
I have relied on the compiler to tell me what is wrong and then verify things still work by using unit tests. But based on your experience are there potentially things that may get broken by this change and we would not even know about it? (see 'Changes to evaluation order' or 'Guaranteed copy elision' in https://developers.redhat.com/articles/2021/08/06/porting-your-code-c17-gcc-11)

wkozaczuk:
@nyh I have updated this PR. The changes include fixes to `core/callstack.cc` and adding new build config option `conf_cxx_level`. So the kernel and all relevant modules get built with non-c++_11 only if explicitly specified.

PR #1306, opened 2024-03-17
vfs futimens: prepend path with a mount point part

wkozaczuk:
When mounting an optional filesystem using the option `--mount-fs`, it is important in the VFS layer to prepend the final path with a mount point path. 

This commit fixes it in `futimens()` implementation.


nyh (approved):
I guess that's correct, I see we already do this in a bunch of other functions.
Maybe eventually we can make this line into an inline function, so it doesn't need to be duplicated everywhere.

PR #1307, opened 2024-03-17
vfs: implement linkat() except for AT_SYMLINK_FOLLOW

wkozaczuk:


fs/vfs/main.cc
```diff
@@ -1245,6 +1245,39 @@ int link(const char *oldpath, const char *newpath)
     return -1;
 }
 
+OSV_LIBC_API
+int linkat(int olddirfd, const char *oldpath, int newdirfd, const char *newpath, int flags)
+{
+    if (flags & AT_SYMLINK_FOLLOW) {
```
nyh, line 1251: Maybe add `UNIMPLEMENTED("linkat() with AT_SYMLINK_NOFOLLOW");`
like we have with accessat(), so in the future if we have an application that needs it, it will be more visible?

fs/vfs/main.cc
```diff
@@ -1245,6 +1245,39 @@ int link(const char *oldpath, const char *newpath)
...
+    }
+
+    if (newpath[0] == '/' || newdirfd == AT_FDCWD) {
+        return vfs_fun_at2(olddirfd, oldpath, [newpath](const char *path) {
+            return link(path, newpath);
+        });
+    } else {
+        char absolute_newpath[PATH_MAX];
+        auto error = vfs_fun_at(newdirfd, newpath, [&absolute_newpath](const char *absolute_path) {
+            strcpy(absolute_newpath, absolute_path);
```
nyh, line 1268: Is this strcpy() safe? Do we know that absolute_path is shorter than PATH_MAX? If not, maybe should use strlcpy or something?

nyh (approved):

fs/vfs/main.cc
wkozaczuk: Added `WARN_ONCE` in the latest update.

fs/vfs/main.cc
wkozaczuk: Replaced with `strlcpy()` here and in `renameat()` where some code was stolen from.

wkozaczuk:
BTW the manual for `link()` says this:

"POSIX.1-2001 says that link() should dereference oldpath if it is a symbolic link. However, since kernel 2.0, Linux does not do so: if oldpath is a symbolic link, then newpath is created as a (hard) link to the same symbolic link file (i.e., newpath becomes a symbolic link to the same file that oldpath refers to). Some other implementations behave in the same manner as Linux. POSIX.1-2008 changes the specification of link(), making it implementation-dependent whether or not oldpath is dereferenced if it is a symbolic link. For precise control over the treatment of symbolic links when creating a link, see [linkat](https://linux.die.net/man/2/linkat)(2)."

I am not sure if our `link()` dereferences or not `oldpath` if symlink and given our `linkat()` is based on `link()` I am not certain the handling of this part of the logic in both `link()` and `linkat()` is the same as on Linux.

PR #1308, opened 2024-03-17
vfs: implement subset of copy_file_range()

wkozaczuk:
This patch implements a subset of `copy_file_range()` that is needed by the GNU `cp` utility to function.

The implementation delegates to `sendfile()` and accepts calls with `off_out` equal to 0 only.


fs/vfs/main.cc
```diff
@@ -2345,6 +2345,26 @@ ssize_t sendfile(int out_fd, int in_fd, off_t *_offset, size_t count)
...
+    //Non-zero flags are rejected according to the manual
+    if (flags != 0) {
+        errno = EINVAL;
+        return -1;
+    }
+    //We do not support writing to a file at specified offset because
+    //we delegate to sendfile() which assumes current position of the output
+    //file
+    if (off_out) {
+        errno = EINVAL;
```
nyh, line 2362: Maybe add UNIMPLEMENTED(...) message?

nyh (approved):

fs/vfs/main.cc
wkozaczuk: I am not sure we want to `abort()` in this case, as I think in this scenario an app may simply fall back to other libc functions. So returning `-1` may be better in this case. But maybe we can add WARN here.

nyh: Right, I mis-remembered that UNIMPLEMENTED only warned.
Perhaps the most appropriate command is WARN_ONCE() - so if some application uses this function in a loop, we'll only see the warning once?

PR #1310, opened 2024-03-18
ext4: add initial implementation

### LINKED ISSUE #1179: Implement minimal ext2/3/4 filesystem driver
This is a verbatim conversation from the mailing list:

On Wed, Oct 13, 2021 at 12:16 AM Gregory Burd <gr...@burd.me> wrote:
Hello OSv-ers,

I'm a huge fan of ZFS, it's an amazing bit of work and I'm thrilled it's a core component in OSv.  That said, it's not a great choice in all cases, the overhead of ZFS can outweigh the benefits.  I've heard many references to "adding another filesystem" into the mix in different contexts, most recently in the (amazing) talk given at p99conf by Waldek.

So, how about ext2 pulled straight from the BSD tree?
https://github.com/freebsd/freebsd-src/tree/main/sys/fs/ext2fs

Why ext2 and not <name other filesystem here>?  Well, it's not my favorite filesystem either but it is popular and well known.  It's easy for Linux users to get comfortable with and the tools are generally installed by default on most distros.  I would imagine that the BSD code is fairly complete and supported and I believe it supports ext2, 3, and 4 (https://wiki.freebsd.org/Ext2fs).

anyone have thoughts?

> On Wednesday, October 13, 2021 at 9:32:18 AM UTC-4 Nadav Har'El wrote:

> I think it makes sense, but only if it's something that you personally care about for some reason - e.g., that it's important for you for OSv to be smaller and you believe that replacing ZFS will make it smaller. Or that some other advantage of ext2 over zfs is interesting for you.

> Something worth keeping in mind is that one of the claimed advantages of OSv over, say, Linux, is that OSv does not need to support a gazillion different drivers and filesystems. It's not like anyone will ever plug a ext2-formatted disk or ntfs-formatted or whatever into OSv - so we don't *need* to support any of these filesystems. If we do want to support them, it should be out of some expected benefit - not out of necessity. So let's just spell out in advance what this benefit might be over the filesystems we already have (zfs, ramfs and rofs).

>> Waldek responded:
> > If we go ahead with implementing ext2 support, we should define a minimal subset of it we want to implement (do we need extents, large files, etc?). We should probably also not make the same mistake as with ZFS and NOT implement tools equivalent to zpool.so, mkfs.so, etc. Let us make it so for all admin functionality we would delegate to the toolset on host OS.  

> Also, I don't remember if Waldek did this or only partially (?), but if you're adding ext2 to reduce the kernel size, we first need to compile a kernel without zfs. We could add a build-time feature of removing zfs (see https://github.com/cloudius-systems/osv/issues/1110) or build it into a shared library that doesn't need to be loaded (https://github.com/cloudius-systems/osv/issues/1009). This would be similar to the Linux build system - which allows keeping some parts of the kernel out of the build, but also keeping some parts of the kernel in the build but as separate modules (sort of shared libraries).
 
anyone have suggestions on where to start?

> I would start with making (at least to myself) the case of what the benefit of adding ext2 would be.
If you think it is the code size, I would start by trying to estimate how much smaller the kernel would be without ZFS. For that I would start adding to our build system an option to compile without ZFS - or compile ZFS into a shared library.

>> I think in my presentation (slides 9/10) I claimed that I was able to trim the size of the kernel by ~0.7MB (after enabling GC). Also adding an option to build OSv without ZFS is something that I was planning to prepare proper patches for. However, this would come in the order described in the presentation - the 3rd step, but does not have to: we can work on the ability to compile ZFS out independently. I think I might have something ready for steps 1 and 2 (hide C++ std, enable GC) in the next 2-4 weeks. Meanwhile, my WIP branch - https://github.com/wkozaczuk/osv/commits/minimize_kernel_size - has all code changes I made for the presentation including adding commenting out ZFS in the right places so you can start experimenting with it. This particular commit deals with ZFS - https://github.com/wkozaczuk/osv/commit/df98287f39b28cb150dc60d1e4847820d71e1582.

> Then of course you can start implementing ext2. I agree you should try to find existing code in freebsd. You can see the other examples in the fs/ subdirectory (ramfs, rofs, devfs, nfs) on how to plug that code into OSv.

>> Yeah ideally, as Nadav points out, we would want to make ext2 driver a pluggable shared library - module. A good example of how to do it is how nfs was changed to become a shared library with this patch - https://github.com/cloudius-systems/osv/commit/4ffb0fa9329849cd587d62f91b6979bc0e0ce6d1. Hopefully, once you read the comments and the code changes it will all make sense. For example, these fragments - https://github.com/cloudius-systems/osv/commit/4ffb0fa9329849cd587d62f91b6979bc0e0ce6d1#diff-4dcb4336d0285de24fc7f3ebdb6805eb0c10ea645d9848dcfd38daa7742b363c and https://github.com/cloudius-systems/osv/commit/4ffb0fa9329849cd587d62f91b6979bc0e0ce6d1#diff-df0d94aa12dd9f4772529c9060f40a3492dc1b034f8bc2ae79bd2f978231f8ed, are key to see how OSv would automatically try to find an ext2 shared library under /usr/lib/fs and call its INIT functions to let it register its vfsops structure.

>> Now with nfs it is easier to achieve because we have a thin adapter layer under modules/nfs which delegates to https://github.com/sahlberg/libnfs which is built separately as libnfs.so.4.0.0. It would be ideal to do a similar thing with ext2 but can you build https://github.com/freebsd/freebsd-src/tree/main/sys/fs/ext2fs as a library or at least not have to copy it under OSv tree and then be stuck with that version of code forever? My concern is that if we fork the code or copy it, it will be difficult or a burden to maintain it in terms of bringing in bug fixes from upstream for example. Is it possible to achieve?.

> One of the problems you'll encounter will be the cache. I have to admit I don't remember everything we did there (Waldek might have a fresher memory, as he did rofs more recently), but because ZFS has such an elaborate caching mechanism, and OSv used ZFS, we avoided having yet another page-cache layer. That means that if ext2 doesn't come with its own page cache (because freebsd assumes a different layer handles the caching) your ext2 will not do any caching, which isn't great. Waldek's rofs dealt a bit with caching, so maybe you can copy it or be inspired from it, or also copy some caching code from freebsd.

>> ROFS comes with its own simple cache layer (https://github.com/cloudius-systems/osv/blob/master/fs/rofs/rofs_cache.cc and the commit to integrate it with pagecache - https://github.com/cloudius-systems/osv/commit/54b30715a6b6defd6b5c5008539824a0d3ca8a45) and here is the commit that finally integrated it into page cache - https://github.com/cloudius-systems/osv/commit/4c0bdbc1011bcc54b50f7d63d2fb5d55b5fb911f. Now, the changes I made are enough for read-only functionality. If we want full read-write ext2 support we would also need to figure out the "write" part.

>> Now having made all these comments about ext2, I think we should consider virtiofs as well. We have already the read-only implementation of virtiofs in OSv thanks to Fotis Xenakis (see this wiki page - https://github.com/cloudius-systems/osv/wiki/virtio-fs - it has many good references as well). Given that we can consider adding the write-support part to it and then delegate it to the host to provide whatever sophisticated filesystem it comes with. It is the beauty of virtiofs. But the downside is that virtiofs is only supported by some VMMs like QEMU and Intel's cloud hypervisor. So I still think it would be nice to have a simple and pluggable reasonably fast read-write filesystem support of ext2. 

-greg

wkozaczuk:
This commit adds an initial implementation of the ext4 filesystem driver based on the lwext4 project
(https://github.com/gkostka/lwext4). It provides a lightweight read-write alternative to the ZFS filesystem.

Please note this implementation is NOT thread-safe and will need to be enhanced to be so in the future. However, it is functional enough to support the test cases examined by `modules/libext/test.sh`.

One can build the OSv like so:

```bash
./scripts/manifest_from_host.sh -w /usr/bin/find && \
./scripts/build fs=rofs image=libext,native-example -j$(nproc) --append-manifest
```

Then create an ext4 filesystem:

```
mkdir -p ext_images
dd if=/dev/zero of=ext_images/ext4 bs=1M count=128 sudo mkfs.ext4 ext_images/ext4
```
Add some files to it if needed:

```
sudo losetup -o 0 -f --show ext_images/ext4
sudo mount /dev/loop0 ext_images/image

.. update content

sudo umount ext_images/image
sudo losetup -d /dev/loop0

qemu-img convert -f raw -O qcow2 ext_images/ext4 ext_images/ext4.img
```

And then run it:
```
./scripts/run.py --execute='--mount-fs=ext,/dev/vblk1,/data /hello' --second-disk-image ./ext_images/ext4.img
```
or using the test.sh
```
./modules/libext/test.sh '/find /data/ -ls'
```

Fixes #1179

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2075,6 +2075,7 @@ endif
 boost-libs := $(boost-lib-dir)/libboost_system$(boost-mt).a
 
 objects += fs/nfs/nfs_null_vfsops.o
+objects += fs/ext/ext_null_vfsops.o
 
 # The OSv kernel is linked into an ordinary, non-PIE, executable, so there is no point in compiling
 # with -fPIC or -fpie and objects that can be linked into a PIE. On the contrary, PIE-compatible objects
--- a/fs/ext/ext_null_vfsops.cc
+++ b/fs/ext/ext_null_vfsops.cc
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2024 Waldemar Kozaczuk
+ *
+ * Based on ramfs code Copyright (c) 2006-2007, Kohsuke Ohtani
+ *
+ * This work is open source software, licensed under the terms of the
+ * BSD license as described in the LICENSE file in the top-level directory.
+ */
+
+#include <osv/mount.h>
+
+#define ext_mount   ((vfsop_mount_t)vfs_nullop)
+#define ext_umount  ((vfsop_umount_t)vfs_nullop)
+#define ext_sync    ((vfsop_sync_t)vfs_nullop)
+#define ext_vget    ((vfsop_vget_t)vfs_nullop)
+#define ext_statfs  ((vfsop_statfs_t)vfs_nullop)
+
+static int ext_noop_mount(struct mount *mp, const char *dev, int flags,
+                          const void *data)
+{
+    printf("The ext module is in-active!. Please add ext module to the image.\n");
+    return -1;
+}
+
+/*
+ * File system operations
+ *
+ * This deactivates the EXT file system when libext.so is not loaded.
+ *
+ */
+struct vfsops ext_vfsops = {
+    ext_noop_mount, /* mount */
+    ext_umount,     /* umount */
+    ext_sync,       /* sync */
+    ext_vget,       /* vget */
+    ext_statfs,     /* statfs */
+    nullptr,        /* vnops */
+};
+
+extern "C" int ext_init(void)
+{
+    return 0;
+}
--- a/fs/vfs/vfs_conf.cc
+++ b/fs/vfs/vfs_conf.cc
@@ -56,6 +56,7 @@ extern struct vfsops zfs_vfsops;
 #if CONF_drivers_virtio_fs
 extern struct vfsops virtiofs_vfsops;
 #endif
+extern struct vfsops ext_vfsops;
 
 extern int ramfs_init(void);
 extern int rofs_init(void);
@@ -67,6 +68,7 @@ extern int nfs_init(void);
 extern int procfs_init(void);
 extern int sysfs_init(void);
 extern "C" int zfs_init(void);
+extern "C" int ext_init(void);
 
 /*
  * VFS switch table
@@ -82,5 +84,6 @@ const struct vfssw vfssw[] = {
 #if CONF_drivers_virtio_fs
 	{"virtiofs", 	virtiofs_init, 	&virtiofs_vfsops},
 #endif
+	{"ext",		ext_init,	&ext_vfsops},
 	{nullptr,	fs_noop,	nullptr},
 };
--- a/modules/libext/.gitignore
+++ b/modules/libext/.gitignore
@@ -1,0 +1,1 @@
+usr.manifest
--- a/modules/libext/Makefile
+++ b/modules/libext/Makefile
@@ -1,0 +1,33 @@
+arch=x64
+include ../common.gmk
+
+module_out := $(out)/modules/libext
+
+CXXFLAGS = -fPIC -std=gnu++11 $(INCLUDES) -I../lwext4/upstream/lwext4/include -I../lwext4/upstream/lwext4/build_lib_only/include \
+	 -D_KERNEL -D_GNU_SOURCE -Wall -fno-exceptions -fno-rtti
+
+# the build target executable:
+TARGET = libext
+CPP_FILES := ext_vfsops.cc ext_vnops.cc
+OBJ_FILES := $(addprefix $(module_out)/,$(CPP_FILES:.cc=.o))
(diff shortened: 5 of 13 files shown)


nyh (approved):
Looks like a nice start.
Please consider adding the instructions and examples which you aded in the commit message to a README file somewhere, or potential users will not remember how to use it.
Is it also possible to use ext4 as a root filesystem instead of zfs?

wkozaczuk:
Thanks for your review!

I will consider your recommendation to add a README do more cleanup and add some more comments.

No ext is not supported as a root filesystem yet. It should be quite easy, however even after some cleanup in the past the code in `loader.cc` and vfs `main.cc` si somewhat convoluted and we would have to add fifth filesystem to support as root besides ramfs, zfs, rofs and virtiofs.

Issue #1335, opened 2024-10-14
Do not use std::locale in kernel


wkozaczuk:
After some research and experimentation, it turns out that using `std::locale` from the C++ standard library can cost upwards of half a megabyte kernel size, and in the case of OSv, it is closer to 640K.

The `std::locale` is referenced indirectly most of the time, as this straightforward example illustrates, by streams:

```c++
#include <iostream>

int main() {
   std::cout << "Hello, world!\n";
}
```

```bash
c++ -static-libstdc++ test1.cc -o /tmp/test

du -h /tmp/test 
928K	/tmp/test
``` 

A similar example that does not use streams is way smaller:
```c++
#include <vector>

int main() {
   std::vector<int> a;
   a.push_back(1);
}
```

```bash
c++ -static-libstdc++ test2.cc -o /tmp/test_no_streams

du -h /tmp/test_no_streams 
172K	/tmp/test_no_streams
```

So any time we use streams directly or indirectly in the kernel (even for printing to standard out using `cout` or `cerr`) costs us over 500K.

Most of the savings come from removing symbols and related code as illustrated by this partial output from `readelf`:
```
10092: 000000004041b410     8 FUNC    LOCAL  DEFAULT    4 std::locale::locale(std::locale::_Impl*)
 10152: 000000004046f6c0   126 FUNC    LOCAL  DEFAULT    4 std::messages<wchar_t>::do_open(std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, st
d::locale const&) const
 10163: 000000004046d380   179 FUNC    LOCAL  DEFAULT    4 void std::__convert_to_v<long double>(char const*, long double&, std::_Ios_Iostate&, __locale_struct* const&)
 10243: 0000000040426330  2935 FUNC    LOCAL  DEFAULT    4 std::locale::facet::_M_cow_shim(std::locale::id const*) const
 10277: 000000004045ccb0    10 FUNC    LOCAL  DEFAULT    4 bool std::has_facet<std::num_put<wchar_t, std::ostreambuf_iterator<wchar_t, std::char_traits<wchar_t> > > >(std::
locale const&)
 10294: 0000000040428f90   558 FUNC    LOCAL  DEFAULT    4 std::istreambuf_iterator<wchar_t, std::char_traits<wchar_t> > std::__facet_shims::__money_get<wchar_t>(std::integ
[...]

PR #1337, opened 2024-10-18
do not use boost::format and streams in osv::printf() and osv::debug()

wkozaczuk:
- reimplement some usage of `osv::sprintf()` by using `std::to_string()` instead

- eliminate `osv::fprintf*() `functions and replace most of their usages with new `osv::sprintf()`

- eliminate usage of `std::ostringstream` and `std::stringstream` in some cases by simple concatenation of `std::string`

- replace template-based `std::string sprintf(const char* fmt, args... as)` with new `std::string sprintf(const char* fmt...)`

- add `std::string vsprintf(const char* fmt, va_list ap)`

- replace `void debug(const char* fmt, args... as)` and its usages with new `void debugf(const char* fmt...)` that delegates to `osv::vsprintf()`

- rename the C version of `void debugf(const char *fmt, ...)` to `void debugff(const char *fmt, ...)` to avoid name collision and adjust references in the code

- add any missing headers

This is another commit to remove code that directly or indirectly depends on std::locale.

Ref #1335

### DIFF
--- a/arch/aarch64/smp.cc
+++ b/arch/aarch64/smp.cc
@@ -9,7 +9,6 @@
 #include <osv/debug.h>
 #include <osv/sched.hh>
 #include <osv/prio.hh>
-#include <osv/printf.hh>
 #include <osv/aligned_new.hh>
 #include <osv/export.h>
 #include "processor.hh"
@@ -47,7 +46,7 @@ void smp_init()
     if (nr_cpus < 1) {
         abort("smp_init: could not get cpus from device tree.\n");
     }
-    debug("%d CPUs detected\n", nr_cpus);
+    debugf("%d CPUs detected\n", nr_cpus);
     u64 *mpids = (u64 *)alloca(sizeof(u64) * nr_cpus);
     if (!dtb_get_cpus_mpid(mpids, nr_cpus)) {
         abort("smp_init: failed to get cpus mpids from device tree.\n");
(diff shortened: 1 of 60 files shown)


arch/x64/mmu.cc
```diff
@@ -15,6 +15,7 @@
 #include <osv/prio.hh>
 #include <osv/elf.hh>
 #include "exceptions.hh"
+#include <algorithm>
```
nyh, line 18: I see in a bunch of files you added this `<algorithm>` but not used it. Is this a mistake?

wkozaczuk: It is used. In this file, it is `std::copy_if()`. 

In general, removing `<*stream>` from `osv/debug.hh` and `<boost/format.hpp>` from `osv/printf.hh` caused some source files like `mmu.cc` no longer compile. These were sometimes symbols from `<algorithm>` but sometimes different standard library headers, which probably were implicitly pulled by `<iostream>` or `<boost/format.hpp>` directly or indirectly.

arch/aarch64/smp.cc
```diff
@@ -47,7 +46,7 @@ void smp_init()
     if (nr_cpus < 1) {
         abort("smp_init: could not get cpus from device tree.\n");
     }
-    debug("%d CPUs detected\n", nr_cpus);
+    debugf("%d CPUs detected\n", nr_cpus);
```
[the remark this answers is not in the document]
wkozaczuk: I think because the C++ version of `debug()` became no longer a template and we had already C versions of `debug()` and `debugf()` and we had a conflict.

PR #1338, opened 2024-11-11
sched: Support POSIX's SCHED_RR scheduling policy

nmeum:
Previously, pull request #1223 added support for the `SCHED_FIFO` but didn't implement `SCHED_RR`. This PR attempts to follow-up on that by proposing an implementation of `SCHED_RR`.

Time slicing support, as mandated by `SCHED_RR`, is implemented though the `set_realtime_time_slice(duration)` API added in the aforementioned pull request. Within the scheduler, the amount of time the thread ran so far is tracked if it exceeds the set duration, the thread is preempted (if there is a runnable thread of same or higher priority). The thread's run time is reset on preemption.

This is my first OSv pull request, I added some basic tests which I can expand further. Additionally, I checked that the tests added #1223 still pass.

Let me know what you think.

### DIFF
--- a/core/sched.cc
+++ b/core/sched.cc
@@ -276,6 +276,10 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
     }
     thread* p = thread::current();
 
+    if (p->_realtime.has_slice()) {
+        p->_realtime._run_time += interval;
+    }
+
     const auto p_status = p->_detached_state->st.load();
     assert(p_status != thread::status::queued);
 
@@ -295,8 +299,17 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         } else if (!called_from_yield) {
             auto &t = *runqueue.begin();
             if (p->_realtime._priority > 0) {
-                // Only switch to a higher-priority realtime thread
-                if (t._realtime._priority <= p->_realtime._priority) {
+                // If the threads have the same realtime priority, then only reschedule
+                // if the currently executed thread has exceeded its time slice (if any).
+                if (t._realtime._priority == p->_realtime._priority &&
+                    ((!p->_realtime.has_slice() || p->_realtime.has_remaining()))) {
+#ifdef __aarch64__
+                    return switch_data;
+#else
+                    return;
+#endif
+                // Otherwise, only switch to a higher-priority realtime thread.
+                } else if (t._realtime._priority < p->_realtime._priority) {
 #ifdef __aarch64__
                     return switch_data;
 #else
@@ -336,15 +349,27 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         p->_detached_state->st.store(thread::status::queued);
 
         if (!called_from_yield) {
-            // POSIX requires that if a real-time thread doesn't yield but
-            // rather is preempted by a higher-priority thread, it be
-            // reinserted into the runqueue first, not last, among its equals.
-            enqueue_first_equal(*p);
+            if (p->_realtime.has_slice() && !p->_realtime.has_remaining()) {
+                // The real-time thread has exceeded it's time slice, enqueue
+                // it at the end among its equals and reset the time slice.
+                enqueue(*p);
+                p->_realtime.reset_slice();
+            } else {
+                // POSIX requires that if a real-time thread doesn't yield but
+                // rather is preempted by a higher-priority thread, it be
+                // reinserted into the runqueue first, not last, among its equals.
+                enqueue_first_equal(*p);
+            }
         }
 
         trace_sched_preempt();
         p->stat_preemptions.incr();
     } else {
+        // p is no longer running, if it has a realtime slice reset it.
+        if (p->_realtime.has_slice()) {
+            p->_realtime.reset_slice();
+        }
+
         // p is no longer running, so we'll switch to a different thread.
         // Return the runtime p borrowed for hysteresis.
         p->_runtime.hysteresis_run_stop();
@@ -380,7 +405,10 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         n->_runtime.add_context_switch_penalty();
     }
     preemption_timer.cancel();
-    if (n->_realtime._priority == 0) {
+    if (n->_realtime.has_slice()) {
+        assert(n->_realtime.has_remaining());
+        preemption_timer.set_with_irq_disabled(now + n->_realtime.remaining());
+    } else if (n->_realtime._priority == 0) {
         if (!called_from_yield) {
             if (!runqueue.empty()) {
                 auto& t = *runqueue.begin();
@@ -920,10 +948,6 @@ unsigned thread::realtime_priority() const
 
 void thread::set_realtime_time_slice(thread_realtime::duration time_slice)
 {
-    if (time_slice > 0) {
-        WARN_ONCE("set_realtime_time_slice() used but real-time time slices"
-                " not yet supported");
-    }
     _realtime._time_slice = time_slice;
 }
 
--- a/tests/tst-thread-realtime.cc
+++ b/tests/tst-thread-realtime.cc
@@ -1,0 +1,128 @@
+#include <atomic>
+#include <string>
+#include <iostream>
+#include <osv/sched.hh>
+
+#define TIME_SLICE 100000
+
+static std::string name = "tst-thr-wrk";
+static int tests = 0, fails = 0;
+
+static void report(bool ok, std::string msg)
+{
+    ++tests;
+    fails += !ok;
+    std::cout << (ok ? "PASS" : "FAIL") << ": " << msg << "\n";
+}
+
+static int fac(int n)
+{
+    if (n == 0) {
+        return 0;
+    } else if (n == 1) {
+        return 1;
+    } else {
+        return fac(n - 1) + fac(n - 2);
+    }
+}
+
+bool runtime_equalized()
+{
+    constexpr int num_threads = 5;
+    constexpr int switches = 3;
+    static std::atomic<bool> stop_threads;
+    sched::thread *threads[num_threads];
+
+    sched::cpu *c = sched::cpu::current();
+    for (int i = 0; i < num_threads; i++) {
+        threads[i] = sched::thread::make([&]{
+                while (!stop_threads.load()) {
+                    fac(10);
+                }
+        }, sched::thread::attr().name(name));
+
+        threads[i]->pin(c);
+        threads[i]->set_realtime_priority(1);
+        threads[i]->set_realtime_time_slice(sched::thread_realtime::duration(TIME_SLICE));
+        threads[i]->start();
+    }
+
+    auto runtime = sched::thread_realtime::duration(TIME_SLICE * num_threads * switches);
+    sched::thread::sleep(runtime);
+    stop_threads = true;
+
+    bool ok = true;
+    long prev_switches = -1;
+    for (int i = 0; i < num_threads; i++) {
+        long switches = threads[i]->stat_switches.get();
+        if (prev_switches != -1 && prev_switches != switches) {
+            ok = false;
+            break;
+        }
+        prev_switches = switches;
+    }
+
+    for (int i = 0; i < num_threads; i++) {
+        delete threads[i];
+    }
+
+    return ok;
+}
+
+bool priority_precedence()
+{
+    static std::atomic<bool> high_prio_stop;
+    sched::thread *high_prio = sched::thread::make([&]{
+        while (!high_prio_stop.load()) {
+            fac(10);
+        }
+    });
+
+    static std::atomic<bool> low_prio_stop;
+    sched::thread *low_prio = sched::thread::make([&]{
+        while (!low_prio_stop.load()) {
+            fac(10);
+        }
+    });
+
+    sched::cpu *c = sched::cpu::current();
+    high_prio->pin(c);
+    low_prio->pin(c);
+
+    high_prio->set_realtime_priority(2);
+    low_prio->set_realtime_priority(1);
+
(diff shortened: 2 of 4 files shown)


core/sched.cc
```diff
@@ -920,10 +948,6 @@ unsigned thread::realtime_priority() const
 
 void thread::set_realtime_time_slice(thread_realtime::duration time_slice)
 {
-    if (time_slice > 0) {
-        WARN_ONCE("set_realtime_time_slice() used but real-time time slices"
-                " not yet supported");
-    }
     _realtime._time_slice = time_slice;
```
nmeum, line 951: If it should be possible to change the time slice while the thread is already running, then we probably need to set the preemption timer here as well. Presently, I just assume that the slice is set before the thread is started.

core/sched.cc
```diff
@@ -336,15 +349,27 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
...
+            }
         }
 
         trace_sched_preempt();
         p->stat_preemptions.incr();
     } else {
+        // p is no longer running, if it has a realtime slice reset it.
+        if (p->_realtime.has_slice()) {
+            p->_realtime.reset_slice();
+        }
```
nmeum, line 371: It is not entirely clear to me if the slice should be reset if the thread is no longer runnable (e.g. because of blocking I/O). POSIX does not explicitly describe when the slice should be reset.

core/sched.cc
wkozaczuk: What if the new `_time_slice` is short enough that it should end its running slice now? Or maybe it means it should run until `now()` + new `_time_slice`?

wkozaczuk (commented):
Your code change looks good. I did ask some questions though.

wkozaczuk:
@nyh What do you think?

core/sched.cc
nyh: Yes, I'm also not sure, but think this if is right. I think the idea of the time slice is to make sure that a single thread never in its priority group never runs more than 1ms (for example) without letting other threads in its group run. But if the thread blocks or yields voluntarily (I believe this if covers both cases, right?), then it gives some other thread a chance to run and it too has a chance to run for a whole time-slice, so it's only fair that this thread's time slice is reset to zero. I think.
I tried searching if anybody mentions this question, and couldn't find such a discussion.

core/sched.cc
```diff
@@ -380,7 +405,10 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         n->_runtime.add_context_switch_penalty();
     }
     preemption_timer.cancel();
-    if (n->_realtime._priority == 0) {
+    if (n->_realtime.has_slice()) {
+        assert(n->_realtime.has_remaining());
+        preemption_timer.set_with_irq_disabled(now + n->_realtime.remaining());
```
nyh, line 410: Please make sure (I think I understood it, but not 100% sure) that the logic above gurantees that if !has_remaining we don't reach here at all.

In particular, what happens if some thread is the only thread in its priority group, so it continues to run? Does the code above recognize this case and remember to reset its time slice?

core/sched.cc
```diff
@@ -380,7 +405,10 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
...
+    } else if (n->_realtime._priority == 0) {
```
nyh, line 411: I would reverse the order if() to check realtime.priority==0 first and the has_slice() will be the else if. This is the most typical case, so it's more efficient to check it first.

Also, as I noted above twice already, I think if priority.realtime==0 we shouldn't even check the has_slice() which is meaningless - which also means you need to reverse the order of the if and else - not just for performance but also for correctness.

core/sched.cc
nyh: I think it is fine for set_realtime_time_slice() to not have an affect on a thread that is currently running.

Note that POSIX doesn't even have (!) a way to configure the realtime slice - if you use SCHED_RR it always uses some default time slice and you can't even change it, so this question doesn't even come up.

The worst case is when you switch from slice=0 to slice>0 - i.e., switch from SCHED_FIFO to SCHED_RR. I think it is fine for the change not to take affect until the thread yields or gets preempted for the first time. You can also add a comment stating that this is the case, and maybe we want to reconsider it in the future.

tests/tst-thread-realtime.cc
```diff
@@ -0,0 +1,128 @@
...
+    for (int i = 0; i < num_threads; i++) {
+        threads[i] = sched::thread::make([&]{
+                while (!stop_threads.load()) {
+                    fac(10);
+                }
+        }, sched::thread::attr().name(name));
+
+        threads[i]->pin(c);
+        threads[i]->set_realtime_priority(1);
+        threads[i]->set_realtime_time_slice(sched::thread_realtime::duration(TIME_SLICE));
```
nyh, line 46: nitpick: you don't need to pass a "thread_realtime::duration" option to set_realtime_time_slice - it's much nicer to pass a familiar unit like std::chrono::milliseconds(1) and C++ will automatically convert it to the right units needed by thread_realtime. That's the beauty of std::chrono!

tests/tst-thread-realtime.cc
```diff
@@ -0,0 +1,128 @@
...
+
+    auto runtime = sched::thread_realtime::duration(TIME_SLICE * num_threads * switches);
+    sched::thread::sleep(runtime);
+    stop_threads = true;
+
+    bool ok = true;
+    long prev_switches = -1;
+    for (int i = 0; i < num_threads; i++) {
+        long switches = threads[i]->stat_switches.get();
+        if (prev_switches != -1 && prev_switches != switches) {
```
nyh, line 58: Am I correct that you want all threads to have exactly the same number of context switches? How can we be confident of this - can't one happen to have one more than the others because of some inaccuracy or something?

tests/tst-thread-realtime.cc
```diff
@@ -0,0 +1,128 @@
...
+    high_prio->start();
+    low_prio->start();
+
+    auto runtime = sched::thread_realtime::duration(TIME_SLICE * 3);
+    sched::thread::sleep(runtime);
+
+    // Since both threads are pinned to the CPU and the higher priority
+    // thread is always runnable, the lower priority thread should starve.
+    bool ok = high_prio->thread_clock().count() > 0 &&
+        low_prio->thread_clock().count() == 0;
```
nyh, line 108: I'm a bit worried that it's theoretically possible that although you start()ed the high prio thread first and only then the low_prio one, maybe the low prio one got to run for a microsecond before the high prio one so its thread clock will not be exactly zero. But maybe in practice it doesn't happen...

nyh: This test can also have the opposite problem. If I understand correctly your TIME_SLICE is absolutely tiny, 0.1 milliseconds, and after starting high_prio and low_prio you only sleep 3 times that, i.e., 0.3 milliseconds, so it is theoretically possible that the test will pass even without any realtime priorities or anything, just because we let the first-ran highprio thread run for 0.3 milliseconds straight.

nyh (changes requested):
Looks very nice to me, and appears correct although I'm a bit worried that I'm rusty in this code and might have missed something. I only left a few minor comments/requests.

I just realized that we never actually implemented the POSIX API for these features ;-) I we think we have such patches in https://github.com/cloudius-systems/osv/issues/386 and maybe it will be nice to revive them.

core/sched.cc
nmeum: Ok, thanks for the input. I added a short comment to the the function prototype in the sched.hh header file in 25357218742054b97f62c37844c6381fb70790f0

core/sched.cc
```diff
@@ -295,8 +299,17 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         } else if (!called_from_yield) {
             auto &t = *runqueue.begin();
             if (p->_realtime._priority > 0) {
-                // Only switch to a higher-priority realtime thread
-                if (t._realtime._priority <= p->_realtime._priority) {
+                // If the threads have the same realtime priority, then only reschedule
+                // if the currently executed thread has exceeded its time slice (if any).
+                if (t._realtime._priority == p->_realtime._priority &&
+                    ((!p->_realtime.has_slice() || p->_realtime.has_remaining()))) {
```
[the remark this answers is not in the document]
nmeum: > […] should stay running if `_time_slice` is 0 (run 'forever' until yields or waits) OR there is still time left to run per its `_time_slice`, right?

Yes `p->_realtime.has_slice()` checks if it has a time slice (i.e. `_time_slice` != 0), `p->_realtime.has_remaining()` checks if there is still time remaining on the slice (if it has one).

Note that, even if the thread has exceeded its time slice it may still be selected to run again if there is no thread with a higher priority. Hence, the priority comparison in the if condition.

core/sched.cc
```diff
@@ -295,8 +299,17 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
...
+#ifdef __aarch64__
+                    return switch_data;
+#else
+                    return;
+#endif
+                // Otherwise, only switch to a higher-priority realtime thread.
```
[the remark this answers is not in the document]
nmeum: Thanks for the input, updated the comment in b2f1957269d46d41b66c06b2849e402e443ad667

core/sched.cc
nmeum: > Please make sure (I think I understood it, but not 100% sure) that the logic above gurantees that if !has_remaining we don't reach here at all.

Unfortunately, I of cause don't have a proof that the assertion always holds. However, from understanding of the scheduler code, I believe that it should. My thinking was to add it and if we discover a path where it fails during testing than we might have found a bug. I can also remove the assertion if you want me to.

> In particular, what happens if some thread is the only thread in its priority group, so it continues to run? Does the code above recognize this case and remember to reset its time slice?

Shouldn't this case be covered by the `priority_precedence` test that I added? If not we should expand the test suite to cover this.

tests/tst-thread-realtime.cc
nmeum: Good catch! Fixed in e9ea1a0a4cbdf27e2f2386598abd9ee484649527

tests/tst-thread-realtime.cc
nmeum: > Am I correct that you want all threads to have exactly the same number of context switches?

Yes. My thinking was: If we assign each thread a time slice of size `N` and then wait `N * NUM_THREADS * EXPECTED_SWITCHES` time units then (on a single core machine under a realtime scheduling policy), we would expect each thread to be preempted after `N` seconds. As such, each thread should have `EXPECTED_SWITCHES` amount of context switches.

Maybe I am missing something obvious, but I haven't seen this test fail yet. However, since this is not a hard-realtime operating systems I assume we could see delays here and there in the scheduler? We can also make the comparison fuzzy, allowing the amount of expected context switches to be off by 1 or 2?

tests/tst-thread-realtime.cc
nmeum: > If I understand correctly your TIME_SLICE is absolutely tiny, 0.1 milliseconds

I increased it further in 3c27113e3e6e2677db99384c42dad8c2c864f4fc

> maybe the low prio one got to run for a microsecond before the high prio one

Note: They are both pinned to same CPU so the higher prio one should also be first in the CPU runqueue and should always be runnable so I am not sure under which scenario the lower prio one would get the CPU. However, I believe you have more expertise with the scheduling code. We can also discard this test case.

What test cases would you like to see instead for `SCHED_RR`?

nmeum:
Thanks a lot to both of you for the detailed comments and feedback! I made some minor changes and left further comments above. I think the main two things that remain to be sorted out are:

1. Do we want want to support setting a realtime slice without a priority? If not, there are a few places where we should check if `priority == 0` as pointed out above.
2. I believe the tests should be expanded a bit, if you have ideas regarding useful test cases for `SCHED_RR` let me know. Maybe there are also tests for this scheduling policy in some POSIXy open source operating system of choice that we can use as a source of inspiration.

core/sched.cc
```diff
@@ -336,15 +349,27 @@ void cpu::reschedule_from_interrupt(bool called_from_yield,
         p->_detached_state->st.store(thread::status::queued);
 
         if (!called_from_yield) {
-            // POSIX requires that if a real-time thread doesn't yield but
-            // rather is preempted by a higher-priority thread, it be
-            // reinserted into the runqueue first, not last, among its equals.
-            enqueue_first_equal(*p);
+            if (p->_realtime.has_slice() && !p->_realtime.has_remaining()) {
```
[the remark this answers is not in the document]
nmeum: > check if realtime.priority>0 because maybe has_slice() just records some old setting and it's not in use now?

In 6f85c713b06f52e261ff842c84a1a4a9c8da0ed4 I moved the `realtime.priority>0` to `has_slice()` so `has_slice()` should *only` return true iff `realtime.priority>0`. I believe this should therefore be resolved.

tests/tst-thread-realtime.cc
nmeum: I removed this test, I don't think it's possible to check if the amount of preemptions/switches are equalized without the test being at least a little bit flaky.

nmeum:
Rebased, removed the flaky test, made the other test a little more useful and added the `priority == 0` check to `has_slice()` so threads without `priority == 0` are never subject to the time slice.

core/sched.cc
nmeum: Unless I am missing something, Linux only seems to reset the time slice once it expires not when the thread is no longer runnable. AFAIK This is the only place in the rt scheduler where it is reset: https://github.com/torvalds/linux/blob/v6.16/kernel/sched/rt.c#L2580-L2583

PR #1366, opened 2025-05-15
elf: Fix loading of shared objects with circular relocations

nmeum:
Without this patch, OSv fails to properly load shared objects with relocations that cause circular dependencies between the shared objects. To resolve such relocations, we need to first load all shared objects into memory and then relocate them [\[1\]][1]. Presently, OSv loads shared objects one-by-one and then relocates them directly in `load_object`.

This commit fixes this by moving the relocation of shared objects into the `get_library` function. The `load_object` function thereby no longer performs any relocations.

This was discovered while attempting to run a Haskell application, dynamically linked using GHC, on OSv. With shared objects emitted by GHC, there is such a circular dependency between the RTS (runtime system) and the Base library.

A minimal reproducer is available: https://gist.github.com/nmeum/5207585d8e4c0711ad10e9145dfc80a8 This example fails to run with current OSv (trying to execute or access missing symbol) and works correctly with this patch.

[1]: https://refspecs.linuxfoundation.org/elf/gabi4+/ch5.dynamic.html#dynamic_linker

### DIFF
--- a/core/elf.cc
+++ b/core/elf.cc
@@ -1517,13 +1517,9 @@ program::load_object(std::string name, std::vector<std::string> extra_path,
             _next_alloc = ef->end();
         add_debugger_obj(ef.get());
         loaded_objects.push_back(ef);
-        //Do not relocate static executables as they are linked with its own
-        //dynamic linker. Also do not try to load any dependant libraries
-        //as they do not apply to statically linked executables.
+        // Do not try to load any dependant libraries for static executable, they don't apply here.
         if (!ef->is_statically_linked_executable() && !ef->is_linux_dl()) {
             ef->load_needed(loaded_objects);
-            ef->relocate();
-            ef->fix_permissions();
         }
         _files[name] = ef;
         _files[ef->soname()] = ef;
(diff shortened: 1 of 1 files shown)


core/elf.cc
```diff
@@ -1553,6 +1551,14 @@ program::get_library(std::string name, std::vector<std::string> extra_path, bool
...
     _loaded_objects_stack.push(loaded_objects);
 
+    // Relocate objects *after* loading all objects, otherwise, we cannot resolve circluar relocations.
+    for (auto ef : loaded_objects) {
+        if (!ef->is_statically_linked_executable() && !ef->is_linux_dl()) {
+            ef->relocate();
+            ef->fix_permissions();
+        }
+    }
+
```
nyh, line 1561: Looks good, thanks! Please move the comment "Do not relocate static executables as they are linked with its own dynamic linker." to this new place as it explains the if(). The original comment probably needs to be edited, as half of it (the half about relocation) is no longer relevant in its original position.

nyh (changes requested):

nyh:
> A minimal reproducer is available: https://gist.github.com/nmeum/5207585d8e4c0711ad10e9145dfc80a8 This example fails to run with current OSv (trying to execute or access missing symbol) and works correctly with this patch.

If you're up to it, you can convert this minimal reproducer to a test. You can see modules/tests/Makefile for the test makefile, and it has other examples of bizarrely linked programs for other dynamic linker tests.

core/elf.cc
nmeum: Good catch, updated the comments accordingly.

nyh (approved):

PR #1367, opened 2025-05-19
tests: Add test case for circular ELF data relocations

nmeum:
As suggested by @nyh, this adds a regression test case for the PR #1366, based on the minimal reproducer in this PR. With the patch from this PR applied, the `tst-elf-circular-reloc` test exits with a zero exit status. Without it, it fails to load with the ELF loader error message: “trying to execute or access missing symbol”.

### DIFF
--- a/tests/lib-circular-reloc1.cc
+++ b/tests/lib-circular-reloc1.cc
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) 2025 Reliable System Software, Technische Universität Braunschweig.
+ *
+ * This work is open source software, licensed under the terms of the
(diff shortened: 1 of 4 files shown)


tests/lib-circular-reloc1.cc
wkozaczuk, line 1: It looks great to me.
But could you please add your copyright note to these 3 cc files, so it matches other tests. 
Thanks

wkozaczuk:
Could you also please drop '-main' from `tst-elf-circular-reloc-main.cc`, which I think is redundant?

nmeum:
Thanks for your feedback, updated the PR accordingly.

PR #1370, opened 2025-06-24
init: fix boot when using DT_GNU_HASH

### LINKED ISSUE #1369: OSv fails to boot when built on Fedora 42
After commit 9c9d4770e3f941128eb5c687fa41d26f90543910, it is possible to build the OSv kernel on Fedora 42 (with Gcc 15.1.1), i.e., run "make".

However, "scripts/build" doesn't work - OSv goes into a reboot loop when we try to run it to build the zfs filesystem. We'll need to debug why.

nyh:
Before this series, OSv fails to boot if built on Fedora 42 with Gcc 15.1.1.

It turns out the problem is symbol resolution during relocation of kernel symbols (specifically, TLS names like sched::s_current). Unfortunately we had two different copies of symbol resultion functions:

1. init_dyn_tabs::lookup() used for the initialization-time relocations.
2. object::lookup_symbol() used for any other lookups.

The latter is the complete implementation of symbol lookup, supporting both the traditional ELF "DT_HASH" hash table format, and the newer (2006) "DT_GNU_HASH" format. But the former implementation was incomplete, and was missing the DT_GNU_HASH support - so failed to do relocations (and boot...) on systems where DT_HASH was no longer output to the executable.

The first patch unifies both lookup functions into one function elf::lookup_symbol called by both paths. The new function's code is based on the more complete object::lookup_symbol().

After that patch, OSv can be built and run on Fedora 42 (and also continues to work on Fedora 41).

The second patch in the series also fixes a test that started to fail on Fedora 42 because of a new optimization that optimized away what the test was trying to do :-)

Fixes #1369.

### DIFF
--- a/core/elf.cc
+++ b/core/elf.cc
@@ -630,6 +630,16 @@ T* object::dynamic_ptr(unsigned tag)
     return static_cast<T*>(_base + dynamic_tag(tag).d_un.d_ptr);
 }
 
+template <typename T>
+T* object::opt_dynamic_ptr(unsigned tag)
+{
+    auto r = _dynamic_tag(tag);
+    if (r) {
+        return static_cast<T*>(_base + r->d_un.d_ptr);
+    }
+    return nullptr;
+}
+
 Elf64_Xword object::dynamic_val(unsigned tag)
 {
     return dynamic_tag(tag).d_un.d_val;
@@ -917,27 +927,6 @@ elf64_hash(const char *name)
     return h;
 }
 
-constexpr Elf64_Versym old_version_symbol_mask = Elf64_Versym(1) << 15;
-
-Elf64_Sym* object::lookup_symbol_old(const char* name)
-{
-    auto symtab = dynamic_ptr<Elf64_Sym>(DT_SYMTAB);
-    auto strtab = dynamic_ptr<char>(DT_STRTAB);
-    auto hashtab = dynamic_ptr<Elf64_Word>(DT_HASH);
-    auto nbucket = hashtab[0];
-    auto buckets = hashtab + 2;
-    auto chain = buckets + nbucket;
-    for (auto ent = buckets[elf64_hash(name) % nbucket];
-            ent != STN_UNDEF;
-            ent = chain[ent]) {
-        auto &sym = symtab[ent];
-        if (strcmp(name, &strtab[sym.st_name]) == 0) {
-            return &sym;
-        }
-    }
-    return nullptr;
-}
-
 uint_fast32_t
 dl_new_hash(const char *s)
 {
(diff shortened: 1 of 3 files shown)


nyh:
>     * Changing the symtab pointer from 'const Elf64_Sym*' to 'Elf64_Sym*' might imply that the symbol table is mutable. If the table should remain read-only, consider retaining const correctness and updating elf::lookup_symbol to accept a const pointer.

This is a good point but I didn't have a good alternative. The problem is that the more important API (object::lookup_symbol()) returns a Elf64_Sym* instead of a const Elf64_Sym* so needs a writable symbol table, so I had to change the second function (during initialization) to also use a writeable symbol table - even if we have no plans to write into the symbol table during initialization.

>     * [nitpick] The parameter 'sym' in init_dyn_tabs::lookup could be renamed to 'index' or a more descriptive name to avoid confusion with other 'sym' variables and enhance clarity.

This is perhaps true, but I kept the original name that was in this code. This code is indeed very inconsistent - sometimes "sym" is a `Elf64_Sym&`, sometimes `Elf64_Sym*`, sometimes u32, sometimes `symbol_module`.  I didn't plan to clear this mess up in this patch.

nyh:
Just to "prove" to myself that the new version I pushed still works on Fedora 42, I played a game of rogue on OSv :-)

```
                       __________
                      /          \
                     /    REST    \
                    /      IN      \
                   /     PEACE      \
                  /                  \
                  |       osv        |
                  |      916 Au      |
                  |   killed by a    |
                  |      quagga      |
                  |       2025       |
                 *|     *  *  *      | *
         ________)/\\_//(\/(/\)/\//\/|_)_______


[Press return to continue]

Top Ten Scores:
   Score Name
 1   916 osv: killed on level 8 by a quagga.
```

wkozaczuk (approved):
Your PR looks good, and I have tested it on both Fedora 37 and Fedora 42 by running the unit tests.

nyh:
The following patch fixes the test:
```patch
diff --git a/modules/tests/Makefile b/modules/tests/Makefile
index f48e2d01..53daec55 100644
--- a/modules/tests/Makefile
+++ b/modules/tests/Makefile
@@ -11,7 +11,7 @@ include ../common.gmk
 
 COMMON += -g -O2 -fPIC -DBOOST_TEST_DYN_LINK \
        -U _FORTIFY_SOURCE -D_KERNEL -D__OSV__ -DCONF_debug_memory=0 \
-       -Wall -Wno-pointer-arith -Wformat=0 -Wno-format-security
+       -Wall -Wno-pointer-arith -Wformat=0 -Wno-format-security -fno-builtin-malloc
 
 LIBS = $(libgcc_s_dir)/libgcc_s.so.1
```

Apparently gcc 15 has (?) a new optimization where if it sees a call to malloc() and the result not saved anywhere, it eliminates that call! This is exactly what this test does - it runs a bunch of malloc() calls never doing anything with the result. So these calls are eliminated, so they are not counted...

Eliminating this silly optimization in the test build makes the test pass.

But I need to come up with a better test patch patch:
1. I'm not sure the new `-fno-builtin-malloc` actually exists in older gcc
2. It should apply just to tst-small-malloc.cc, not to all files.

I think I prefer to do this without any changes to the build. The way I want to do this is to save all the allocated addresses in some global array, so the compiler doesn't think they are ignored.

PR #1371, opened 2025-07-01
refactor rootfs mounting code to support ext

wkozaczuk:
This patch enhances and refactors the rootfs mounting code to support the new ext filesystem. The libext module implements the ext filesystem as a library and can be loaded and mounted as explained in its readme using the `--mount-fs` boot option like so:

```bash
./scripts/run.py --execute='--mount-fs=ext,/dev/vblk1,/data /hello' --second-disk-image ./ext_images/ext4.img
```

However, up to this patch, it was not possible to boot OSv with the ext filesystem mounted at root. To accommodate this, we could have added a new `mount_ext_rootfs()` function to `fs/vfs/main.cc`, very similar to `mount_rofs_rootfs()`, `mount_zfs_rootfs()`, and `mount_virtiofs_rootfs()`. But instead, we replace all these 3 functions with a new generic `mount_rootfs()` that is flexible enough to mount any type of filesystem at the root.

Mounting ext filesystem is somewhat more involved but similar to zfs, because we first have to load the filesystem library - `libext.so` in case of ext, and `libsolaris.so` in case of zfs. To accommodate both, we replace the `load_zfs_library()` in `loader.cc` with a generic new function `load_fs_library()` and provide additional new function `load_ext_library_and_mount_ext_root()` that combines loading the ext filesystem library and mounting the ext at the root. We also refactor the `load_zfs_library_and_mount_zfs_root()` to mostly use newly introduced functions.

Another patch will implement necessary enhancements to the build script to support building OSv ext images.

Ref #1179

### DIFF
--- a/fs/vfs/main.cc
+++ b/fs/vfs/main.cc
@@ -2504,7 +2504,7 @@ void unpack_bootfs(void)
     }
 }
 
-void mount_rootfs(void)
+static void mount_init_rootfs(void)
 {
     int ret;
 
@@ -2543,7 +2543,7 @@ int nmount(struct iovec *iov, unsigned niov, int flags)
     return sys_mount(a.from, a.fspath, a.fstype, flags, nullptr);
 }
 
-static void import_extra_zfs_pools(void)
+extern "C" void import_extra_zfs_pools(void)
 {
     struct stat st;
     int ret;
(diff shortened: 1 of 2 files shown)


nyh (approved):
Looks good.
Later please explain in some document how to use those different filesystems. In particular, are there scripts/build parameters to use them in the root file system, and how to mount a second disk, etc. If I will want to use such a thing, I won't remember how :-)
By the way, what is the motivation to use ext? Is it to make it easier to create the filesystem in Linux instead of OSv? Is ext somehow smaller or better or whater than zfs?

wkozaczuk:
I have created a new pull request to support building ext images, and I have added some info that hopefully addresses some of your questions - https://github.com/cloudius-systems/osv/wiki/Filesystems#ext

wkozaczuk:
@nyh Also, how do you request a copilot review? I do not see it available in the list. Do I need to get some permission?
I am not sure how valuable they are, but is better than nothing.

PR #1393, opened 2026-04-01
Running Rust/tokio applications on OSv

bestouff:
Two fixes:

1. socketpair: mask `SOCK_NONBLOCK|SOCK_CLOEXEC` flags before the type assertion, apply `FNONBLOCK` to both fds
2. TLS size: bump default from `64` → `512` bytes to prevent silent memory corruption of thread-locals for Rust/Go apps

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -458,9 +458,9 @@ kernel_base := 0x200000
 lzkernel_base := 0x100000
 kernel_vm_base := 0x40200000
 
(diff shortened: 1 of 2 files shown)


bestouff:
(More details in the individual commits)

bestouff:
Thanks !
Do you know when this will be released ?

PR #1396, opened 2026-06-08
libc: upgrade musl from 1.1.24 to 1.2.1

gburd:
> **Depends on #1395** (build: modern-toolchain compat). This branch stacks on that PR; until it merges, this PR's diff includes the build-compat commit as well as the musl bump.

Bump the bundled musl from 1.1.24 (2019) to 1.2.1 (2020), the last musl
release with the BSD-style internal layout that OSv's libc shim expects
before musl's 1.2.x source re-architecture. The upgrade picks up roughly
four years of upstream bug fixes in stdio, math, dns, and threading.

This depends on PR 0 (modern-toolchain compat): musl 1.2.1 cannot build
on GCC 14 / Boost 1.87 without it. I verified that the lower_bound /
conditional_t breakage PR 0 fixes is toolchain breakage, not musl's --
bare master fails identically on the same toolchain.

### Changes

- Repin the `musl_1.1.24` submodule to the public musl v1.2.1 tag
  (73cc775b).
- Add `-isystem musl/arch/$(musl_arch)` to the musl per-object include
  path: 1.2.x relocated several arch-specific headers there.
- `include/api/sys/__socket.h`: provide `struct msghdr` / `struct cmsghdr`
  (with correct big/little-endian padding) at the common path -- 1.2.x
  moved them out of the arch-specific `bits/socket.h`.
- `libc/stdio/__string_read.c`: add OSv's copy of `__string_read` and move
  it from the `musl` group to the `libc` group (it is now referenced from
  OSv-built stdio objects, not the musl tree).
- nftw(): silence a 1.2.1 `-Wmaybe-uninitialized` and initialise the field
  explicitly so results no longer depend on allocator state (the 1.1.24
  code relied on 1.1.x malloc zeroing).
- Suppress `-Wincompatible-pointer-types` on `tempnam`/`tmpnam`/`__map_file`
  (musl 1.2.x signatures vs OSv's `syscall_to_function.h` shim) -- ABI
  unchanged; these are OSv-shim build warnings only.
- Minor aarch64/x64 `bits/` header deltas (dirent `d_off`, socket type)
  to match 1.2.1.

### Notes

- No ABI break for OSv apps: the libc layout OSv exposes is unchanged;
  this is a within-1.x musl bump, deliberately stopping at 1.2.1 before
  the 1.2.x internal re-architecture.
- The in-tree libc tests are used as the musl unit-test bar. tst-time's
  TZ cases require a zoneinfo database in the image; on an FHS host this
  comes from `/usr/share/zoneinfo` and the tests pass. (On a non-FHS host
  the manifest glob can miss tzdata and those cases read UTC -- a
  packaging artifact, not a libc fault.)
- Second in the build/libc modernization series (PR 0 -> this -> virtio
  multi-queue -> ...).


wkozaczuk:
Your PR looks good, and the unit tests pass.

But I have one request. Could you please rename the `musl_1.1.24` to `musl_1.2.1` otherwise it kind of looks misleading? Also, the `libc/README.md` still has some references left to 1.1.24 instead of 1.2.1. Also I wonder if we should update it so it reflects this change - 1.1.24 -> 1.2.1, and the last one 0.9.12 -> 1.1.24 - maybe a short note that we are upgrading again.

I have also run some analysis to get a better sense of how this upgrade affects the subset of musl OSv includes. And based on the diff between 1.2.1 and 1.1.24 (see below), there are ~25 files used by OSv that changed. 

```
grep ^musl\ + Makefile | sort | sed 's/musl += /git diff v1.1.24 src\//' | sed 's/\.o/\.c/'

## after running the above script, I see these files changed:
diff --git a/src/ctype/towctrans.c b/src/ctype/towctrans.c
diff --git a/src/ctype/wcwidth.c b/src/ctype/wcwidth.c
diff --git a/src/errno/strerror.c b/src/errno/strerror.c
diff --git a/src/math/coshf.c b/src/math/coshf.c
diff --git a/src/math/cosh.c b/src/math/cosh.c
diff --git a/src/math/__expo2f.c b/src/math/__expo2f.c
diff --git a/src/math/__expo2.c b/src/math/__expo2.c
diff --git a/src/math/__rem_pio2f.c b/src/math/__rem_pio2f.c
diff --git a/src/math/__rem_pio2l.c b/src/math/__rem_pio2l.c
diff --git a/src/math/__rem_pio2.c b/src/math/__rem_pio2.c
diff --git a/src/math/sinhf.c b/src/math/sinhf.c
diff --git a/src/math/sinh.c b/src/math/sinh.c
diff --git a/src/misc/nftw.c b/src/misc/nftw.c
diff --git a/src/network/lookup_name.c b/src/network/lookup_name.c
diff --git a/src/network/res_mkquery.c b/src/network/res_mkquery.c
diff --git a/src/stdio/tempnam.c b/src/stdio/tempnam.c
diff --git a/src/stdio/tmpnam.c b/src/stdio/tmpnam.c
diff --git a/src/stdio/ungetc.c b/src/stdio/ungetc.c
diff --git a/src/stdio/vfscanf.c b/src/stdio/vfscanf.c
diff --git a/src/string/memccpy.c b/src/string/memccpy.c
diff --git a/src/string/memmem.c b/src/string/memmem.c
diff --git a/src/string/strsignal.c b/src/string/strsignal.c
diff --git a/src/string/strstr.c b/src/string/strstr.c
diff --git a/src/time/__map_file.c b/src/time/__map_file.c
``` 

There are also some files under `libc` that link to the `musl` directory. Then there are all the musl headers symlinked from `include/api` (btw the `include/api/aarch64\x64/bits/dirent.h` symlinks need to be renamed to 1.2.1). 

Besides the notes in the PR at the top, do you know if there are any potential impacts these changes may have? I may be paranoid here.

BTW, I see you are still pointing to the https://github.com/osvunikernel/musl repo as it happens to have the 1.2.1 tag you need - nice (I did not realize that I cloned it when it must have been already 1.2.1, but somehow decided to use 1.1.24 - there must have been a good reason for that at the time).

libc/README.md
```diff
@@ -11,7 +11,7 @@ Original commit
...
+The current version of musl that OSv uses is 1.2.1. Most of the musl C files (`grep -P '^musl \+=' Makefile` - 577 files at this point) are directly referenced in the main makefile via a [`musl/`](/musl) symlink that currently points to the [`/musl_1.1.24`](/musl_1.1.24) git subproject. Some of the musl files are also symlinked from [`/libc`](/libc) subdirectory. Please also note that most header files under [`/include/api`](/include/api) symlink to the the musl copies under [`musl/include`](/musl/include) directory, but some are actually modified copies of original musl files. The internal musl headers under [`/include/api/internal_musl_headers`](/include/api/internal_musl_headers) symlink to files under [`musl/src/include`](/musl/src/include].
```
wkozaczuk: We still have references to 1.1.24 left.

gburd:
Pushed an amended `pr/musl` (now `8e5b093c`) that folds the rename into the single musl commit, so the submodule is `musl_1.2.1` from the start of the history rather than a follow-up:

- Renamed the submodule `musl_1.1.24` -> `musl_1.2.1` end-to-end: the `.gitmodules` stanza (name + path), the `.git/config` section, the `.git/modules/` dir, and the gitlink path. The top-level `musl` symlink now points at `musl_1.2.1/`. The Makefile only ever references `musl/`, so the dir rename is transparent to the build.
- Renamed both dirent symlinks: `include/api/aarch64/bits/dirent.h` and `include/api/x64/bits/dirent.h` now target `musl_1.2.1/arch/generic/bits/dirent.h`.
- `libc/README.md`: fixed the remaining `1.1.24` references (State section + symlink target), bumped the Upgrades note to `> 1.2.1`, and added the upgrade-history line you asked for:

  ```
  Upgraded from 0.9.12 to 1.1.24.

  Upgraded from 1.1.24 to 1.2.1.
  ```

The tree now has zero `musl_1.1.24` paths (`git ls-tree -r 8e5b093c | grep -c musl_1.1.24` -> 0).

One README reference to `1.1.24` is intentionally retained, under the time module:

> time (all but `__tz.c`) that comes from musl 1.1.24 and has been adapted

That line is accurate and not a stale leftover. OSv does not compile the musl submodule's `src/time/__tz.c`; it is absent from the `musl +=` list. Instead OSv builds an adapted copy at `libc/time/__tz.c` (Makefile, `libc/time/__tz.o`), which derives from the 1.1.24 `__tz.c` and was hand-modified to use OSv mutexes and the local-function syscall shim. I confirmed this on the build host: dropping the submodule's 1.2.1 `__tz.c` into the build fails to compile against OSv internals (lock-type mismatch, missing return, unused `index`), while the adapted `libc/time/__tz.c` builds clean. So the provenance note stays at 1.1.24 by design; bumping it to 1.2.1 would be wrong.

On impact: I agree with your read. I went through the subset of changed files OSv actually compiles (your `grep ^musl\ + Makefile` list intersected with the 1.1.24->1.2.1 diff). Every delta is an upstream correctness/bug fix, not a behavioral redesign:

- math (`cosh`/`sinh`/`__expo2*`/`__rem_pio2*`): accuracy and signed-shift UB fixes (casts to unsigned), no API/ABI change.
- `network/lookup_name.c`, `res_mkquery.c`: DNS rcode handling and query-id fixes.
- `stdio/vfscanf.c`, `ungetc.c`: scanf buffer priming / pushback correctness.
- `string/memmem.c`, `strstr.c`, `memccpy.c`, `strsignal.c`: substring-search and signal-string fixes.
- `ctype/towctrans.c`, `wcwidth.c`, `errno/strerror.c`: table/lookup fixes.
- `misc/nftw.c`, `time/__map_file.c`, `stdio/tempnam.c`/`tmpnam.c`: the handful that needed OSv-side touch-ups, already called out in the PR notes (nftw field init no longer relying on 1.1.x malloc zeroing; the tempnam/tmpnam/__map_file shim warning suppressions).

No struct layout or symbol-visibility changes reach the OSv ABI; this is deliberately the last within-1.x bump before musl's 1.2.x internal re-architecture. The in-tree libc tests (the unit-test bar) pass.

Re: the `osvunikernel/musl` remote -- yes, still pointing there since it carries the v1.2.1 tag. Nothing else changed about the remote.

libc/README.md
gburd: Fixed in 8e5b093c. The State-section and symlink-target 1.1.24 refs are now 1.2.1, and the tree has no musl_1.1.24 paths left. The one remaining 1.1.24 mention (the time/__tz.c provenance line) is intentional and accurate -- see the PR comment for why (OSv builds the adapted libc/time/__tz.c, not the submodule copy).

PR #1397, opened 2026-06-20
mmu/shm: hugepage hints, VMA virt_to_phys, POSIX shm, in_vma_range fix

gburd:
Five memory-management fixes that the later ZFS and Crucible work
depends on. Each is independent and self-contained.

This bases directly on current master (623ff90e); the build-compat and
musl 1.2.1 PRs that preceded it in the series are already merged.

### Changes

- **mmu: add MADV_HUGEPAGE and MAP_HUGETLB**. Implement the two Linux
  hugepage hint APIs against OSv's existing 2 MiB transparent-hugepage
  support. `madvise(MADV_HUGEPAGE)` records the hint on the VMA;
  `mmap(... MAP_HUGETLB ...)` allocates directly from the huge-page
  pool (ENOMEM rather than silent small-page fallback). OSv's
  huge_page_size is 2 MiB and the MAP_HUGE_2MB / MAP_HUGE_1GB
  size-selector bits are not decoded, so MAP_HUGETLB always maps with
  2 MiB pages regardless of any size hint.

- **mmu: fix virt_to_phys for VMA-mapped DMA buffers**.
  `virt_to_phys()` was correct only for kernel-direct-mapped
  addresses; on a VMA-mapped buffer it returned the linear-mapping
  offset rather than the page frame, producing silent DMA corruption.
  Walk the page tables and return the PTE's physical frame, keeping
  the direct-map fast path. Underlying cause of intermittent
  virtio-blk / Crucible DMA misalignment.

- **mempool: guard l2::refill against mid-batch allocation failure**.
  `l2::refill()` fills a page_batch with nr_pages calls to
  `free_page_ranges.alloc()`, then reuses the last page as the batch
  header. Under memory pressure alloc() can return nullptr; the old
  loop ignored that, turning a failed last slot into a null-header
  write and a failed earlier slot into a later fault. Check every
  iteration, stop at the first failure, return the partial pages, and
  account only for what was actually allocated.

- **shm: fix stale PTE bug, implement POSIX shm and ftruncate**.
  `shm_file::put_page()` returned without clearing the PTE on reuse,
  so a fresh SHM segment could read the previous tenant's data. Clear
  the PTE before returning false (false keeps the TLB-gather path from
  freeing the page; `shm_file::close()` owns it). Implements the rest
  of the POSIX shm surface the fix exposed: shm_open/unlink,
  shmctl(IPC_STAT), shm_file::truncate (sys_ftruncate now delegates to
  fp->truncate() for dentry-less files).

- **mmu: fix in_vma_range to use debug_base after mem_area refactor**.
  The mem_area refactor moved debug memory to a positive base; the old
  `addr >= 0` predicate then matched every normal allocation, firing
  vpopulate's assert on the first debug allocation and making
  conf_debug_memory unbootable. Discriminate on debug_base.

### Verified

Kernel compiles and links clean on GCC 14.3 / Boost 1.87 (fresh
loader.elf, RC=0). tst-shm-consistency (30/30) and tst-huge PASS on
the ZFS image (the test sources are carried by the openzfs PR later in
the series; the kernel-side APIs land here).

### DIFF
--- a/core/mempool.cc
+++ b/core/mempool.cc
@@ -1550,12 +1550,30 @@ void l2::refill()
                 // predictable.
                 reclaimer_thread.wait_for_memory(mmu::page_size);
             }
-            auto total_size = 0;
-            for (size_t i = 0 ; i < page_batch::nr_pages; i++) {
-                batch.pages[i] = free_page_ranges.alloc(page_size);
-                total_size += page_size;
+            size_t got = 0;
+            for (; got < page_batch::nr_pages; got++) {
+                page_range *pr = free_page_ranges.alloc(page_size);
+                if (!pr) {
+                    // Out of memory mid-batch. A page_batch is only usable
+                    // when completely filled: the last slot doubles as the
+                    // batch header (written below), and alloc_page_batch
+                    // hands out all nr_pages entries. A short batch with a
+                    // null in any slot would fault, so stop here and return
+                    // whatever we managed to grab. Checking inside the loop
+                    // (rather than only the last slot afterwards) is what
+                    // makes this sound: alloc can fail on any iteration.
+                    break;
+                }
+                batch.pages[got] = pr;
             }
-            on_alloc(total_size);
+            on_alloc(got * page_size);
+            if (got < page_batch::nr_pages) {
+                for (size_t i = 0; i < got; i++) {
+                    free_page_range_locked(
+                        new (batch.pages[i]) page_range(page_size));
+                }
+                break;
+            }
         }
         // Use the last page to store other page address
         pb = static_cast<page_batch*>(batch.pages[page_batch::nr_pages - 1]);
--- a/core/mmu.cc
+++ b/core/mmu.cc
@@ -178,9 +178,12 @@ phys virt_to_phys(void *virt)
     }
 #endif
 
-    // For now, only allow non-mmaped areas.  Later, we can either
-    // bounce such addresses, or lock them in memory and translate
-    assert(virt >= phys_mem);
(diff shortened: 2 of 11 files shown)


core/mempool.cc
```diff
@@ -1556,6 +1556,20 @@ void l2::refill()
                 total_size += page_size;
             }
             on_alloc(total_size);
+            if (batch.pages[page_batch::nr_pages - 1] == nullptr) {
```
wkozaczuk: I think this deserves a separate commit and explanation. Ideally, a unit test, if possible, would be great to verify it.
But I also have some questions:
1. Why do we look at the last page? What if an earlier page fails? Or do we assume that if any earlier page fails, then all subsequent ones fail as well?
2. Could we simply check for null in the `for(size_t i = 0 ; i < page_batch::nr_pages; i++)` loop above and stop right there instead of looping again and reversing the effect, or am I missing something?

core/mmu.cc
```diff
@@ -1152,6 +1152,21 @@ class initialized_anonymous_page_provider : public uninitialized_anonymous_page_
     }
 };
 
+// Page provider for MAP_HUGETLB strict mode: refuses 4KB small-page fallback.
```
wkozaczuk, line 1158: The description of this commit above mentions `MAP_HUGE_1GB` but I do not see how your changes make OSv support the 1GB pages.

wkozaczuk (commented):
Overall, your PR looks good. I have left one comment/question and another suggestion to extract an extra commit.

Issue #936, opened 2018-01-11
page_fault on lockfree::mutex::lock()


yuwang888:
i try to porting shop center  to osv, but it failed
/# page fault outside application, addr: 0x0000000000000000
[registers]
RIP: 0x00000000003d0c77 <lockfree::mutex::lock()+39>
RFL: 0x0000000000010206  CS:  0x0000000000000008  SS:  0x0000000000000010
RAX: 0x0000000000000001  RBX: 0xfffffffffffff8d8  RCX: 0xffff8000220b4090  RDX: 0xffff800000e8e040
RSI: 0x0000000000001fb0  RDI: 0x0000000000000000  RBP: 0x00002000205fe3b0  R8:  0x000020001bbcfb70
R9:  0xffff800000e914d0  R10: 0xffff80000b826170  R11: 0x0000000000000fff  R12: 0x0000000000000000
R13: 0xffff800000e914d0  R14: 0xffff80000b826170  R15: 0x0000000000000fff  RSP: 0x00002000205fe360
Aborted
gdb backtrace:
#0  0x000000000039c882 in cli_hlt () at arch/x64/processor.hh:247
#1  halt_no_interrupts () at arch/x64/arch.hh:48
#2  osv::halt () at arch/x64/power.cc:24
#3  0x0000000000226445 in abort (fmt=fmt@entry=0x635996 "Aborted\n") at runtime.cc:132
#4  0x0000000000226470 in abort () at runtime.cc:98
#5  0x000000000032f990 in mmu::vm_sigsegv (addr=<optimized out>, ef=0xffff8000220b9068) at core/mmu.cc:1316
#6  0x0000000000330b4b in mmu::vm_fault (addr=<optimized out>, addr@entry=0, 
    ef=ef@entry=0xffff8000220b9068) at core/mmu.cc:1338
#7  0x0000000000395f28 in page_fault (ef=0xffff8000220b9068) at arch/x64/mmu.cc:38
#8  <signal handler called>
#9  fetch_add (__m=std::memory_order_acquire, __i=1, this=0x0)
    at /usr/include/c++/4.8.2/bits/atomic_base.h:614
#10 lockfree::mutex::lock (this=0x0) at core/lfmutex.cc:29
#11 0x0000000000237646 in soabort (so=0xffffa00013074400) at bsd/sys/kern/uipc_socket.cc:705
[...]

PR #1399, opened 2026-06-24
kernel: TRIM/multiqueue, io_uring, TCP teardown, wakeup_one, ELF errors

gburd:
A batch of independent kernel- and block-layer changes found while running the ZFS and Crucible workloads. This branch bases directly on current `master` (3df7df76). Five commits, grouped below.

### Block layer

- **block: TRIM/DISCARD support and multiqueue I/O** (f9903dc9).
  - `BIO_DISCARD`: a new block-layer request type plumbed into virtio-blk's request descriptor (`VIRTIO_BLK_T_DISCARD`). Maps to ZFS's `ZIO_TYPE_TRIM` and to Crucible's protocol-level discard. Drivers without discard support return ENOTSUP; the caller falls back to overwrite-with-zero.
  - Multiqueue: per-CPU queue dispatch in virtio-blk so a multi-vCPU guest submits I/O on multiple queues concurrently. Adds a `scripts/run.py` helper wiring QEMU's `num-queues`, a per-queue completion-notifier locking-race fix (the lock was released too early under preemption), and a `device_delete_child` return-type fix surfaced while threading the multi-queue teardown path.

### Async I/O

- **fs: implement io_uring async I/O interface** (df0962ed).
  io_uring(7) implementation backed by OSv's existing async-I/O plumbing. Supported opcodes: READ, WRITE, FSYNC, NOP, OPENAT, CLOSE, READV, WRITEV, POLL_ADD, POLL_REMOVE, TIMEOUT, ACCEPT, CONNECT, RECV, SEND. Submission/completion queues mmap'd into user memory; SQE/CQE layout matches the Linux 5.15 ABI. `io_uring_setup`/`enter`/`register` syscalls; SQ poll thread (`IORING_SETUP_SQPOLL`); linked SQEs (`IOSQE_IO_LINK`) and drain barriers (`IOSQE_IO_DRAIN`). `tst-io_uring.cc` covers each opcode, link semantics, drain ordering, and SQ-poll. Note: OSv's io_uring is not used by OpenJDK 21 virtual threads (JDK 21 still routes through epoll); native code can use it.

### Kernel correctness

- **net: harden TCP net-channel fast path against connection teardown** (f96d88fe).
  `tcp_net_channel_packet()` assumed SOCK_LOCK held and the inpcb still attached. Under connection churn a queued packet can be drained after `in_pcbdetach()` cleared `inp_socket`, or from the IRQ classifier path without SOCK_LOCK, and a packet can arrive after the tcpcb moved to CLOSED/TIME_WAIT -- tripping `tcp_do_segment()`'s state KASSERT and panicking. Re-resolve the socket in the callback and drop the packet if detached; acquire SOCK_LOCK if not already held (OSv's recursive mutex makes re-locking a cheap depth bump); skip segments for connections at/below LISTEN or in TIME_WAIT, matching the slow path. This fix is load-bearing for the Crucible driver's sustained network I/O.

- **synch: fix wakeup_one to honor empty equal_range result** (f5a8ac54).
  `wakeup_one()` tested the equal_range lower bound against `_evlist.end()`. When the channel is absent, equal_range returns an empty range whose first iterator points at the next-larger key, not `end()`, so the old guard woke an unrelated waiter and erased its node while leaving the intended sleeper stranded. Test `first != second`.

- **elf: include path and sizes in "executable too short" error** (7a82addf).
  Report pathname, actual file size, and required length on a short ELF read so a truncated or mis-pathed binary is diagnosable.

### Verification

Kernel compiles and links clean on GCC 14.3 / Boost 1.87 (`./scripts/build image=empty`, fresh loader.elf, RC=0).

### DIFF
--- a/arch/x64/lzloader.ld
+++ b/arch/x64/lzloader.ld
@@ -10,7 +10,7 @@ SECTIONS
 {
     . = OSV_LZKERNEL_BASE + 0x1000;
     .text : { *(.text) }
-    . = OSV_LZKERNEL_BASE + 0x3000;
+    .rodata : { *(.rodata .rodata.*) }
     .data : { *(.data) }
     .bss : { *(.bss) }
     .edata = .;
--- a/bsd/porting/bus.h
+++ b/bsd/porting/bus.h
@@ -107,7 +107,7 @@ static inline int
 device_delete_child(device_t dev, device_t child)
 {
     printf("Implement me, line %s:%d\n",__FILE__,  __LINE__);
-    return NULL;
+    return 0;
 }
         
 // The two function below aim just at printing device information in
--- a/bsd/porting/synch.cc
+++ b/bsd/porting/synch.cc
@@ -120,8 +120,6 @@ int synch_port::_msleep(void *chan, struct mtx *mtx,
         if (chan) {
             // A pointer to the local "wait" may still be on the list -
             // need to remove it before we can return:
-            // A pointer to the local "wait" may still be on the list -
-            // need to remove it before we can return:
             mutex_lock(&_lock);
             auto ppp = _evlist.equal_range(chan);
             for (auto it=ppp.first; it!=ppp.second; ++it) {
@@ -160,8 +158,8 @@ void synch_port::wakeup_one(void* chan)
 
     mutex_lock(&_lock);
     auto ppp = _evlist.equal_range(chan);
-    auto it = ppp.first;
-    if (it != _evlist.end()) {
+    if (ppp.first != ppp.second) {
+        auto it = ppp.first;
         synch_thread* wait = (*it).second;
         _evlist.erase(it);
         trace_synch_wakeup_one_waking(chan, wait->_thread);
--- a/bsd/sys/netinet/tcp_input.cc
+++ b/bsd/sys/netinet/tcp_input.cc
@@ -3186,7 +3186,25 @@ tcp_newreno_partial_ack(struct tcpcb *tp, struct tcphdr *th)
 #include <bsd/sys/net/ethernet.h>
 #include <bsd/sys/net/netisr.h>
 
-// INP_LOCK held
+/*
+ * Net-channel packet callback.  Called from process_queue() in the
+ * context of whoever drained the per-connection queue: usually the
+ * thread that holds SOCK_LOCK after socket_file::poll() / sosend() /
+ * soreceive() / soclose() acquired it before the call.  But a packet
+ * can also be drained from the IRQ-side classifier path which does
+ * NOT hold SOCK_LOCK, and the inpcb may have been detached from the
+ * socket (in_pcbdetach sets inp_socket to NULL) between when the
+ * net_channel was registered and when this callback runs.
+ *
+ * Defensive shape:
+ *   - Re-resolve `so` inside the callback; bail if the inpcb has
+ *     been detached (no socket to deliver to).
+ *   - Acquire SOCK_LOCK if we don't already own it.  OSv's mutex is
+ *     recursive so re-locking when the caller already holds it is a
+ *     cheap depth bump rather than a deadlock.
+ *   - The old SOCK_LOCK_ASSERT guarded against a class of bugs that
+ *     no longer apply once we acquire the lock ourselves; remove it.
+ */
 static void
 tcp_net_channel_packet(tcpcb* tp, mbuf* m)
 {
@@ -3202,14 +3220,56 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
 	auto drop_hdrlen = h - start;
 	tcp_fields_to_host(th);
 	trace_tcp_input_ack(tp, th->th_ack.raw());
-	auto so = tp->t_inpcb->inp_socket;
+	auto inp = tp->t_inpcb;
+	if (!inp) {
+		m_freem(m);
+		return;
+	}
+	auto so = inp->inp_socket;
+	if (!so) {
+		/* inpcb detached from socket; drop and let the close path
+		 * tear down the channel via tcp_free_net_channel(). */
+		m_freem(m);
+		return;
+	}
 	auto ip_len = ntohs(ip_hdr->ip_len);
 	auto tlen = ip_len - (ip_size + (th->th_off << 2));
 	auto iptos = ip_hdr->ip_tos;
-	SOCK_LOCK_ASSERT(so);
+	bool taken = !SOCK_OWNED(so);
+	if (taken) {
+		SOCK_LOCK(so);
+		/* Re-validate after acquiring the lock: detach could have
+		 * raced with us between the !SOCK_OWNED check and the lock
+		 * acquisition. */
+		if (!tp->t_inpcb || tp->t_inpcb->inp_socket != so) {
+			SOCK_UNLOCK(so);
+			m_freem(m);
+			return;
+		}
+	}
+	/*
+	 * tcp_do_segment() asserts the connection is past TCPS_LISTEN and not
+	 * in TIME_WAIT — it only knows how to process segments for a live
+	 * data-bearing connection.  The slow path (tcp_input) drops segments
+	 * for CLOSED/LISTEN connections before ever reaching tcp_do_segment;
+	 * the net-channel fast path must do the same.  Under connection churn
+	 * (e.g. a client tearing down and reconnecting), a packet can still be
+	 * queued in the channel after the tcpcb has transitioned to CLOSED,
+	 * which previously tripped the KASSERT and panicked the kernel.
+	 */
+	if (tp->get_state() <= TCPS_LISTEN || tp->get_state() == TCPS_TIME_WAIT) {
+		if (taken) {
+			SOCK_UNLOCK(so);
+		}
+		m_freem(m);
+		return;
+	}
 	bool want_close;
 	m_trim(m, ETHER_HDR_LEN + ip_len);
 	tcp_do_segment(m, th, so, tp, drop_hdrlen, tlen, iptos, TI_UNLOCKED, want_close);
+	if (taken) {
+		SOCK_UNLOCK(so);
+	}
 	// since a socket is still attached, we should not be closing
 	assert(!want_close);
 }
--- a/core/elf.cc
+++ b/core/elf.cc
@@ -327,8 +327,14 @@ void file::read(Elf64_Off offset, void* data, size_t size)
 {
     // read(fileref, ...) is void, and crashes with assertion failure if the
     // file is not long enough. So we need to check first.
-    if (::size(_f) < offset + size) {
-        throw osv::invalid_elf_error("executable too short");
+    auto fsize = ::size(_f);
+    if (fsize < offset + size) {
+        char buf[256];
+        snprintf(buf, sizeof(buf),
+                 "executable too short %s: file_size=%lu need=%lu",
+                 _pathname.c_str(), (unsigned long)fsize,
+                 (unsigned long)(offset + size));
+        throw osv::invalid_elf_error(buf);
     }
     ::read(_f, data, offset, size);
 }
--- a/docs/block-discard.md
+++ b/docs/block-discard.md
@@ -1,0 +1,157 @@
+# Block Device DISCARD/TRIM Support
+
+## Overview
+
+OSv now supports DISCARD (also known as TRIM) operations for block devices. DISCARD allows the operating system to inform the storage device about blocks that are no longer in use, enabling space reclamation and improved performance on SSDs and thin-provisioned storage.
+
+## Architecture
+
+### BIO Commands
+
+The `BIO_DISCARD` command has been added to the bio subsystem (`include/osv/bio.h`):
+
+```c
+#define BIO_DISCARD  0x20  /* Space reclamation (TRIM) */
+```
+
+### VirtIO Block Device
+
+The virtio-blk driver supports DISCARD through the `VIRTIO_BLK_F_DISCARD` feature flag. When negotiated with the host, the driver can send `VIRTIO_BLK_T_DISCARD` requests to reclaim space.
+
+#### Configuration
+
+The driver reads the following DISCARD-related configuration from the device:
+- `max_discard_sectors`: Maximum number of sectors per DISCARD request
+- `max_discard_seg`: Maximum number of DISCARD segments
+- `discard_sector_alignment`: Required alignment for DISCARD operations
+
+### IOCTL Interface
+
+The `BLKDISCARD` ioctl allows userspace to issue DISCARD commands:
+
+```c
+#define BLKDISCARD _IO(0x12, 119)
+```
+
+Usage:
+```c
+uint64_t range[2];
+range[0] = start_offset;  // in bytes
+range[1] = length;        // in bytes
+ioctl(fd, BLKDISCARD, range);
+```
+
+## Filesystem Support
+
+### fstrim
+
+The `fstrim` utility can be used to discard unused blocks on mounted filesystems:
+
+```bash
+# Trim all unused blocks on /mnt
+fstrim -v /mnt
+
+# Trim with minimum extent size
+fstrim -v -m 1048576 /mnt  # Only trim extents >= 1MB
+```
+
+### Supported Filesystems
+
+- **ext4**: Full DISCARD support (mount with `-o discard` for automatic TRIM or use fstrim)
+- **XFS**: Full DISCARD support
+- **ZFS**: DISCARD support for space reclamation
+- **Btrfs**: Full DISCARD support
+
+## Performance Considerations
+
+### Automatic vs. Manual TRIM
+
+**Automatic TRIM** (mount option `-o discard`):
+- Pros: Immediate space reclamation
+- Cons: Performance overhead on every delete operation
+
+**Manual TRIM** (periodic fstrim):
+- Pros: Lower overhead, batched operations
+- Cons: Delayed space reclamation
+
+Recommendation: Use periodic fstrim (e.g., daily/weekly) rather than automatic TRIM for better performance.
+
+### DISCARD Alignment
+
+For optimal performance, DISCARD operations should be aligned to the device's discard alignment boundary. The virtio-blk driver automatically handles this based on device capabilities.
+
+## Testing
+
+### Basic DISCARD Test
+
+```bash
+# Create a filesystem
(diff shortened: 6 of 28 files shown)


wkozaczuk:
Hi,

Would you mind breaking this PR into 5 separate ones corresponding to the commits? It makes it easier to review and work on them independently.

Thanks,
Waldek

docs/block-discard.md
wkozaczuk, line 1: I know that the discard and multi-queue changes are somewhat intertwined, or at least touch the same areas of the code. But logically, I think, this commit should be split into 2, which could still be part of the same PR.

bsd/porting/bus.h
```diff
@@ -107,7 +107,7 @@ static inline int
 device_delete_child(device_t dev, device_t child)
 {
     printf("Implement me, line %s:%d\n",__FILE__,  __LINE__);
-    return NULL;
+    return 0;
```
wkozaczuk, line 110: Is this related to this part of the comment above, and if so, which part?
```
a per-queue locking-race fix (the per-queue completion-notifier
    lock was being released too early under preemption), removed
    next_queue_idx (became unused after dispatch was hashed by CPU),
    and a sys/dev device_delete_child return-type fix that surfaced
    while threading the multi-queue tear-down path.
```

drivers/blk-mq.cc
```diff
@@ -0,0 +1,78 @@
...
+ * BSD license as described in the LICENSE file in the top-level directory.
+ */
+
+#include <osv/blk_mq.h>
+#include <osv/sched.hh>
+#include <osv/debug.h>
+#include <vector>
+#include <errno.h>
+
+int blk_mq_init_tag_set(struct blk_mq_tag_set* set)
```
wkozaczuk, line 14: Are `blk_mq_init_tag_set()` and other functions in this module part of the C API intended to be used by the applications that want to take advantage of the multi-queue support? Does it mean that apps using standard libc functions like `read()/write()` will not benefit from your changes? Or will they, because of the changes to `virtio-blk.cc`, but to a lesser extent?
Or will some of these functions be used by the drivers you will add in future requests?

drivers/blk-mq.cc
```diff
@@ -0,0 +1,78 @@
...
+{
+    if (!set || !set->ops || set->nr_hw_queues == 0) {
+        return -EINVAL;
+    }
+
+    set->queue_map.resize(set->nr_hw_queues);
```
wkozaczuk, line 20: Should any of this be protected by a lock? I saw a `lock` mutex field in one of the structures.

drivers/blk-mq.cc
```diff
@@ -0,0 +1,78 @@
...
+        return nullptr;
+    }
+
+    unsigned int cpu_id = sched::cpu::current()->id;
+    unsigned int queue_idx = cpu_id % set->nr_hw_queues;
+
+    return set->queue_map[queue_idx];
+}
+
+int blk_mq_submit_bio(struct blk_mq_tag_set* set, struct bio* bio)
```
wkozaczuk, line 58: So maybe `blk_mq_submit_bio()` is an app API, and other functions in these modules are used by the driver?
Can we add some comments explaining the intention behind these functions, and maybe something in the beginning of this file?

drivers/virtio-blk.cc
```diff
@@ -130,49 +130,59 @@ blk::blk(virtio_device& virtio_dev)
...
+    // polls the remaining queues (simple single-interrupt multiqueue model).
+    for (int qid = 0; qid < _num_queues; qid++) {
+        std::string tname = std::string("virtio-blk") +
+                            (_num_queues > 1 ? "/q" + std::to_string(qid) : "");
+        sched::thread* t = sched::thread::make(
+            [this, qid] { this->req_done(qid); },
+            sched::thread::attr().name(tname));
+        t->start();
+        if (qid == 0) {
+            // Only queue-0's thread is woken by the interrupt.
```
wkozaczuk, line 147: This obviously simplifies the implementation, but it means all interrupts will be delivered from the 1st cpu to all others, leading to many ipis. I am just making a mental note. I wonder what we do in the nvme driver.

drivers/virtio-blk.cc
```diff
@@ -235,43 +245,101 @@ void blk::read_config()
...
+        if (qid == 0) {
+            /* Wait for the interrupt to signal any activity. */
+            virtio_driver::wait_for_queue(myqueue, &vring::used_ring_not_empty);
+            trace_virtio_blk_wake();
+
+            /* Drain all queues while we're awake.
+             * Hold each queue's lock so we don't race with the polling
+             * threads (qid > 0) that also call drain_queue on their queue. */
+            for (int q = 0; q < _num_queues; q++) {
+                WITH_LOCK(_queue_locks[q]) {
```
wkozaczuk, line 325: In the comments (`block-design.md`), you mention greater parallelism due to a lock-less design (each cpu operates on a separate queue), but here we create contention when the interrupt handler (`req_done()`) running on cpu 0 locks access to the queues belonging to other cpus and drains them on their behalf. Is there a better way to do it? Maybe each queue should have its own handler used by its cpu owner.
I think we do it in the nvme drivers - see `driver::register_io_interrupt()` in `nvme.cc`.

drivers/virtio-blk.cc
```diff
@@ -284,19 +352,21 @@ int64_t blk::size()
...
+            trace_virtio_blk_make_request_seg_max(bio->bio_bcount, _config.seg_max);
+            return EIO;
         }
+    }
+
+    /* Select a queue by CPU id so parallel CPUs use independent rings. */
+    int qid = sched::cpu::current()->id % _num_queues;
 
-        auto* queue = get_virt_queue(0);
+    WITH_LOCK(_queue_locks[qid]) {
```
wkozaczuk, line 367: So we lock on `_queue_locks[qid]` because we may race with `req_done()`, right?

arch/x64/lzloader.ld
```diff
@@ -10,7 +10,7 @@ SECTIONS
 {
     . = OSV_LZKERNEL_BASE + 0x1000;
     .text : { *(.text) }
-    . = OSV_LZKERNEL_BASE + 0x3000;
+    .rodata : { *(.rodata .rodata.*) }
```
wkozaczuk, line 13: This seems like an unrelated change. What does it fix?

include/api/x64/bits/syscall.h
```diff
@@ -644,6 +644,9 @@
 #define SYS_kcmp				312
 #define SYS_finit_module			313
 #define SYS_statx				332
+#define SYS_io_uring_setup			425
```
wkozaczuk, line 647: Is there a reason we do not expose those for aarch64?

fs/io_uring.cc
```diff
@@ -0,0 +1,1764 @@
...
+ *   - Linked requests (IOSQE_IO_LINK / IOSQE_IO_HARDLINK)
+ *   - IO drain (IOSQE_IO_DRAIN)
+ *   - Async cancel (IORING_OP_ASYNC_CANCEL)
+ *   - Timeouts (IORING_OP_TIMEOUT / IORING_OP_TIMEOUT_REMOVE / IORING_OP_LINK_TIMEOUT)
+ *   - Probe (IORING_REGISTER_PROBE)
+ *   - Files update (IORING_REGISTER_FILES_UPDATE)
+ *   - Eventfd notification (IORING_REGISTER_EVENTFD)
+ */
+
+#include <osv/io_uring.h>
```
wkozaczuk, line 28: Frankly, I am not sure I am equipped to review this code. But it looks very impressive. Did you base this implementation on what Linux does and use OSv primitives (threads, mutex, etc) where applicable? Did you have Claude implement most of it?
In either case, this seems like a very standalone change - 3 syscalls - this should not affect anything else but the app using it. So I think we can merge it as long as it is a separate PR.

modules/zfs-tools/usr.manifest
```diff
@@ -3,3 +3,9 @@
 /zfs.so: zfs.so
 /libzfs.so: libzfs.so
 /libuutil.so: libuutil.so
+/libtpool.so: libtpool.so
```
wkozaczuk, line 6: Again seems like an unrelated change.

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3186,7 +3186,25 @@ tcp_newreno_partial_ack(struct tcpcb *tp, struct tcphdr *th)
...
+ * Net-channel packet callback.  Called from process_queue() in the
+ * context of whoever drained the per-connection queue: usually the
+ * thread that holds SOCK_LOCK after socket_file::poll() / sosend() /
+ * soreceive() / soclose() acquired it before the call.  But a packet
+ * can also be drained from the IRQ-side classifier path which does
+ * NOT hold SOCK_LOCK, and the inpcb may have been detached from the
+ * socket (in_pcbdetach sets inp_socket to NULL) between when the
+ * net_channel was registered and when this callback runs.
+ *
+ * Defensive shape:
```
wkozaczuk, line 3199: Could https://github.com/cloudius-systems/osv/issues/936 be related?

As I understand, this affects the slow-path only, right?

I am re-reading this Wiki page - https://github.com/cloudius-systems/osv/wiki/Networking-Stack#net-channels-slow-path-vs-fast-path - I wrote some time ago, and I believe the part starting with this is the most relevant to understand your changes:

"The post_packet() pushes an mbuf onto the net channel only if one exists. But when does a net channel get created?"

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3202,14 +3220,56 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
...
+		}
+		m_freem(m);
+		return;
+	}
 	bool want_close;
 	m_trim(m, ETHER_HDR_LEN + ip_len);
 	tcp_do_segment(m, th, so, tp, drop_hdrlen, tlen, iptos, TI_UNLOCKED, want_close);
+	if (taken) {
+		SOCK_UNLOCK(so);
+	}
```
wkozaczuk, line 3272: I think it looks all correct.

bsd/porting/synch.cc
```diff
@@ -160,8 +158,8 @@ void synch_port::wakeup_one(void* chan)
 
     mutex_lock(&_lock);
     auto ppp = _evlist.equal_range(chan);
-    auto it = ppp.first;
-    if (it != _evlist.end()) {
+    if (ppp.first != ppp.second) {
```
wkozaczuk, line 161: So this is to make sure we do not try to process "empty" range represented by `ppp`?

core/elf.cc
```diff
@@ -327,8 +327,14 @@ void file::read(Elf64_Off offset, void* data, size_t size)
 {
     // read(fileref, ...) is void, and crashes with assertion failure if the
     // file is not long enough. So we need to check first.
-    if (::size(_f) < offset + size) {
-        throw osv::invalid_elf_error("executable too short");
+    auto fsize = ::size(_f);
```
wkozaczuk, line 330: Looks good!

wkozaczuk (commented):
Most of it looks great to me except the mq/discard commit where I left most comments. Also I am asking to to split it into 5 separate PRs.

docs/block-discard.md
gburd: Done. Discard and multiqueue are now in #1400, kept in a single PR since they touch the same regions of virtio-blk.cc, but the docs (block-discard.md, block-multiqueue.md) and the change are organized so each concern is independently reviewable.

bsd/porting/bus.h
gburd: Not related — it was an orphan change. `device_delete_child()` returns `int` but had `return NULL;`; virtio-blk never calls it (its destructor is a no-op TODO), and the only in-tree caller is xenbusb.cc. I removed the hunk entirely; it is not present in #1400.

drivers/blk-mq.cc
gburd: These were dead code. `blk-mq.cc`/`blk_mq.h` were not referenced by any driver, test, or app, and `read()`/`write()` already benefit from multiqueue purely via virtio-blk.cc (queue selected by submitting CPU). Rather than document/lock a speculative, unused API, I deleted the module; it is not present in #1400. If a future driver needs a generic MQ layer we can add it then, against a real consumer.

drivers/blk-mq.cc
gburd: Same as above — this was unused scaffolding and has been removed in #1400, so there is no API to lock. The real multiqueue locking lives in virtio-blk.cc, where each queue has its own mutex in `_queue_locks`.

drivers/blk-mq.cc
gburd: There was no app or driver consumer of any `blk_mq_*` function — it was speculative. The module is deleted in #1400; the multiqueue support that actually matters is the per-CPU queue selection in virtio-blk.cc.

drivers/virtio-blk.cc
gburd: You are right, and nvme.cc's `driver::register_io_interrupt()` is the better model. The current driver wires a single interrupt: cpu0's completion thread is woken and drains every queue under each queue's lock, while queues 1..N poll their own ring. Only completion is centralized — the submission path already scales (each CPU uses its own ring and lock). I documented this honestly in block-multiqueue.md, including the known limitation and the per-queue-interrupt path (as nvme does) as the future enhancement that lets each queue's completions land on its owning CPU and drops the cross-queue drain loop. See #1400. I did not claim a lock-less design.

drivers/virtio-blk.cc
gburd: The greater parallelism is on the submission path, which is genuinely lock-split: each CPU submits to its own ring under its own `_queue_locks[qid]`, so two CPUs on different queues never block each other. Completion is the part that is still centralized (single interrupt, cpu0 drains all). block-multiqueue.md now states this explicitly rather than implying end-to-end parallelism. See #1400.

drivers/virtio-blk.cc
gburd: Yes. `make_request()` holds `_queue_locks[qid]` because cpu0's `req_done()` drain-all loop and the qid>0 poll loop both call `drain_queue()` on that same ring. The per-queue lock serializes submission against completion on each ring.

arch/x64/lzloader.ld
gburd: Correct, unrelated. It gives `.rodata` its own output section in the lzloader link and is not part of io_uring. Moved to its own PR: #1405.

include/api/x64/bits/syscall.h
gburd: aarch64 already defines these — see include/api/aarch64/bits/syscall.h: `__NR_io_uring_setup/enter/register` (425/426/427) and the `SYS_*` aliases. The syscall registration (linux.cc, syscalls.cc.in) is arch-neutral, so io_uring works on aarch64 too. The x64 addition just brought x64 to parity; no aarch64 change was needed. See #1401.

fs/io_uring.cc
gburd: The ABI follows Linux 5.15 (SQ/CQ ring layout, io_uring_params, io_uring_setup/enter/register), implemented on OSv primitives (sched::thread, mutex, condvar; submissions run on the existing async I/O path). Yes, Claude did most of the implementation, reviewed and tested against the ring ABI. It is standalone (3 syscalls, nothing else depends on it) and is now its own PR: #1401.

modules/zfs-tools/usr.manifest
gburd: Correct, unrelated. Reverted — the change is not present in the io_uring PR (#1401).

bsd/sys/netinet/tcp_input.cc
gburd: Plausibly, yes — I'll reference #936 in the PR. Note this is not a slow-path change; it's the opposite. It hardens the net-channel **fast path** (`tcp_net_channel_packet`) so it mirrors what the slow path (`tcp_input`) already does: the slow path drops segments for CLOSED/LISTEN/TIME_WAIT before `tcp_do_segment()`, but the fast-path callback did not, so under connection churn a packet could be queued on the channel after the tcpcb went CLOSED and trip the KASSERT in `tcp_do_segment()`. Two fixes: (1) re-resolve `so` in the callback and bail if the inpcb was detached (`inp_socket == NULL`), since the callback can run from the IRQ-side classifier path without SOCK_LOCK; (2) reject `state <= TCPS_LISTEN || state == TCPS_TIME_WAIT`, matching the slow path. This guards exactly the lifetime window where the channel outlives the socket detach. Now in #1402.

bsd/sys/netinet/tcp_input.cc
gburd: Thanks — no change. Carried as-is into #1402.

bsd/porting/synch.cc
gburd: Exactly. `equal_range` returns `[first, second)`; an empty result has `first == second`, but `first` is not necessarily `end()` (it points at the first element greater than `chan`), so the old `if (it != _evlist.end())` could dereference a non-matching element. The fix tests `ppp.first != ppp.second`, matching how the range is correctly consumed elsewhere in the file. Now in #1403.

core/elf.cc
gburd: Thanks — carried as-is into #1404.

gburd:
Thanks for the thorough review. Per your structural request, I've split this into one PR per subsystem and addressed every inline comment (replies posted in-thread above). Closing this in favor of:

- **#1400** — block: TRIM/DISCARD + virtio-blk multiqueue (discard and MQ kept together since they touch the same code; speculative `blk-mq.cc` scaffolding deleted; orphan `bus.h` hunk removed)
- **#1401** — fs: io_uring (Linux 5.15 ABI on OSv primitives; `usr.manifest` change reverted; lzloader split out)
- **#1402** — net: harden TCP net-channel **fast path** against connection teardown (mirrors the slow-path CLOSED/LISTEN/TIME_WAIT drops; likely related to #936)
- **#1403** — synch: fix `wakeup_one` empty `equal_range` handling
- **#1404** — elf: richer "executable too short" error
- **#1405** — lzloader: give `.rodata` its own output section (the unrelated linker-script change)

Each split branch was build-qualified (kernel compile+link, `image=empty`) on a binutils 2.44 / g++ 14.3.0 host before pushing.

Issue #726, opened 2016-02-12
tcp_do_segment abort (NFS related?)


justinc1:
Crash:

```
sudo ./scripts/run.py -d -V -n -v -e '--ip=eth0,192.168.122.202,255.255.255.0 --defaultgw=192.168.122.1 --nameserver=192.168.122.1 /tools/mount-nfs.so nfs://192.168.122.1/ggg/?uid=0 /fff; tst-execve.so 30 /tst-file-read.so 30' --novnc -i build/debug/usr.img -c8 1>log-723-b/issue723-$ii 2>&1;

PP 7 tst-file-read.c ii=29/30
PP 14 tst-file-read.c ii=29/30
PP 9 fn=/fff/openfoam/etc/cellModels
PP 7 fn=/fff/openfoam/etc/cellModels
PP 14 fn=/fff/openfoam/etc/cellModels
PP 3 tst-file-read.c main EXIT ret=0
Assertion failed: tp->get_state() > 1 (bsd/sys/netinet/tcp_input.cc: tcp_do_segment: 1076)

[backtrace]
0x000000000022d8ea <abort(char const*, ...)+249>
0x000000000022d94b <__assert_fail+58>
PP 23 tst-file-read.c main EXIT ret=0
PP 16 tst-file-read.c ii=29/30
PP 16 fn=/fff/openfoam/etc/cellModels
0x00000000002a7206 <???+2781702>
0x00000000002abc3e <???+2800702>
0x00000000002abd19 <???+2800921>
0x00000000002ac6b7 <???+2803383>
0x00000000006313f8 <std::function<void (mbuf*)>::operator()(mbuf*) const+72>
0x0000000000630299 <net_channel::process_queue()+57>
0x0000000000244bb1 <???+2378673>
0x0000000000244ed2 <soclose+797>
0x0000000000254c77 <socket_file::close()+47>
0x0000000000675770 <fdrop+147>
0x000000000067529d <fdclose(int)+163>
0x000000000064cfae <close+35>
0x000000000070afbe <rpc_destroy_context+318>
0xffff8000091d903f <???+152932415>
```

Running on fedora 23 host, code is master@c93ebf9d140f + :

```
__cxa_atexit/finalize: be thread-safe
Add test for osv_execve.
Add osv_execve, osv_waittid
```

gdb:

```
(gdb) where
#0  processor::cli_hlt () at arch/x64/processor.hh:248
[...]

Issue #696, opened 2015-12-10
Assertion failed, crash


chrisbetz:
Hi,

I'm trying to run latest riemann on OSv (for sake of evaluating OSv and having a unikernel-based logmon solution). From time to time things go badly wrong like this:

Assertion failed. ttp->get_statell) o>c 1 (assd/sys/nlet",net/tcp_input.cc: tcp_do_segm ent:  1s076)

[backtracee]
rvice "r0x0000000000227898 <__assert_fail+24>
i0x0000000000278223 <???+2589219>
0x000000000027979c <???+2594716>
0x000000000041a3ae <net_channel::process_queue()+94>
0x00000000002377bf <int sbwait_tmo<osv::clock::uptime>(socket_, sockbuf_, boost::optional<std::chrono::time_point<osv::clock::uptime, osv::clock::uptime::duration> >)+335>
0x0000000000237625 <sbwait+37>
0x000000000023c144 <soreceive_generic+3732>
0x000000000023e96d <kern_recvit+269>
0x000000000023ecbb <sys_recvfrom+251>
0x000000000024e498 <linux_recv+56>
0x000000000023fca1 <recv+17>
0x000010000840f45c <???+138474588>
0x000010000840ed6f <Java_java_net_SocketInputStream_socketRead0+159>
0x000020000061675a <???+6383450>
0x0000200001340133 <???+20185395>
0x00002000006073cf <???+6321103>
0x000020000122dc93 <???+19061907>
0x000020000060770f <???+6321935>
0x000020000060770f <???+6321935>
0x00002000006077e3 <???+6322147>
0x000020000060770f <???+6321935>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x000020000060770f <???+6321935>
0x000020000060770f <???+6321935>
0x000020000060770f <???+6321935>
0x00002000006077e3 <???+6322147>
0x000020000060770f <???+6321935>
0x000020000060770f <???+6321935>
0x000020000060770f <???+6321935>
0x00002000008d5873 <???+9263219>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
0x00002000006077e3 <???+6322147>
[...]

Issue #454, opened 2014-08-13
closed socket, pending packets and rtx timer results in assertion failure in tcp_do_segment


tgrabiec:
I got this after running workloadf on my laptop.

```
Assertion failed: tp->get_state() > 1 (/home/tgrabiec/src/osv/bsd/sys/netinet/tcp_input.cc: tcp_do_segment: 1076)
```

```
#0  0x00000000003e0692 in cli_hlt () at /home/tgrabiec/src/osv/arch/x64/processor.hh:242
#1  halt_no_interrupts () at /home/tgrabiec/src/osv/arch/x64/arch.hh:48
#2  osv::halt () at /home/tgrabiec/src/osv/core/power.cc:34
#3  0x00000000002232a5 in abort (fmt=fmt@entry=0x5e7140 "Assertion failed: %s (%s: %s: %d)\n") at /home/tgrabiec/src/osv/runtime.cc:143
#4  0x00000000002232e9 in __assert_fail (expr=<optimized out>, file=<optimized out>, line=<optimized out>, func=<optimized out>) at /home/tgrabiec/src/osv/runtime.cc:149
#5  0x0000000000270470 in tcp_do_segment (m=m@entry=0xffffa0017f20e500, th=th@entry=0xffff800140fef02e, so=so@entry=0xffffa00109e84a00, tp=tp@entry=0xffffa00130062800, drop_hdrlen=0x42, tlen=tlen@entry=0x3d, iptos=iptos@entry=0x0, ti_locked=0x2, ti_locked@entry=0x1, want_close=@0xffff80010059deb0: 0x0) at /home/tgrabiec/src/osv/bsd/sys/netinet/tcp_input.cc:1075
#6  0x0000000000271f4f in tcp_net_channel_packet (m=0xffffa0017f20e500, tp=0xffffa00130062800) at /home/tgrabiec/src/osv/bsd/sys/netinet/tcp_input.cc:3210
#7  operator() (m=0xffffa0017f20e500, __closure=<optimized out>) at /home/tgrabiec/src/osv/bsd/sys/netinet/tcp_input.cc:3229
#8  std::_Function_handler<void(mbuf*), tcp_setup_net_channel(tcpcb*, ifnet*)::__lambda7>::_M_invoke(const std::_Any_data &, mbuf *) (__functor=..., __args#0=0xffffa0017f20e500) at /home/tgrabiec/src/osv/external/x64/gcc.bin/usr/include/c++/4.8.2/functional:2071
[...]

PR #1402, opened 2026-06-29
net: harden TCP net-channel fast path against connection teardown

gburd:
Split out of #1399 per review (one PR per subsystem).

Hardens the TCP net-channel **fast path** (`tcp_net_channel_packet`) so it
mirrors the protections the slow path (`tcp_input`) already has. The slow path
drops segments for CLOSED/LISTEN/TIME_WAIT connections before reaching
`tcp_do_segment()`; the fast-path callback did not, so under connection churn a
packet could be queued on the channel after the tcpcb transitioned to CLOSED
and trip the `KASSERT` in `tcp_do_segment()`, panicking the kernel.

Two fixes:
1. Lock the connection through the **inpcb** rather than the socket, and
   re-resolve `inp_socket` underneath that lock; bail if the inpcb was
   detached (`inp_socket == NULL`). For a TCP socket `SOCK_LOCK(so)` and
   `INP_LOCK(inp)` are the same underlying mutex, but the inpcb lock has stable
   lifetime while `inp_socket` is what `in_pcbdetach()` tears away. The callback
   can run from the IRQ-side classifier path without the lock held, so it takes
   `INP_LOCK` if it doesn't already own it (OSv's mutex is recursive) and
   re-validates after acquiring it.
2. Reject segments for `state <= TCPS_LISTEN || state == TCPS_TIME_WAIT`,
   matching the slow path.

This guards the lifetime window where the net channel outlives the socket
detach.

### How this was found and tested

The lifetime hazard was found by **code inspection** while auditing the
net-channel path, not from a captured panic, so there is no deterministic
in-tree reproducer to turn into a `tests/` case: the race is between
`net_channel` callback drain and socket detach under connection churn, and
reproducing it reliably would need fault-injection hooks the tree does not have.
The crash signature matches the NULL-socket-lock page fault reported in #936
(`soabort` -> `lockfree::mutex::lock(this=0x0)` during `soclose`), which is the
same `inp_socket`-outlives-socket window this guards. Verification is therefore
by inspection plus build-qualification: the guard applies the exact
CLOSED/LISTEN/TIME_WAIT checks the slow path already trusts.

g++ 14.3.0 host.

### DIFF
--- a/bsd/sys/netinet/tcp_input.cc
+++ b/bsd/sys/netinet/tcp_input.cc
@@ -3186,7 +3186,6 @@ tcp_newreno_partial_ack(struct tcpcb *tp, struct tcphdr *th)
 #include <bsd/sys/net/ethernet.h>
 #include <bsd/sys/net/netisr.h>
 
-// INP_LOCK held
 static void
 tcp_net_channel_packet(tcpcb* tp, mbuf* m)
 {
@@ -3202,14 +3201,56 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
 	auto drop_hdrlen = h - start;
 	tcp_fields_to_host(th);
 	trace_tcp_input_ack(tp, th->th_ack.raw());
-	auto so = tp->t_inpcb->inp_socket;
+
+	/*
+	 * Almost always we arrive here with the inpcb lock already held (the
+	 * old code simply assumed it).  The exception is the IRQ-side
+	 * classifier path, which can drain a queued packet without the lock.
+	 * So acquire INP_LOCK(inp) only if we don't already own it, and
+	 * release it below the same number of times (tracked by locked_here).
+	 *
+	 * We lock via the inpcb because tp->t_inpcb is the pointer we know is
+	 * valid; inp_socket is what in_pcbdetach() tears away.  Holding the
+	 * inpcb lock lets us re-read inp_socket and NULL-check it before use.
+	 */
+	auto inp = tp->t_inpcb;
+	if (!inp) {
+		m_freem(m);
+		return;
+	}
+	bool locked_here = !INP_LOCKED(inp);
+	if (locked_here) {
+		INP_LOCK(inp);
+	}
+	auto so = inp->inp_socket;
+	if (!so) {
+		if (locked_here) {
+			INP_UNLOCK(inp);
+		}
+		m_freem(m);
+		return;
+	}
 	auto ip_len = ntohs(ip_hdr->ip_len);
 	auto tlen = ip_len - (ip_size + (th->th_off << 2));
 	auto iptos = ip_hdr->ip_tos;
-	SOCK_LOCK_ASSERT(so);
+	/*
+	 * tcp_do_segment() only handles a live data-bearing connection.  The
+	 * slow path (tcp_input) drops segments for CLOSED/LISTEN/TIME_WAIT
+	 * before reaching it; mirror that here.
+	 */
+	if (tp->get_state() <= TCPS_LISTEN || tp->get_state() == TCPS_TIME_WAIT) {
+		if (locked_here) {
+			INP_UNLOCK(inp);
+		}
+		m_freem(m);
+		return;
+	}
 	bool want_close;
 	m_trim(m, ETHER_HDR_LEN + ip_len);
 	tcp_do_segment(m, th, so, tp, drop_hdrlen, tlen, iptos, TI_UNLOCKED, want_close);
+	if (locked_here) {
+		INP_UNLOCK(inp);
+	}
 	// since a socket is still attached, we should not be closing
 	assert(!want_close);
 }
--- a/bsd/sys/netinet/tcp_input.cc
+++ b/bsd/sys/netinet/tcp_input.cc
@@ -3203,30 +3203,24 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
 	trace_tcp_input_ack(tp, th->th_ack.raw());
 
 	/*
-	 * Almost always we arrive here with the inpcb lock already held (the
-	 * old code simply assumed it).  The exception is the IRQ-side
-	 * classifier path, which can drain a queued packet without the lock.
-	 * So acquire INP_LOCK(inp) only if we don't already own it, and
-	 * release it below the same number of times (tracked by locked_here).
-	 *
-	 * We lock via the inpcb because tp->t_inpcb is the pointer we know is
-	 * valid; inp_socket is what in_pcbdetach() tears away.  Holding the
-	 * inpcb lock lets us re-read inp_socket and NULL-check it before use.
+	 * Every caller reaches this through net_channel::process_queue(),
+	 * which always runs with the inpcb lock held (that lock aliases the
+	 * socket mutex - see in_pcb.cc: so->set_mutex(&inp->inp_lock)).  The
+	 * IRQ classifier only enqueues via post_packet(), it never invokes
+	 * the process function directly, so we can assert the lock rather
+	 * than re-acquire it.
 	 */
 	auto inp = tp->t_inpcb;
-	if (!inp) {
-		m_freem(m);
-		return;
-	}
-	bool locked_here = !INP_LOCKED(inp);
-	if (locked_here) {
-		INP_LOCK(inp);
-	}
+	INP_LOCK_ASSERT(inp);
+
+	/*
+	 * process_queue() drains a batch of mbufs in a loop; an earlier
+	 * segment can close the connection and detach the socket
+	 * (in_pcbdetach() clears inp_socket) before a later iteration runs.
+	 * Drop the packet instead of dereferencing a torn-down socket.
+	 */
 	auto so = inp->inp_socket;
 	if (!so) {
-		if (locked_here) {
-			INP_UNLOCK(inp);
-		}
 		m_freem(m);
 		return;
 	}
@@ -3239,18 +3233,12 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
 	 * before reaching it; mirror that here.
 	 */
 	if (tp->get_state() <= TCPS_LISTEN || tp->get_state() == TCPS_TIME_WAIT) {
-		if (locked_here) {
-			INP_UNLOCK(inp);
-		}
 		m_freem(m);
 		return;
 	}
 	bool want_close;
 	m_trim(m, ETHER_HDR_LEN + ip_len);
 	tcp_do_segment(m, th, so, tp, drop_hdrlen, tlen, iptos, TI_UNLOCKED, want_close);
-	if (locked_here) {
-		INP_UNLOCK(inp);
-	}
 	// since a socket is still attached, we should not be closing
 	assert(!want_close);
 }


bsd/sys/netinet/tcp_input.cc
```diff
@@ -3186,7 +3186,25 @@ tcp_newreno_partial_ack(struct tcpcb *tp, struct tcphdr *th)
 #include <bsd/sys/net/ethernet.h>
 #include <bsd/sys/net/netisr.h>
 
-// INP_LOCK held
+/*
```
nyh: The old comment said "INP_LOCK held". Is this still true here? Is it relevant?

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3186,7 +3186,25 @@ tcp_newreno_partial_ack(struct tcpcb *tp, struct tcphdr *th)
...
+ * net_channel was registered and when this callback runs.
+ *
+ * Defensive shape:
+ *   - Re-resolve `so` inside the callback; bail if the inpcb has
+ *     been detached (no socket to deliver to).
+ *   - Acquire SOCK_LOCK if we don't already own it.  OSv's mutex is
+ *     recursive so re-locking when the caller already holds it is a
+ *     cheap depth bump rather than a deadlock.
+ *   - The old SOCK_LOCK_ASSERT guarded against a class of bugs that
+ *     no longer apply once we acquire the lock ourselves; remove it.
```
nyh: The last bullet point belongs more in the commit message, less in a comment (a reader of this comment doesn't know that an "old SOCK_LOCK_ASSERT" was even here).

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3202,14 +3220,56 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
...
+		/* inpcb detached from socket; drop and let the close path
+		 * tear down the channel via tcp_free_net_channel(). */
+		m_freem(m);
+		return;
+	}
 	auto ip_len = ntohs(ip_hdr->ip_len);
 	auto tlen = ip_len - (ip_size + (th->th_off << 2));
 	auto iptos = ip_hdr->ip_tos;
-	SOCK_LOCK_ASSERT(so);
+	bool taken = !SOCK_OWNED(so);
```
nyh: The name "taken" is confusing - you set it to true if the lock is not already taken (by us). Later this makes sense ("taken" means we took the lock and need to release it), but right here, it's a confusing name.

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3186,7 +3186,25 @@ tcp_newreno_partial_ack(struct tcpcb *tp, struct tcphdr *th)
...
+ * NOT hold SOCK_LOCK, and the inpcb may have been detached from the
+ * socket (in_pcbdetach sets inp_socket to NULL) between when the
+ * net_channel was registered and when this callback runs.
+ *
+ * Defensive shape:
+ *   - Re-resolve `so` inside the callback; bail if the inpcb has
+ *     been detached (no socket to deliver to).
+ *   - Acquire SOCK_LOCK if we don't already own it.  OSv's mutex is
+ *     recursive so re-locking when the caller already holds it is a
+ *     cheap depth bump rather than a deadlock.
```
nyh: This paragraph is misleading, because you did NOT make use of the fact that OSv's mutex is recursive, and in fact you try hard to not re-take it if already owned, and remember not to release it if it was already owned. You could have made use of exactly that recusrive locking feature and avoid the "taken" variable, but you didn't. So please update either the comment or the implementation.

nyh (changes requested):
Looks good to me, I left only some minor requests.

However, I would like to know how you tested this fix. Do you have a way to reproduce this bug and checked that this fix really fixes it? It would have been nice to add a test in tests/ that reproduces it, but I realize it might be hard to reproduce and maybe you reproduced it in a long manual test, but surely you must have reproduced it somehow (and verified the fix somehow), so please mention in the commit message how.

gburd:
Thanks for the review. To your question about how this was tested:

I found this by **code inspection** while auditing the net-channel path, not from a captured panic, so I don't have a deterministic reproducer to turn into a `tests/` case. The race is between the `net_channel` callback drain and socket detach under connection churn — `in_pcbdetach()` nulls `inp_socket` and the TCP inpcb can outlive its socket, so a packet drained from the channel after detach reaches `tcp_do_segment()` against a torn-down connection and trips its state `KASSERT`. Reproducing that reliably from a test would need fault-injection hooks the tree doesn't have (a way to stall the callback across a detach).

The crash signature matches #936 — `soabort` → `lockfree::mutex::lock(this=0x0)` during `soclose` is the same `inp_socket`-outlives-socket window. So verification here is by inspection plus build-qualification rather than a new test: the guard applies the exact CLOSED/LISTEN/TIME_WAIT checks the slow path (`tcp_input`) already relies on.

I've rewritten the commit message and the PR description to spell all of this out, and (per the AI's and your comments) the fast path now locks via `INP_LOCK(inp)` with re-validation of `inp_socket` under the lock, rather than `SOCK_LOCK`. Updated head is `5fb8c3be`. Could you re-review when you have a moment?

bsd/sys/netinet/tcp_input.cc
gburd: Fixed in 5fb8c3be — removed the stale `// INP_LOCK held` line. The lock state is now tracked explicitly via the `locked_here` flag rather than an out-of-date comment.

bsd/sys/netinet/tcp_input.cc
gburd: Fixed in 5fb8c3be — moved this rationale into the commit message. The inline comment is now trimmed to just what a reader at this line needs.

bsd/sys/netinet/tcp_input.cc
gburd: Fixed in 5fb8c3be — renamed `taken` to `locked_here` so the variable reads as 'we acquired the lock in this frame' at each unlock site.

bsd/sys/netinet/tcp_input.cc
gburd: Fixed in 5fb8c3be — rewrote the recursive-mutex paragraph. It no longer claims re-locking is free; it now states plainly that OSv's mutex is recursive so re-entering when the caller already holds it is a depth bump (not a deadlock), and we unlock the same number of times via `locked_here`.

nyh:
> Thanks for the review. To your question about how this was tested:
> 
> I found this by **code inspection** while auditing the net-channel path, not from a captured panic, so I don't have a deterministic reproducer to turn into a `tests/` case. The race is between the `net_channel` callback drain and socket detach under connection churn — `in_pcbdetach()` nulls `inp_socket` and the TCP inpcb can outlive its socket, so a packet drained from the channel after detach reaches `tcp_do_segment()` against a torn-down connection and trips its state `KASSERT`. Reproducing that reliably from a test would need fault-injection hooks the tree doesn't have (a way to stall the callback across a detach).

I'm not worried about a "deterministic" reproducer, I'm more worried about whether you were able to ever reproduce this bug - or even more simply - to run a workload that you are sure reaches this code.

What scares me is that I'm **very rusty** in this code - I haven't looked at this networking code in nearly a decade and it wasn't even me who originally wrote it (it was @avikivity) - so it's hard for me to be confident in my review. I didn't even remember how or whether INP_LOCK and SOCK_LOCK are the same thing (I checked now, and it seems they do point to the same thing, I don't remember why). So unless you run some sort of workload that somehow reaches this code, not consistently or reliably but at least reaches it - how do we know the new code works? How do we know it doesn't outright crash?

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3202,14 +3201,57 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
...
 	tcp_fields_to_host(th);
 	trace_tcp_input_ack(tp, th->th_ack.raw());
-	auto so = tp->t_inpcb->inp_socket;
+
+	/*
+	 * Lock the connection through the inpcb, not the socket.  For a TCP
+	 * socket SOCK_LOCK(so) and INP_LOCK(inp) are the same mutex (see
+	 * in_pcballoc: so->set_mutex(&inp->inp_lock)), but the inpcb lock has
+	 * stable lifetime while inp_socket does not -- in_pcbdetach() nulls
+	 * inp_socket and the TCP inpcb can outlive the socket.  Taking the
```
nyh: I don't understand this lengthy explanation. We still use "so" below, so what is this discussion of a stable lifetime?
If both locks are the same, how can they have different lifetimes?

When I asked why you used SOCK_LOCK instead of INP_LOCK I though those were different, but maybe they are the same, and maybe there is no need to any explanation why you use one and not the other?

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3202,14 +3201,57 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
...
+	 * lock via inp lets us hold a valid lock and re-resolve the socket
+	 * underneath it.  The callback may run without the lock held (IRQ-side
+	 * classifier path), so acquire it if we don't already own it; OSv's
+	 * mutex is recursive, so re-locking when the caller already holds it
+	 * is a cheap depth bump.
```
nyh: But... you do **not** do re-locking in the code below! If the lock is already taken, INP_LOCKED is true so locked_here is false, and you do NOT call INP_LOCK.

So I don't understand why you describe recursive locking in this comment - when you deliberately tried to avoid using recursive locking.

If you wanted to use recursive locking, you could just call INP_LOCK() without checking if it was already locked or not.

bsd/sys/netinet/tcp_input.cc
```diff
@@ -3202,14 +3201,57 @@ tcp_net_channel_packet(tcpcb* tp, mbuf* m)
...
-	auto so = tp->t_inpcb->inp_socket;
+
+	/*
+	 * Lock the connection through the inpcb, not the socket.  For a TCP
+	 * socket SOCK_LOCK(so) and INP_LOCK(inp) are the same mutex (see
+	 * in_pcballoc: so->set_mutex(&inp->inp_lock)), but the inpcb lock has
+	 * stable lifetime while inp_socket does not -- in_pcbdetach() nulls
+	 * inp_socket and the TCP inpcb can outlive the socket.  Taking the
+	 * lock via inp lets us hold a valid lock and re-resolve the socket
+	 * underneath it.  The callback may run without the lock held (IRQ-side
```
nyh: I think this line "the callback may run without the lock held (IRQ-side classifier path), so acquire it if we don't already own it" is the most important line in this comment (and perhaps the only one I agreed with ;-)). I think you should explain that almost always we come here with INP_LOCK held (in fact, this is what the old comment assumed) but in this-and-that case we can come here without the lock, and we need to take it before we can continue calling more functions that assume it is taken.

bsd/sys/netinet/tcp_input.cc
gburd: Rewritten in 56c73def. You're right — the two locks are the same underlying mutex, so the 'different lifetimes' framing was wrong and is gone. The comment now says only why we take the lock *via the inpcb*: tp->t_inpcb is the pointer we know stays valid, whereas inp_socket is what in_pcbdetach() NULLs out. Taking INP_LOCK(inp) lets us then re-read inp_socket and NULL-check it before use. No more stable-lifetime discussion.

bsd/sys/netinet/tcp_input.cc
gburd: Fixed in 56c73def. Agreed — the code deliberately avoids recursive re-locking (if INP_LOCKED is already true, locked_here is false and we do NOT call INP_LOCK), so describing recursive locking was misleading. That whole paragraph is removed. The comment now just explains the acquire-if-not-owned / release-the-same-number-of-times pattern tracked by locked_here.

bsd/sys/netinet/tcp_input.cc
gburd: Done in 56c73def. This is now the lead of the comment: 'Almost always we arrive here with the inpcb lock already held (the old code simply assumed it). The exception is the IRQ-side classifier path, which can drain a queued packet without the lock. So acquire INP_LOCK(inp) only if we don't already own it...'

gburd:
Pushed 56c73def addressing the latest round of review on the `tcp_net_channel_packet()` comment.

Concretely, the inline comment was rewritten to fix the three things you flagged:

- **Removed the recursive-locking paragraph.** The code deliberately does *not* rely on recursive re-locking — if the inpcb lock is already held, `locked_here` is false and we skip `INP_LOCK` entirely. Describing recursive locking contradicted what the code does, so it's gone.
- **Dropped the stable-lifetime discussion.** `SOCK_LOCK(so)` and `INP_LOCK(inp)` are the same underlying mutex (`in_pcballoc` does `so->set_mutex(&inp->inp_lock)`), so 'different lifetimes' was simply wrong. The comment now explains only *why we lock via the inpcb*: `tp->t_inpcb` is the pointer we know stays valid, while `inp_socket` is what `in_pcbdetach()` tears away — so we take the inpcb lock, then re-read and NULL-check `inp_socket`.
- **Led with the point you agreed with.** The comment now opens with: almost always we arrive with the inpcb lock already held (the old code assumed it); the exception is the IRQ-side classifier path that can drain a queued packet without it, so we acquire the lock only if we don't already own it.

Also switched the unlock sites to `INP_UNLOCK(inp)` (matching the `INP_LOCK(inp)` acquire) per Copilot's note, so the unlock no longer depends on socket fields. Compiles clean.

gburd:
On whether the code is actually reached and whether it runs without crashing: yes, this is the default TCP receive hot path, not a rare configuration. tcp_setup_net_channel() is called unconditionally whenever a connection reaches ESTABLISHED, on both the active-open path (tcp_input.cc:1528, the SYN_SENT -> ESTABLISHED transition) and the passive-open path (tcp_input.cc:1936, SYN_RECEIVED -> ESTABLISHED). From then on the interface classifier delivers that connection's packets through the net channel, so tcp_net_channel_packet() is the function that processes essentially all in-order data segments for every established TCP connection. There is no sysctl or feature flag gating it; any TCP traffic exercises it.

That means the in-tree TCP tests all run through this code: tst-bsd-tcp1 and its zero-copy variants (tst-bsd-tcp1-zrcv/zsnd/zsndrcv), misc-tcp, misc-tcp-sendonly, misc-tcp-hash-srv, misc-tcp-close-without-reading, plus the httpserver-based tests. If any of these connect, transfer data, and close, they drive the new locking and the added state/socket re-validation on every received segment. The teardown checks (inp_socket re-read under INP_LOCK, and the CLOSED/LISTEN/TIME_WAIT drop) are what protect the close-without-reading and churn cases.

To be fully honest about the qualification level: I verified the fix by inspection plus build/link, and by reasoning that the added guards are the same checks tcp_input already trusts on the slow path (so a segment that the guard drops here is one the slow path would also have dropped rather than passed to tcp_do_segment). I have not stood up a high-connection-churn stress workload that deterministically hits the post-detach window. If you would like a concrete run-through before merge, the natural one is 'make check' (which runs the tst-bsd-tcp1* and misc-tcp* suite) on this branch; those all traverse this function for every segment, so a clean pass demonstrates the reworked path does not crash or regress ordinary traffic, even though it does not prove the rare detach race is closed. I can post the results of that run if that would give you enough confidence to merge.

gburd:
Ran the in-tree TCP suite against this branch (head `56c73def`) under QEMU to answer the "does it actually run without crashing" question directly, not by inspection.

All of these drive `tcp_net_channel_packet()` on every received segment (the function this PR modifies), and all pass clean with no panic, no `KASSERT`, and no `soabort -> lockfree::mutex::lock(this=0x0)` (the #936 signature this guard targets):

- `tst-bsd-tcp1` - 2/2, No errors detected
- `tst-bsd-tcp1-zrcv` - 2/2 (zero-copy receive)
- `tst-bsd-tcp1-zsnd` - 2/2 (zero-copy send)
- `tst-bsd-tcp1-zsndrcv` - 2/2 (zero-copy both)
- `tst-tcp-listen` - 2/2
- `tst-tcp-nbwrite` - 2/2 (blocking + non-blocking write, 600000 + 51200 bytes verified end to end)

For the teardown window specifically, I drove `misc-tcp-close-without-reading` directly: guest server accepts and closes without ever reading, host client floods connect/send/close cycles against it. The server tears the connection down while data is still arriving through the net channel - which is the exact post-detach path (`in_pcbdetach()` nulls `inp_socket`, packet drained from the channel afterward). No crash; the guest stays up and the close path behaves as designed (client sees the reset/EPIPE).

To be precise about qualification level: this exercises the modified path under real traffic including the close-without-reading teardown case, so we know the new `INP_LOCK`/`inp_socket` re-validation runs and does not fault. It is not a high-churn stress reproducer of the original race - I still don't have a deterministic trigger for the post-detach window - but the guard drops exactly the segments the slow path (`tcp_input`) already drops, so a segment it filters is one that would not have reached `tcp_do_segment()` anyway.

Build + runs were local QEMU (x86_64, gcc 15.2). Happy to add any specific workload you'd like to see run through it.

gburd:
Ping on this. Review feedback is addressed in-thread above; the branch is rebased on current master and mergeable.

For scheduling context: this is one of three prerequisites (#1398, #1400) under the OpenZFS 2.4.3 cutover replacing the old BSD-ZFS port. Landing these three unblocks a clean ZFS-only PR. Glad to iterate if anything remains.

wkozaczuk:
I have spent quite a bit of time reading the networking stack code and the Wiki page I wrote about it some time ago. 

Firstly, you were right when correcting me in my original review that this change affects the **fast path** (connection is established - ESTABLISHED, CLOSE_WAIT, FIN_WAIT_2, or TIME_WAIT). The `tcp_net_channel_packet()` that you are modifying here is assigned as a `_process_packet` function of the net channel object constructed by `tcp_setup_net_channel()`, and `_process_packet` is called in a loop by `net_channel::process_queue()`. The `net_channel::process_queue()` is then called by the nc **consumer** side in places like these:
```
bsd/sys/kern/uipc_sockbuf.cc:		so->so_nc->process_queue();
bsd/sys/kern/sys_socket.cc:        so->so_nc->process_queue();
bsd/sys/kern/sys_socket.cc:        so->so_nc->process_queue();
bsd/sys/kern/uipc_socket.cc:		so->so_nc->process_queue();
bsd/sys/netinet/tcp_input.cc:		nc->process_queue();
```

Therefore, I do not see how the IRQ classifier code, which calls `post_packet()` (see https://github.com/cloudius-systems/osv/blob/366e988dfdbb544dba3c622fd5a221d9bed6c700/drivers/virtio-net.cc#L542-L545), would be calling `tcp_net_channel_packet()` - instead it simply pushes mbuf on a channel queue and wakes consumers. And this is where you suggest the `tcp_net_channel_packet()` may be called without acquiring the socket/inpcb lock. So are we sure we could have a case when `tcp_net_channel_packet()` is called without the lock?

Otherwise, I think your changes to check the state of `t_inpcb` and `inp_socket` hanging of `tp` and dispose of mbuf if these are null seem to be correct to prevent a crash caused by the original line:
```
auto so = tp->t_inpcb->inp_socket;
```
However, we are not sure of the scenarios when `t_inpcb` and `inp_socket` could be null, and now I think #936 is about a different case - a null mutex when calling `SO_LOCK` in `soabort`. So it is a kind of `catch-all just in case` type of change to harden this part of the nc logic.

Also, I think your change to drop mbuf if the socket does not qualify as an established connection (the fast path scenario) is correct, but again, do we understand all the cases when this would happen? But again, this should not hurt. I think I have found a number of possibly related open and closed issues - see #454, #696, #726 (and maybe others - https://github.com/cloudius-systems/osv/issues?q=is%3Aissue%20state%3Aclosed%20tcp_do_segment).

Finallly, I have run some stress tests using the golang and rust "hello world" http server examples with apache bench, and I have not come across any crashes before this patch nor after. 

So other than the open (to me at least) question of why we need to possibly acquire the socket lock, this PR would be fine to merge.

gburd:
You're right, and I traced it to ground: the conditional lock-acquire was dead code, so I removed it in 53e26ca8.

The chain is exactly as you described. `tcp_net_channel_packet()` is only ever reached as the net channel's `_process_packet`, invoked by `net_channel::process_queue()` (core/net_channel.cc:27). Every caller of `process_queue()` already holds the inpcb lock:

- the five `flush_net_channel(so)` sites in `uipc_socket.cc` run under `SOCK_LOCK(so)`;
- the `tcp_flush_net_channel(tp)` sites in `tcp_input.cc:685` (under `INP_LOCK_ASSERT`) and `tcp_timer.cc` (delack/persist/rexmt, each under `INP_LOCK(inp)`);
- `sys_socket.cc` poll paths hold `SOCK_MTX_REF(so)`.

And `SOCK_LOCK` and `INP_LOCK` are the same mutex - `in_pcb.cc:476` does `so->set_mutex(&inp->inp_lock)`. As you noted, the IRQ classifier only enqueues via `post_packet()` and wakes consumers; it never calls the process function directly. So the lock is always already held here, and my original "IRQ-side path without the lock" justification was wrong.

53e26ca8 replaces the whole `locked_here`/`INP_LOCK` dance with `INP_LOCK_ASSERT(inp)`, matching the rest of tcp_input.cc. I kept the two guards you endorsed as harmless-and-correct: the `inp_socket` NULL-check (a batched segment can `tcp_close()`/detach the socket mid-drain, since `process_queue()` loops over a burst of mbufs and `in_pcbdetach()` clears `inp_socket`), and the connection-state check that mirrors what the slow path does before `tcp_do_segment()`. Both now run under the asserted lock.

Rebuilt the kernel with the change; the branch is at 53e26ca8.

PR #1403, opened 2026-06-29
synch: fix wakeup_one to honor empty equal_range result

gburd:
Split out of #1399 per review (one PR per subsystem).

`equal_range` returns `[first, second)`. An empty result has
`first == second`, but `first` is not necessarily `end()` — it points at the
first element greater than `chan`. The old `if (it != _evlist.end())` could
therefore dereference a non-matching element.

The fix tests `ppp.first != ppp.second`, matching how the range is correctly
consumed elsewhere in the file. A duplicated comment block in `_msleep` is also
removed.

g++ 14.3.0 host.

### DIFF
--- a/bsd/porting/synch.cc
+++ b/bsd/porting/synch.cc
@@ -120,8 +120,6 @@ int synch_port::_msleep(void *chan, struct mtx *mtx,
         if (chan) {
             // A pointer to the local "wait" may still be on the list -
             // need to remove it before we can return:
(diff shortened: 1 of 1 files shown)


nyh (approved):
Looks correct. If I understand BSD synchronization correctly, indeed if nobody is waiting on a channel, wakeup_one() should silently do nothing (and obviously not wake up something unrelated).

How did you discover this bug? Would be nice to write a test for it (I think it should be fairly easy?)

PR #1410, opened 2026-07-04
vfs: saturate block-device I/O by removing two serialization layers

gburd:
## Problem

OSv could not saturate the host's block-device bandwidth. Measured on real AWS Nitro (m6a.large / m6i.large, EBS gp3) against a raw device node: throughput was pinned at **QD1 == QD-high** - queue depth had no effect. randread held ~200 IOPS flat from QD1 to QD64; randwrite ~136 IOPS flat. A ramfs io_uring run on the same build hit 241k IOPS, so the io_uring engine was not the bottleneck. The bottleneck was the device read/write path.

Two independent serialization layers were pinning concurrency to one in-flight request.

### Layer 1 - buffer-cache 512B bio loop (`fs/vfs/vfs_bdev.cc`)

`bdev_read`/`bdev_write` looped one `BSIZE` (512-byte) block per iteration through the buffer cache (`bread`/`getblk` -> `rw_buf`), issuing exactly one synchronous bio and `bio_wait`ing on it before starting the next, all under the single global `bio_lock`. A 128k read became 256 serialized 512B round-trips at EBS ~0.5-1ms each.

The async fan-out primitive already existed and was unused by this path: `multiplex_strategy()` (the registered `devops->strategy`) splits a large bio by `max_io_size`, pre-inits the refcount, and issues all children in flight before any wait.

**Fix:** when the request is 512-aligned (offset and every `iov_len`), build ONE bio per iovec and dispatch through `dev->driver->devops->strategy` (== `multiplex_strategy`, which preserves `seg_max` splitting and adds `dev->offset` exactly once), then `bio_wait`. Unaligned / sub-block requests fall back to the existing `bread`/`uiomove` slow path for correctness (VFS does not force 512-alignment on device-node reads, e.g. `read(fd,buf,100)`).

### Layer 2 - exclusive per-vnode lock held across bio_wait (`fs/vfs/vfs_fops.cc`)

`vfs_file::read`/`write` took `vn_lock(vp)` - the exclusive per-vnode mutex - and held it across the entire synchronous `VOP_READ`/`VOP_WRITE`, including `bio_wait`. All threads sharing one block-device fd share one vnode, so concurrent `pread`/`pwrite` fully serialized (QD1 == QD64) even after Layer 1 was fixed.

**Fix:** for `VBLK` vnodes on the `FOF_OFFSET` path (`pread`/`pwrite`, which carry their own offset and do not touch `fp->f_offset`), skip `vn_lock` so N callers keep N requests in flight. The plain `read`/`write` path (`FOF_OFFSET == 0`, shared `fp->f_offset`) still takes the lock. `O_DSYNC`/`O_SYNC` -> `IO_SYNC` is preserved on the write fast path. Non-block filesystems (ramfs / ZFS / rofs) are unchanged.

### AWS Nitro EC2 (m6a.large, EBS gp3 4GiB volume = 3000 IOPS provisioned cap), raw device node

| workload | pre-fix | post-fix |
|---|---|---|
| randread QD1 | ~200 IOPS | 1658 IOPS |
| randread QD2 | ~200 IOPS | 3249 IOPS |
| randread QD4..QD64 | ~200 IOPS (flat) | ~3003 IOPS (device cap) |

Post-fix latency grows linearly with QD (Little's Law: QD64 x ~333us p50 ≈ 21ms), confirming requests run concurrently and the plateau is the gp3 provisioned IOPS ceiling, not an OSv serialization limit.

### Local KVM, uncapped host NVMe (isolates OSv scaling from the EBS cap)

| workload | QD1 | QD16 |
|---|---|---|
| randread | 2080 IOPS | 20597 IOPS (9.9x) |
| randwrite | - | 20738 IOPS |

Clean ~10x scaling on an uncapped device proves genuine queue concurrency.

### Same-host OSv-vs-Linux parity (i4i.large, uncapped local instance-store NVMe)

The EBS numbers above are masked by the gp3 provisioned-IOPS cap. To prove OSv now saturates the hardware rather than hitting a software serialization limit, I ran an A/B on **i4i.large** (local instance-store NVMe ~436 GB, uncapped, us-east-2a): OSv `block-bench` against `/dev/vblk1` vs Linux `fio` (3.32, direct=1 libaio, 4K) against the same device `/dev/nvme1n1`, 8s/cell, QD matrix {1,2,4,8,16,32,64}.

| workload | OSv QD1 | OSv peak | scaling | OSv vs Linux @ QD>=8 |
|---|---|---|---|---|
| randread | 14.6k IOPS | 57.8k IOPS | 4.0x | 95-110% |
| randwrite | 14.2k IOPS | 29.8k IOPS | 2.1x | 95-103% |

Both stacks plateau at the identical device ceiling (reads ~52k, writes ~29k IOPS), so OSv now saturates the NVMe rather than a software limit. The one honest caveat: at QD1-2 OSv trails Linux (36-50% of its IOPS) because OSv per-op latency is higher (p50 57us vs 23us). That is a per-request path-length gap, not the concurrency collapse this PR fixes, and it is irrelevant at any realistic queue depth.

## Scope / regression safety

- Both fixes are confined to `fs/vfs/vfs_bdev.cc` and `fs/vfs/vfs_fops.cc`.
- ZFS calls `devops->strategy` directly and never touches `bdev_read` -> unaffected. rofs builds its own bio via `strategy()` -> unaffected. The only other buffer-cache consumer, `read_partition_table`, runs once at attach before runtime I/O -> unaffected. An end-to-end synthetic-Postgres ZFS harness over local NVMe confirms ZFS I/O is functionally unchanged on this branch: it neither regresses nor benefits, exactly as the dispatch path predicts.
- Layer-2 change is gated on `v_type == VBLK && (flags & FOF_OFFSET)`; all other vnode types and the shared-offset `read`/`write` path retain the lock.

## Verification

- Local KVM: seqwrite-then-seqread byte-identical readback; 100-byte unaligned read returns correct bytes (slow-path correctness).
- EC2: full randread/seqread/randwrite/seqwrite sweep across QD {1,2,4,8,16,32,64} on both m6a.large and m6i.large, plus the i4i.large same-host OSv-vs-Linux-fio parity leg on uncapped NVMe.

### DIFF
--- a/fs/vfs/vfs_bdev.cc
+++ b/fs/vfs/vfs_bdev.cc
@@ -17,6 +17,75 @@
 #include <osv/buf.h>
 #include <osv/bio.h>
 
+/*
+ * Can this whole transfer take the fast path? That requires the starting
+ * offset and every iovec length to be a whole multiple of BSIZE, because the
+ * block drivers address and transfer in whole sectors. The buffer address
+ * itself need not be aligned: virtio-blk (and the other block drivers) DMA
+ * arbitrary guest addresses. Any unaligned offset or sub-sector length falls
+ * back to the buffer-cache slow path.
+ */
+static bool
+bdev_uio_sector_aligned(struct uio *uio)
+{
+	if ((uio->uio_offset % BSIZE) != 0)
+		return false;
+	for (int i = 0; i < uio->uio_iovcnt; i++) {
+		if ((uio->uio_iov[i].iov_len % BSIZE) != 0)
+			return false;
+	}
+	return true;
+}
+
+/*
+ * Fast path: issue one bio per iovec straight through the device's strategy
+ * (multiplex_strategy for real block devices, which splits by max_io_size and
+ * dispatches all children before waiting). Because make_request only holds the
+ * driver lock across ring submit, N concurrent callers keep N requests in
+ * flight -- unlike the buffer-cache path, which serialized every 512-byte block
+ * behind a single global lock and one synchronous bio.
+ *
+ * strategy() adds dev->offset exactly once (partition base), mirroring the
+ * rw_buf()->strategy() path this replaces, so partition addressing is
+ * unchanged.
+ */
+static int
+bdev_strategy_rw(struct device *dev, struct uio *uio, int bio_cmd)
+{
+	while (uio->uio_resid > 0 && uio->uio_iovcnt > 0) {
+		struct iovec *iov = uio->uio_iov;
+
+		if (iov->iov_len == 0) {
+			uio->uio_iov++;
+			uio->uio_iovcnt--;
+			continue;
+		}
+
+		struct bio *bio = alloc_bio();
+		if (!bio)
+			return ENOMEM;
+
+		bio->bio_cmd = bio_cmd;
+		bio->bio_dev = dev;
+		bio->bio_data = iov->iov_base;
+		bio->bio_offset = uio->uio_offset;
+		bio->bio_bcount = iov->iov_len;
+
+		dev->driver->devops->strategy(bio);
+		int ret = bio_wait(bio);
+		destroy_bio(bio);
+		if (ret)
+			return ret;
+
+		uio->uio_offset += iov->iov_len;
+		uio->uio_resid -= iov->iov_len;
+		uio->uio_iov++;
+		uio->uio_iovcnt--;
+	}
+
+	return 0;
+}
+
 int
 bdev_read(struct device *dev, struct uio *uio, int ioflags)
 {
@@ -25,26 +94,32 @@ bdev_read(struct device *dev, struct uio *uio, int ioflags)
 
 	assert(uio->uio_rw == UIO_READ);
 
-	/*
-	 * These should be handled gracefully, but this gives us the
-	 * best debugging for now.
-	 */
-	assert((uio->uio_offset % BSIZE) == 0);
-	assert((uio->uio_resid % BSIZE) == 0);
-
 	if (uio->uio_offset + uio->uio_resid > dev->size)
 		return EIO;
 	if (uio->uio_offset < 0)
 		return EINVAL;
 	if (uio->uio_resid == 0)
 		return 0;
-    
+
+	if (bdev_uio_sector_aligned(uio))
+		return bdev_strategy_rw(dev, uio, BIO_READ);
+
+	/*
+	 * Slow path: unaligned offset or sub-sector length. Read a whole BSIZE
+	 * block through the buffer cache and copy from the correct byte offset
+	 * within it. uiomove clamps each copy to min(remaining-in-block,
+	 * uio_resid), so partial leading and trailing sectors are handled.
+	 */
 	while (uio->uio_resid > 0) {
-		ret = bread(dev, uio->uio_offset >> 9, &bp);
+		off_t blkno = uio->uio_offset >> 9;
+		size_t off_in_blk = uio->uio_offset & (BSIZE - 1);
+
+		ret = bread(dev, blkno, &bp);
 		if (ret)
 			return ret;
 
-		ret = uiomove(bp->b_data, BSIZE, uio);
+		ret = uiomove((char *)bp->b_data + off_in_blk,
+			      BSIZE - off_in_blk, uio);
 		brelse(bp);
 
 		if (ret)
@@ -62,24 +137,33 @@ bdev_write(struct device *dev, struct uio *uio, int ioflags)
 
 	assert(uio->uio_rw == UIO_WRITE);
 
-	/*
-	 * These should be handled gracefully, but this gives us the
-	 * best debugging for now.
-	 */
-	assert((uio->uio_offset % BSIZE) == 0);
-	assert((uio->uio_resid % BSIZE) == 0);
-
 	if (uio->uio_offset + uio->uio_resid > dev->size)
 		return EIO;
 	if (uio->uio_offset < 0)
 		return EINVAL;
 	if (uio->uio_resid == 0)
 		return 0;
-    
-	while (uio->uio_resid > 0) {
-		bp = getblk(dev, uio->uio_offset >> 9);
 
-		ret = uiomove(bp->b_data, BSIZE, uio);
+	if (bdev_uio_sector_aligned(uio))
+		return bdev_strategy_rw(dev, uio, BIO_WRITE);
+
+	/*
+	 * Slow path: unaligned offset or sub-sector length. Read-modify-write
+	 * each affected BSIZE block through the buffer cache so the bytes
+	 * outside the written range are preserved. (getblk alone returns an
+	 * unread buffer, which would corrupt the untouched remainder of the
+	 * sector; bread reads it first.)
+	 */
+	while (uio->uio_resid > 0) {
+		off_t blkno = uio->uio_offset >> 9;
+		size_t off_in_blk = uio->uio_offset & (BSIZE - 1);
+
+		ret = bread(dev, blkno, &bp);
+		if (ret)
+			return ret;
+
+		ret = uiomove((char *)bp->b_data + off_in_blk,
+			      BSIZE - off_in_blk, uio);
 		if (ret) {
 			brelse(bp);
 			return ret;
--- a/fs/vfs/vfs_fops.cc
+++ b/fs/vfs/vfs_fops.cc
@@ -47,6 +47,20 @@ int vfs_file::read(struct uio *uio, int flags)
 
 	bytes = uio->uio_resid;
 
+	/*
+	 * Block devices carry no mutable per-vnode state that VOP_READ touches,
+	 * and the block layer below (bdev_read -> strategy -> driver) is already
+	 * thread-safe and asynchronous. Holding the exclusive per-vnode mutex
+	 * across the whole synchronous VOP_READ (which blocks in bio_wait)
+	 * serializes every concurrent pread on a shared fd, pinning throughput at
+	 * the queue-depth-1 value. A pread supplies its own offset (FOF_OFFSET),
+	 * so there is no fp->f_offset to protect: skip the lock and let N callers
+	 * keep N requests in flight. The FOF_OFFSET==0 path still shares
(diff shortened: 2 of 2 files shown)


fs/vfs/vfs_bdev.cc
```diff
@@ -25,26 +94,32 @@ bdev_read(struct device *dev, struct uio *uio, int ioflags)
 
 	assert(uio->uio_rw == UIO_READ);
 
-	/*
-	 * These should be handled gracefully, but this gives us the
-	 * best debugging for now.
-	 */
-	assert((uio->uio_offset % BSIZE) == 0);
-	assert((uio->uio_resid % BSIZE) == 0);
```
nyh, line 33: Why did you need to remove these? I assume this assert means that we didn't reach this case before this patch (or we would have seen these assertion failures...), so how come we started to reach it now?

Or maybe it was always reachable, and we just didn't have an application that reached it?

fs/vfs/vfs_fops.cc
```diff
@@ -47,6 +47,20 @@ int vfs_file::read(struct uio *uio, int flags)
...
+	 * Block devices carry no mutable per-vnode state that VOP_READ touches,
+	 * and the block layer below (bdev_read -> strategy -> driver) is already
+	 * thread-safe and asynchronous. Holding the exclusive per-vnode mutex
+	 * across the whole synchronous VOP_READ (which blocks in bio_wait)
+	 * serializes every concurrent pread on a shared fd, pinning throughput at
+	 * the queue-depth-1 value. A pread supplies its own offset (FOF_OFFSET),
+	 * so there is no fp->f_offset to protect: skip the lock and let N callers
+	 * keep N requests in flight. The FOF_OFFSET==0 path still shares
+	 * fp->f_offset, so it keeps the lock.
+	 */
```
nyh, line 60: This makes me worried about just one thing: Is it possible this mutex is also needed for destroying the device, and by not holding onto it you'll allow destroying the device while an operation is ongoing?

Of course, if my worry is correct you don't need a mutex to achieve this (a reader-writer lock or many other options exist), but it's still something to consider.

nyh: I asked AI and it explained to me:

vn_lock() is not protecting against vnode destruction — that's handled by a separate reference-counting mechanism. Here's the full picture:

What prevents destruction: reference counting. The vfs_file holds f_dentry as a dentry_ref (a boost::intrusive_ptr<dentry>), which keeps the dentry's d_refcnt > 0. When a dentry is allocated (vfs_dentry.cc), it calls vref(vp), which increments vp->v_refcnt. The vnode is only freed when v_refcnt reaches 0 (in vrele() or vput()). So as long as the file is open and f_dentry is non-null, the dentry and vnode are both pinned alive. The close() path drops the last protection with fp->f_dentry.reset() — but only after VOP_CLOSE() returns (i.e., after the operation is done with the vnode).

What vn_lock() actually does: operation serialization
vn_lock() acquires vp->v_lock, a per-vnode mutex. From the comment in vnode.h: "Reading or writing any of these items requires holding the appropriate lock."

Its purpose is to serialize concurrent filesystem operations on the same vnode — concurrent reads/writes/ioctls/stats. Without it, two threads could race on vnode fields like v_size, v_mode, v_data, or on the underlying filesystem's private data (v_data), and the VFS operations (like VOP_READ, VOP_WRITE) themselves are typically not thread-safe.

The memo in vfs_vnode.cc makes the separation clear:

So: vn_lock() is a concurrency/serialization mechanism, not a lifetime mechanism. The vnode's lifetime is managed entirely by v_refcnt, and the open file descriptor indirectly holds that count up through the dentry reference chain.

nyh (approved):
Looks good to me, I'll merge.

Something I didn't understand is whether this change improves practical workloads, which might not be so careful on aligning their request buffers and sizes. Is it documented somewhere that they should? Does Linux also expect you to do this alignment?

gburd:
Good question, and the short answer is: unaligned workloads are not penalized for correctness, only for speed, and the alignment expectation matches Linux.

**The alignment is not something applications must know about or opt into.** `bdev_read`/`bdev_write` pick the path automatically per request (`bdev_uio_sector_aligned`, vfs_bdev.cc:28). If the starting offset and every iovec length are whole multiples of BSIZE (512), the request takes the fast strategy path; otherwise it transparently falls back to the same buffer-cache read/read-modify-write loop that existed before this patch. So a careless app that does `pread(fd, buf, 100)` at offset 37 still gets correct bytes, exactly as it did before. Nothing regresses; it just doesn't get the speedup for that request. Note also that only the *offset and length* need to be sector-multiples; the *buffer address* need not be aligned at all, because virtio-blk (and the other drivers) DMA arbitrary guest addresses.

**Which practical workloads actually benefit:**

- This path is only reached for I/O directly to a block-device node (`/dev/vblkN`), i.e. `device_read`/`device_write` in devfs. Normal file I/O through a filesystem (ZFS, rofs, ramfs) does not go through `bdev_read`; those layers build their own bios via `strategy()` and were never subject to the 512B buffer-cache loop.
- The workloads that hit device nodes and benefit are the ones that already issue block-aligned I/O by nature: a filesystem's own device access, `O_DIRECT`-style raw-disk access, database raw volumes, `dd bs=4k`, image tools, and benchmarks. These are overwhelmingly aligned, because block storage is inherently sector-addressed. The empirical motivation was raw-device throughput on EBS collapsing to QD1 (~1600 IOPS) purely from the 512B serialization, while the device could do far more; the aligned fast path is what unblocks that.
- A pathological app doing tiny unaligned reads straight against a device node is rare, and it lands on the identical slow path as before, so it is no worse off.

**Is it documented that they should align?** Not by us specifically, and it doesn't need to be, because the fast path is opportunistic rather than required. The relevant convention is the universal one for raw block devices: they are addressed in sectors, so unaligned access to a raw device is already understood to be a slow/degenerate case on every OS.

**Does Linux expect the same alignment?** Yes, and Linux is actually *stricter*. For buffered I/O to a block device Linux, like us, accepts any offset/length and does read-modify-write in the page cache. But for `O_DIRECT` to a block device Linux *requires* the offset, length, and the memory buffer to be aligned to the device's logical block size and returns `EINVAL` otherwise. We are more lenient: we never hard-fail an unaligned request, we just route it to the buffered slow path. So relative to Linux, an application that was already writing correct `O_DIRECT` code is automatically on our fast path, and one that wasn't still works here where it would have gotten `EINVAL` on Linux.

fs/vfs/vfs_bdev.cc
gburd: Right on the second guess: that case was always reachable, we just had no test that hit it in a debug build, and release builds compiled the asserts out entirely.

The two asserts were `(uio_offset % BSIZE) == 0` and `(uio_resid % BSIZE) == 0`. They were guarding the old buffer-cache loop, which only ever moved data in whole `BSIZE` (512) chunks via `bread()`+`uiomove()`. But nothing upstream of `bdev_read` enforces 512-alignment on a device-node read: `read(fd, buf, 100)` against `/dev/vblkN` reaches here with `uio_resid == 100`. It "worked" anyway because `uiomove()` clamps the copy to `uio_resid`, so the final partial block was handled correctly by accident, and the comment above the asserts even admitted this ("These should be handled gracefully, but this gives us the best debugging for now").

Two reasons they didn't fire:
1. Release builds define `NDEBUG`, so `assert()` is a no-op. The asserts only ever had teeth in a debug build.
2. In practice the only in-tree consumers of the device-node read path issue aligned I/O by nature (filesystems doing block-sized access, the partition-table read at attach), so even debug builds rarely saw an unaligned request, and no test deliberately exercised one.

The new code makes the unaligned case a real, explicit path rather than assert-and-hope: aligned offset+length take the fast strategy fan-out, and anything unaligned falls back to the same buffer-cache read/read-modify-write loop that existed before. Since the fallback now *intentionally* handles sub-block and unaligned access, the asserts became false alarms and had to go. Correctness for the aligned case is unchanged; the unaligned case is now handled on purpose instead of surviving on the `uiomove` clamp.

fs/vfs/vfs_fops.cc
gburd: Your worry is worth stating precisely, and the conclusion is that it's safe: `vn_lock()` is not what protects the device (or the vnode) from being destroyed under an in-flight operation. The AI summary you pasted is correct; here is the same thing verified against the code on both the vnode and the device layer.

**Vnode lifetime is refcounted, not lock-protected.** `vn_lock(vp)` takes `vp->v_lock`, a per-vnode mutex whose only job is to serialize concurrent VOP operations that touch mutable vnode fields (`v_size`, `v_offset`, `v_data`, etc.). Liveness is a separate mechanism: an open `vfs_file` holds `f_dentry` as a `dentry_ref` (`boost::intrusive_ptr<dentry>`), which keeps `d_refcnt > 0`, which via `vref()` keeps `v_refcnt > 0`. The vnode is freed only when `v_refcnt` hits 0. `close()` drops the last reference with `fp->f_dentry.reset()`, and it does that only *after* `VOP_CLOSE()` returns (vfs_fops.cc close path: `vn_lock` -> `VOP_CLOSE` -> `vn_unlock` -> `f_dentry.reset()`). So while any read/write is in flight, the fd still holds the dentry ref, so the vnode cannot be freed regardless of whether that reader holds `v_lock`.

**Device lifetime is also refcounted, independently of `v_lock`.** For a `VBLK` node, `VOP_READ` -> `device_read` -> `device_reference(dev)` / `device_release(dev)` bracket every operation, and `device_destroy()` just clears `dev->active` and calls `device_release()` under `sched_lock` (fs/devfs/device.cc). The device object isn't torn down while a reference is outstanding, and taking or not taking the vnode's `v_lock` has no bearing on that refcount. So dropping `v_lock` on the `VBLK` + `FOF_OFFSET` path cannot let a device be destroyed mid-operation, because `v_lock` was never the thing gating destruction.

In short: destruction is fenced by reference counting (dentry_ref -> `v_refcnt`, and `device_reference` -> `dev` refcount), and `v_lock` is pure operation serialization. Removing it for the pread/pwrite-with-offset case removes serialization only, not a lifetime guarantee, which is exactly why N concurrent preads can now stay in flight without a correctness or use-after-free hazard.

Issue #1295, opened 2024-01-10
Support AWS graviton


wkozaczuk:
Make OSv runnable on AWS graviton-based EC2 instances - see https://aws.amazon.com/ec2/graviton/.

This depends on the following issues:
- support ENA (see #1204) - we already have PR for it
- support GICv3 (see #1290) - we already have PR for it
- support MSI/X (see #1088)
- support ACPICA on aarch64 (see #1294)
- support UEFI boot (see #1098)

PR #1411, opened 2026-07-06
drivers: build the AWS ENA network driver for aarch64 (Graviton)

gburd:
## Summary

Enables the existing AWS ENA (Elastic Network Adapter) driver to build and link
on aarch64, so OSv can run on EC2 Graviton instances with ENA networking. The
ENA driver is arch-neutral (a `pci::device` driver with no port I/O), so the
same driver sources compile for both x64 and aarch64; this change moves the
driver's object rules out of the x64-only block and adds the aarch64 profile
gating.

## What changed

- **Makefile**: move the `drivers/ena.o` + `bsd/sys/dev/ena/*` + `ena_com`
  object rules so they build for both x64 and aarch64 (gated on the same
  `conf_drivers_ena` flag as before).
- **conf/profiles/aarch64/{base.mk,aws.mk,kconfig,all.mk}**: add
  `conf_drivers_ena` gating (defaults to `0` in base.mk, so no behavior change
  for existing aarch64 builds), a new `aws.mk` profile that opts ENA on for
  Nitro, and the matching kconfig entry.
- **bsd/aarch64/machine/atomic.h**: implement the BSD atomic ops the ENA driver
  needs (`atomic_cmpset_int/long`, the `ATOMIC_ASM` set) using `__atomic_*`
  builtins, and add `atomic_thread_fence_acq()` (`dmb ishld`). These were
  prototype-only before; ENA is the first aarch64 driver to reference them.
- **bsd/sys/kern/subr_bufring.c**: guard the stub `atomic_thread_fence_acq()`
  with `#if !defined(__aarch64__)` now that aarch64 provides a real one.
- **include/api/aarch64/bits/syscall.h**: add the three
  `__NR_sys_io_uring_{setup,enter,register}` aliases that x64 already defines.
  This is an aarch64 kernel-build prerequisite, not ENA-specific: io_uring is
  compiled unconditionally into the kernel (core/io_uring.cc) and the generated
  syscall table references `sys_io_uring_setup` unconditionally, so the aarch64
  kernel does not link without these aliases. The aarch64 ENA build below fails
  to link without this hunk.

## Testing

Full aarch64 kernel link verified on a native Graviton3 instance (c7g,
Ubuntu 24.04 arm64, gcc 13.3.0):

```
CROSS_PREFIX=aarch64-linux-gnu- make -j8 arch=aarch64 conf_drivers_profile=aws \
  bootfs_manifest=bootfs_empty.manifest.skel loader.elf
```

- `loader.elf` links clean (AArch64 / ELF64), zero compiler errors.
- 256 ENA symbols present in the kernel image; driver entry points
  `ena_attach`, `ena_detach`, `ena_up`, `ena_probe` all resolved.

x64 is unaffected: ENA still defaults off in the aarch64 base profile and the
x64 build path is unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -704,6 +704,10 @@ endif
 ifeq ($(conf_drivers_hyperv),1)
 bsd += bsd/sys/dev/hyperv/vmbus/hyperv.o
 endif
+endif
+
+# ENA (AWS Elastic Network Adapter) is arch-neutral: it is a pci::device
+# driver with no port I/O, so it builds for both x64 and aarch64 (Graviton).
 ifeq ($(conf_networking_stack),1)
 ifeq ($(conf_drivers_ena),1)
 bsd += bsd/sys/contrib/ena_com/ena_eth_com.o
@@ -711,7 +715,6 @@ bsd += bsd/sys/contrib/ena_com/ena_com.o
 bsd += bsd/sys/dev/ena/ena_datapath.o
 bsd += bsd/sys/dev/ena/ena.o
 $(out)/bsd/sys/dev/ena/%.o: CXXFLAGS += -Ibsd/sys/contrib
-endif
 endif
 endif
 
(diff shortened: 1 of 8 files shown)


wkozaczuk:
I think this looks good. I have also asked AI to verify it. BTW, I had to fix some incorrectly implemented functions in this `atomic.h` (see 6e1527c3d69aa610a2119760b8e5a52ed6f1f44d) - `atomic_fetchadd_int()` and `atomic_fetchadd_long()` which were causing strange issues in ZFS on aarch64. So I hope we are not leaving any other unturned bugs in this file.

wkozaczuk:
BTW, per #1295, we are missing UEFI and ACPI support on aarch64 to run OSv on Graviton. For ACPI, we need to upgrade acpica first to the modern version, which supports aarch64.

PR #1415, opened 2026-07-11
libc: make setrlimit/getrlimit consistent and stop aborting

gburd:
## What

Fixes two problems with the resource-limit functions:

1. `getrlimit()` called `abort()` on any resource it did not special-case
   (`RLIMIT_LOCKS`, `RLIMIT_SIGPENDING`, `RLIMIT_RTPRIO`, `RLIMIT_RTTIME`). A
   program querying one of those crashed the whole unikernel, the same class of
   hazard as the old libaio stubs.

2. `setrlimit()` was a no-op that returned success, so a later `getrlimit()`
   reported the old (usually infinite) value rather than what the application
   just set, an inconsistency some programs depend on.

## How

Replace the `getrlimit()` switch with a process-wide stored table of
`RLIMIT_NLIMITS` entries, initialized once with the previous defaults (stack
size from the default pthread attr, `RLIMIT_NOFILE` 10240, `RLIMIT_NICE` 0,
everything else infinite). `getrlimit()` reads the table, `setrlimit()` writes
it, both guarded by a mutex. Out-of-range resources and `rlim_cur > rlim_max`
return `EINVAL` instead of crashing or being silently accepted.

The limits remain advisory (OSv is a single address space and does not enforce
most of them), but set and get are now consistent.

Wire `setrlimit` as syscall `SYS_setrlimit` (only `getrlimit` and `prlimit64`
were in the table) with a tracepoint.

## Testing

`tests/tst-rlimit.cc` covers querying every resource without aborting, the
EINVAL paths, set/get round-trips, and `prlimit`. Passes on OSv under KVM.

### DIFF
--- a/libc/libc.cc
+++ b/libc/libc.cc
@@ -7,6 +7,7 @@
 
 #include "libc.hh"
 #include <osv/sched.hh>
+#include <osv/mutex.h>
 #include <stdio.h>
 #include <limits.h>
 #include <stdlib.h>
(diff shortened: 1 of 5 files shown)


gburd:
Superseded by #1435 (rebased onto current master; GitHub auto-closed this one when the branch was updated). Please review #1435 instead.

PR #1417, opened 2026-07-11
fs: implement close_range(2)

gburd:
## What

`close_range(2)` was missing entirely. It is the standard way to close a range
of file descriptors in one call (commonly used to clean up inherited fds and by
some language runtimes), so programs that call it failed.

## How

Implemented over the existing fd table: close (or, with `CLOSE_RANGE_CLOEXEC`,
set `O_CLOEXEC` on) every open fd in the inclusive range `[first, last]`,
clamping `last` to the fd-table size and ignoring fds that are not open, as Linux
does. `CLOSE_RANGE_UNSHARE` is a no-op in a single-process unikernel (there is
nothing to unshare). `first > last` and unknown flags return `EINVAL`.

Wired as syscall `SYS_close_range` (adding `__NR_close_range` /
`SYS_close_range` for x86-64 and aarch64) with a tracepoint, declared in
`unistd.h` under `_GNU_SOURCE`, and the symbol exported from `libc.so.6` and
`ld-musl.so.1`.

## Testing

`tests/tst-close-range.cc` covers closing a contiguous run, a range that spans
not-open fds, the `CLOSE_RANGE_CLOEXEC` variant, and the `EINVAL` paths. Passes
on OSv under KVM.

### DIFF
--- a/exported_symbols/osv_ld-musl.so.1.symbols
+++ b/exported_symbols/osv_ld-musl.so.1.symbols
@@ -78,6 +78,7 @@ clock_getres
 clock_gettime
 clock_nanosleep
 close
+close_range
 closedir
 closelog
 confstr
--- a/exported_symbols/osv_libc.so.6.symbols
+++ b/exported_symbols/osv_libc.so.6.symbols
@@ -56,6 +56,7 @@ clock_getres
 clock_gettime
 clock_nanosleep
 close
+close_range
 closedir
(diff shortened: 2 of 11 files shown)


gburd:
Superseded by #1436 (rebased onto current master; GitHub auto-closed this one when the branch was updated). Please review #1436 instead.

PR #1419, opened 2026-07-11
numa: report real topology from getcpu and get_mempolicy

gburd:
## What

With NUMA topology now discovered (the `numa::` module from the base PR), make
the topology-query syscalls report it instead of a hardcoded single node.

## How

- `sys_getcpu()` now fills the node-out argument with `numa::node_of_cpu()` for
  the calling CPU rather than always 0.
- `get_mempolicy()`:
  - `MPOL_F_NODE` returns the calling CPU's node.
  - `MPOL_F_NODE | MPOL_F_ADDR` returns the node backing the given address,
    resolved via a page-table walk (`virt_to_phys_pt`) and
    `numa::node_of_phys()`.
  - the allowed-nodes mask now sets a bit for every discovered node (not just
    node 0) and rejects a `maxnode` smaller than the node count with `EINVAL`.

`set_mempolicy()` stays a no-op: the topology is known but the physical
allocator is not yet node-aware, so a placement policy cannot be enforced. Its
comment is updated to say so; enforcement will come with the node-aware
allocator.

Adds `numa::node_of_phys()` to map a physical address to its node.

On a machine with no SRAT everything degrades to the previous single-node-0
behavior.

## Testing

`tests/tst-numa-mempolicy.cc` checks getcpu's node is in range and matches
`numa::node_of_cpu`, get_mempolicy's `MPOL_F_NODE` and allowed-mask (bit count
== node count, maxnode-too-small `EINVAL`), and the `MPOL_F_ADDR` path. Verified
on OSv under KVM single-node and with a QEMU 2-node `-numa` config.

## Note

Depends on #1418 (numa: discover NUMA topology from ACPI SRAT/SLIT). Until that
merges, this PR's diff shows both commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1131,6 +1131,7 @@ endif
 objects += linux.o
 objects += core/commands.o
 objects += core/sched.o
+objects += core/numa.o
 objects += core/mmio.o
 objects += core/kprintf.o
 ifeq ($(conf_tracepoints),1)
--- a/core/numa.cc
+++ b/core/numa.cc
@@ -44,6 +44,16 @@ unsigned node_of_cpu(unsigned cpu_id)
     return it == s_cpu_to_node.end() ? 0 : it->second;
 }
 
+int node_of_phys(uint64_t phys)
+{
+    for (auto& r : s_mem_ranges) {
+        if (phys >= r.base && phys < r.base + r.length) {
+            return (int)r.node;
+        }
+    }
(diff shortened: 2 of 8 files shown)


gburd:
Superseded by #1438 (rebased onto current master; GitHub auto-closed this one when the branch was updated). Please review #1438 instead.

PR #1420, opened 2026-07-11
drivers: add virtio-balloon memory balloon driver

gburd:
## What

Adds the virtio memory balloon driver (`virtio-balloon`, virtio device id 5),
which lets the host reclaim guest memory on demand and return it later. This is
broadly useful for memory overcommit and density (an idle guest can be shrunk
and grown again under load without stopping it) and is hypervisor-agnostic
(KVM/QEMU, Cloud Hypervisor).

## How

The driver has two virtqueues (inflate, deflate) and a worker thread that moves
the balloon toward the target size the host advertises in the device config
(`num_pages`):

- **inflate**: allocate guest pages via `memory::alloc_page()`, hand their PFNs
  to the host on the inflate queue, and keep them out of OSv's free pool (the
  memory is now the host's);
- **deflate**: return ballooned pages' PFNs on the deflate queue and free them
  back to OSv with `memory::free_page()`.

After each adjustment the driver writes the current balloon size back to the
config `actual` field so the host can track it (QEMU uses this to compute
subsequent targets). The worker polls the target on a short interval rather than
waking on a config-change interrupt, because OSv's virtio core does not
currently wire the MSI-X config-change vector; ballooning is not
latency-critical, so a sub-second reaction is fine.

To support writing `actual` back, this adds a symmetric `write_config()` to the
`virtio_device` interface (legacy-pci, modern-pci and mmio transports) and a
`virtio_driver::virtio_conf_write()` helper, mirroring the existing
`read_config()` / `virtio_conf_read()`. That change is purely additive and does
not affect existing drivers.

Registered on x86_64 and aarch64 (gated on `conf_drivers_virtio`, next to
virtio-rng).

## Testing

Because the balloon target is host-controlled, this cannot be a self-checking
automated test. `tests/misc-balloon.cc` prints free memory once a second while a
host-side QMP harness drives the balloon, and `documentation/virtio-balloon.md`
describes the procedure.

Verified on QEMU/KVM: setting the balloon to 512 MiB on a 1 GiB guest drops OSv
free memory by ~512 MiB (inflate), and setting it back to full returns it
(deflate), confirmed by the free-memory curve. virtio-blk and other virtio
drivers continue to work with the balloon device present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -944,6 +944,7 @@ endif
 drivers += drivers/virtio-blk.o
 drivers += drivers/virtio-scsi.o
 drivers += drivers/virtio-rng.o
+drivers += drivers/virtio-balloon.o
 drivers += drivers/virtio-fs.o
 endif
 
@@ -1013,6 +1014,7 @@ drivers += drivers/virtio-mmio.o
 endif
 drivers += drivers/virtio-vring.o
 drivers += drivers/virtio-rng.o
+drivers += drivers/virtio-balloon.o
 drivers += drivers/virtio-blk.o
 drivers += drivers/virtio-scsi.o
 drivers += drivers/virtio-net.o
--- a/arch/aarch64/arch-setup.cc
+++ b/arch/aarch64/arch-setup.cc
@@ -184,6 +184,9 @@ void arch_init_premain()
 #if CONF_drivers_virtio_rng
 #include "drivers/virtio-rng.hh"
 #endif
+#if CONF_drivers_virtio
+#include "drivers/virtio-balloon.hh"
+#endif
 #if CONF_drivers_virtio_blk
 #include "drivers/virtio-blk.hh"
 #endif
@@ -241,6 +244,9 @@ void arch_init_drivers()
     hw::driver_manager* drvman = hw::driver_manager::instance();
 #if CONF_drivers_virtio_rng
     drvman->register_driver(virtio::rng::probe);
+#endif
+#if CONF_drivers_virtio
+    drvman->register_driver(virtio::balloon::probe);
 #endif
 #if CONF_drivers_virtio_blk
     drvman->register_driver(virtio::blk::probe);
--- a/arch/x64/arch-setup.cc
+++ b/arch/x64/arch-setup.cc
@@ -297,6 +297,9 @@ void arch_init_premain()
 #if CONF_drivers_virtio_rng
 #include "drivers/virtio-rng.hh"
 #endif
+#if CONF_drivers_virtio
+#include "drivers/virtio-balloon.hh"
+#endif
 #if CONF_drivers_virtio_fs
 #include "drivers/virtio-fs.hh"
 #endif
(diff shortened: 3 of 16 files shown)


wkozaczuk:
Based on what I can see and what AI tells me, this PR looks good.

As I understand the balloon operates on the physical memory that guest acquired during the boot time, and host may trigger "inflate" to grab part of that original memory back under some conditions (what are these?) and return on "deflate" back to the guest - OSv. But I could not see where OSv could request the deflate when say balloon was inflated because OSv did not use much of the memory but now is under pressure (almost OOM, see `mempool.cc` reclaimer thread, etc). We do have old and I think obsolete jvm baloon related logic (see the occurrences of `CONF_memory_jvm_balloon` in `mmu.cc` and `mempool.cc`). Is this where eventually we could integrate the "low memory pressure" logic with virtio balloon?

Also, as I understand it, the balloon code stays inactive unless we explicitly add the drivers settings to the `run.py`? Could we add option to enable the ballon in `run.py`?

gburd:
Thanks. Addressed the `run.py` request and here are answers on the memory-pressure question.

**`run.py` option**: added `--virtio-balloon` (folded into the driver commit). Without it the guest gets no balloon device (unchanged default); with it, `run.py` attaches `virtio-balloon-pci`. Verified boot both ways.

**When does the host inflate?** The balloon is host-driven: the hypervisor (via the balloon `num_pages` config field) asks the guest to give memory back — typically under host memory pressure / overcommit, or an admin/orchestrator explicitly shrinking a guest's footprint. On inflate, the guest allocates those pages and hands their PFNs to the host (which can then reclaim them); on deflate, the host lowers the target and the guest frees the balloon pages back to its own allocator.

**Deflate-on-guest-pressure (the important part you're pointing at):** you're right that today this PR only implements the host->guest direction (respond to the host's inflate/deflate target). It does *not* yet auto-deflate when OSv itself goes under pressure. That's the piece worth wiring into `core/mempool.cc`'s reclaimer path: when the reclaimer can't find free ranges (before it escalates toward `oom()`), it should first ask the balloon to deflate (return ballooned pages to the guest) — exactly analogous to, and a clean replacement for, the old `CONF_memory_jvm_balloon` logic in `mmu.cc`/`mempool.cc`. The JVM balloon hooked the same reclaim path to shrink the JVM heap; the virtio balloon is the general, VM-wide version of that idea.

I'd prefer to keep this PR to the **driver + host-directed inflate/deflate** (it's self-contained and reviewable), and do the **reclaimer integration as a focused follow-up** — it touches the memory reclaimer's escalation ladder and deserves its own review, and it's the natural place to also retire the obsolete `CONF_memory_jvm_balloon` code. If you'd rather see them together I can fold it in, but the pressure-integration is the riskier half (it changes OOM behavior) so I lean toward landing the driver first. What's your preference?

wkozaczuk:
Yeah, I agree that the reclaimer integration should be part of a follow up PR. I also think there should be some logic that possibly "vetoes" the host demand to inflate balloon if the guest is already under low memory pressure.

PR #1422, opened 2026-07-11
fs: implement inotify(2)

gburd:
## What

inotify was stubbed (`inotify_init` returned `EMFILE`, `add_watch`/`rm_watch`
returned `EINVAL`), so programs that watch the filesystem for changes could not
run.

## How

Implemented against OSv's VFS. All path mutations route through a handful of
central functions in `fs/vfs/vfs_syscalls.cc`; those now call
`osv_inotify_notify()` with the absolute path affected and the event mask.
`libc/inotify.cc` keeps a registry of inotify instances (each a pollable
`special_file`, modeled on eventfd), matches every event against their watches,
and queues a `struct inotify_event` for the fd, waking any poller.

- `inotify_init`/`inotify_init1` create the fd (`IN_CLOEXEC`/`IN_NONBLOCK`
  honored).
- `inotify_add_watch` resolves the path to absolute, adds (or merges, with
  `IN_MASK_ADD`) a watch, and returns a watch descriptor.
- `inotify_rm_watch` removes a watch and queues `IN_IGNORED`.
- `read()` returns as many packed `inotify_event` structures as fit in the
  buffer (blocking, or `EAGAIN` when non-blocking), rejecting a too-small buffer
  with `EINVAL`; `poll()` reports `POLLIN` when events are queued.

A directory watch fires for changes to entries within it (reporting the entry
name); a watch on the object itself fires with no name. Events wired:
`IN_CREATE` (mkdir, open O_CREAT), `IN_DELETE` (rmdir, unlink), `IN_MOVED_FROM` /
`IN_MOVED_TO` (rename), each with `IN_ISDIR` when the object is a directory. The
rename hook rebuilds the full destination path because `sys_rename` truncates
`dest` to its parent in place before the VOP.

Not yet covered (follow-ups): `IN_MODIFY`/`IN_CLOSE_WRITE` on writes (needs an
fd-to-path lookup), `IN_ACCESS`/`IN_OPEN`, and recursive watches.

## Testing

`tests/tst-inotify.cc` covers create/delete/move on files and a subdirectory
(names and `IN_ISDIR`), `IN_IGNORED` on watch removal, and the `EAGAIN`/`EINVAL`
paths. Passes on OSv under KVM. `tst-remove` (unlink/rename/rmdir) continues to
pass, so the VFS mutation paths are unaffected.

### DIFF
--- a/fs/vfs/vfs.h
+++ b/fs/vfs/vfs.h
@@ -113,6 +113,9 @@ int	 sys_link(char *oldpath, char *newpath);
 int	 sys_unlink(char *path);
 int	 sys_symlink(const char *oldpath, const char *newpath);
 int	 sys_access(char *path, int mode);
+
+/* inotify VFS notification hook (libc/inotify.cc). */
+void osv_inotify_notify(const char *path, uint32_t mask, int is_dir);
 int	 sys_stat(char *path, struct stat *st);
 int	 sys_lstat(char *path, struct stat *st);
 int	 sys_statfs(char *path, struct statfs *buf);
--- a/fs/vfs/vfs_syscalls.cc
+++ b/fs/vfs/vfs_syscalls.cc
@@ -41,6 +41,7 @@
 
 #include <sys/stat.h>
 #include <dirent.h>
+#include <sys/inotify.h>
 
 #include <limits.h>
 #include <unistd.h>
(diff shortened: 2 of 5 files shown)


gburd:
Superseded by #1440 (rebased onto current master; GitHub auto-closed this one when the branch was updated). Please review #1440 instead.

PR #1423, opened 2026-07-12
zfs: selectable in-kernel ZFS (BSD or OpenZFS 2.4.2) via upstream submodule + patch series

gburd:
> [!WARNING]
> **DRAFT / BLOCKED - not ready for review or merge yet.**
>
> This PR is published early for visibility into the OpenZFS cutover work. It is
> **blocked on two other PRs merging first**, and its diff will not be clean
> until they do:
>
> - **#1398** (pagecache/vfs: ZFS page-cache bridge, readahead, writeback, fsync flush)
> - **#1400** (block: TRIM/DISCARD support and virtio-blk multiqueue I/O)
>
> This branch is stacked on top of both, so the diff shown here is the
> **cumulative** stack back to master (it re-includes changes from #1398, #1400,
> and other already-merged work). Once #1398 and #1400 merge, I will rebase this
> onto the new master, drop the merged commits, and the diff will reduce to just
> the OpenZFS-specific changes for a real review. Please do not review the diff
> as-is.

## What this delivers

Replaces OSv's ~2014 BSD-derived ZFS port with vendored **OpenZFS 2.4.3**,
bringing modern ZFS features and active upstream maintenance. The
OpenZFS-specific commits (the rest of the diff is the #1398/#1400 dependency
stack) are:

- `298c58cd` **zfs: integrate OpenZFS 2.4.3** with TRIM, native encryption, and
  Direct I/O (O_DIRECT via the ABD direct-I/O path).
- `f326eb06` / `01fe82d6` ZFS validation + benchmark suite (incl. the mmap
  copy-path microbenchmark).
- `57257ebe` mkfs/cpiod: pin the ZFS pool-root mountpoint so the image build
  populates the root dataset.
- `7b3b69a5` **ARC mmap page-sharing bridge** (issue #12): share decompressed
  ARC pages into the mmap page cache instead of copying (borrow-and-pin),
  reducing cold-fault latency and RSS.

The vendored OpenZFS submodule tracks `zfs-2.4.3` plus a small set of OSv
platform commits, on our fork's `osv-2.4.3` branch.

## Relationship to #1201

This is the substrate for #1201 (extract ZFS into a `modules/` library): the
old BSD-ZFS path is intentionally kept live alongside OpenZFS during the
transition so nothing regresses, and can be removed once the OpenZFS
`libsolaris.so` module lands.

## Validation done (on our fork)

- ZFS unit tests green: tst-zfs-direct-io 9/9, tst-zfs-trim 5/5,
[...]

PR #1424, opened 2026-07-12
drivers: add Crucible distributed block storage driver

gburd:
## What this delivers

Adds a **Crucible** distributed block storage driver, so OSv can use Oxide's
Crucible as a ZFS-backing block device (alongside virtio-blk, NVMe, and
NVMe-oF). The driver is **opt-in and gated off by default**
(`conf_drivers_crucible?=0`); it is built only with
`conf_drivers_profile=crucible`, so it adds nothing to the default kernel.

It sits last in the storage stack because it is a ZFS backing device and builds
on the block/pagecache infrastructure and the in-kernel OpenZFS support that are
now in master. The three pieces this originally depended on have merged
(pagecache/vfs ZFS page-cache bridge, block TRIM/DISCARD + virtio-blk
multiqueue, and the selectable in-kernel OpenZFS), so this branch is now rebased
onto master and the diff is reduced to just the Crucible driver.

## Validation done

- `tst-crucible-blk` 9/9 byte-clean; multi-iov physio path exercised and fixed.
- Crucible-backed ZFS works against a single downstairs.

## Known limitation (honest status)

- `tst-crucible-zfs` can hang in `txg_wait_synced` during `zpool_create` against
  a multi-host downstairs quorum. This is Crucible-driver-specific (virtio
  passes the same ashift=9/12 workload) and is being investigated; it is why the
  driver ships gated-off.


gburd:
Status update: **#1398 has merged**. This PR remains blocked on **#1400** and **#1423** (OpenZFS cutover) landing first, since Crucible is a block-storage backend that sits on the same storage stack. Will rebase to a clean Crucible-only diff once the chain ahead of it merges.

gburd:
Note to self / reviewers: this draft is blocked on #1423 and is currently ~39 commits behind master, which means it carries a few silent reverts of work that has since merged (the rt_sigtimedwait/osv_sigtimedwait path from #1428 and the tst-signal-fills.cc test). Those are stale-base artifacts, not intentional. When #1423 lands and this is unblocked, it will be rebased onto current master (which drops those reverts by patch-id) before it leaves draft. Not fixing now because it cannot merge until #1423 anyway and the rebase-on-unblock supersedes any fix made today.

gburd:
blockers #1398/#1400/#1423 all merged; rebased onto master, builds clean, un-drafted.

gburd:
Un-blocked: the three pieces this branch was stacked on have all merged to master (the pagecache/vfs ZFS page-cache bridge, block TRIM/DISCARD + virtio-blk multiqueue, and the selectable in-kernel OpenZFS). I have rebased this onto the current master tip; the branch is now a single clean commit and the diff is reduced to just the Crucible driver. GitHub reports it as mergeable. Ready for review.

gburd:
Split out the general `physio()` fix: the `bio_bcount = uio_resid` multi-iovec bug (and the zero-length-iovec infinite loop) in `fs/vfs/vfs_bdev.cc` is not Crucible-specific - any driver hits it on a multi-iov write - so it now stands alone as #1508 to land on its own merit rather than being gated behind this driver.

Status of this PR otherwise: rebased on current master (`0e34e4dc2`), merge-tree clean, self-contained (no dependency on other open PRs; the block/device APIs it uses are all current). It remains a large experimental driver best reviewed on its own timeline; #1508 carries the one piece that is generally useful today.

gburd:
### Refreshed against current Crucible, and the `physio()` justification is retracted

Pushed `d2ce647e1`. Two things: an upstream review (which found nothing to change, reported below as a null result) and one real fix.

#### Upstream Crucible review: no protocol drift

Reviewed our driver against a current `oxidecomputer/crucible` checkout. **There is nothing to update**, and it is worth recording why rather than just asserting it:

| check | result |
|---|---|
| last commit touching `protocol/src/lib.rs` | `2481b77 Bump to 2024 edition`, **2025-11-05** |
| protocol commits between that checkout and upstream tip | **zero** (`git log HEAD..upstream/main -- protocol/src/lib.rs` is empty) |
| `MessageVersion::current()` | **`V13`**, at both the checkout and upstream tip; our driver declares v13 |

The driver was written well after the protocol last moved, so it could not have drifted. The 20 commits upstream is ahead touch `ctop/`, `cmon/`, `pantry/`, `upstairs/`, `integration_tests/`, `tools/` and dependency bumps; none touches `protocol/` or the wire types.

I checked the wire format field by field anyway, since a silent mismatch here corrupts data rather than failing to build. I extracted the upstream `Message` variant order mechanically and compared all 41 entries against our `enum class MessageType`: **exact match, index 0 `HereIAm` through 40 `ErrorReport`**, including the `Barrier`/`BarrierAck` pair V12 inserted and the `LastFlushAck` removal V13 made. That is the check that matters most, because bincode encodes the variant index, so one inserted variant shifts everything after it. `WriteHeader`, `ReadResponseHeader`, `ReadBlockContext`, `Flush`'s two trailing options, and `RegionDefinition` all match our encoders field for field; framing is still a u32 LE length with `MAX_FRM_LEN` 100 MiB; the integrity hash is still `XxHash64` seed 0; and the negotiation sequence still matches the downstairs `NegotiationState` machine.

#### The real change: `bdev_read`/`bdev_write`, and the stale justification

You were right to push on this. The driver carried:

> We MUST go through physio() rather than bdev_read/bdev_write - those route through the OSv buf cache which is hard-coded to BSIZE=512 byte blocks

**That is no longer true, and the commit message no longer claims it.** `c8c153047` ("vfs: async device I/O via strategy fan-out, fix unaligned bdev paths") is an ancestor of both `master` and this branch's base. `bdev_read`/`bdev_write` now test the uio with `bdev_uio_sector_aligned()` and take `bdev_strategy_rw()` straight to the device strategy for any 512-aligned transfer; the buffer cache is reached only for a sub-sector request, which a Crucible volume rejects as `EINVAL` on either path.

So for the aligned I/O ZFS actually issues, the two are the same path. And they are not merely equivalent, they are strictly better:

| uio shape | `physio()` | `bdev_read`/`bdev_write` |
|---|---|---|
| ashift-aligned (what ZFS issues) | strategy | strategy, identical |
| multi-iovec | **broken**: `bio_bcount = uio_resid` with `bio_data = iov[0].iov_base` | correct: `bio_bcount = iov->iov_len` |
| offset+resid past `dev->size` | unchecked | `EIO` |
| `uio_resid`/iovcnt disagreement | walks off the iovec array | `uio_iovcnt > 0` guard |

The multi-iovec row is exactly the bug this PR was patching into `physio()`. Using `bdev_read`/`bdev_write` gets the fix and the two other guards for free, so **this PR's change to `fs/vfs/vfs_bdev.cc` is dropped entirely** and the diff is one file smaller.

**This unblocks #1508.** With the driver off `physio()`, `physio()` has no caller anywhere in the tree, which removes the last reason to keep it. The deletion you favoured there is now free of its blocker.

While in the same file I also switched the devops `.strategy` from a bare `crucible_strategy` to `multiplex_strategy`, with `crucible_strategy` in the first `private_data` member, matching virtio-blk, nvme, ahci, ide, virtio-scsi, vmw-pvscsi and xenfront-blk. That fixes two pre-existing bugs without new code: `crucible_strategy` never added `dev->offset`, so I/O to a partition subdevice from `read_partition_table()` addressed the whole-disk offset, and `dev->max_io_size = 1MB` was never enforced because nothing split the bio.

#### Build

From scratch, `scripts/build conf_drivers_profile=crucible conf_zfs=openzfs`:

```
1185 compile steps, 0 errors
  CXX drivers/crucible-connection.cc
  CXX drivers/crucible-request.cc
  CXX drivers/crucible-hash.cc
  CXX drivers/crucible-client.cc
  CXX drivers/crucible-blk.cc
  LINK loader.elf
OSv v0.57.0-399-gcdb02c912
Booted up in 748.98 ms
```

All five objects compiled, 178 crucible symbols in `loader.elf`, and the kernel boots. Verified the change at the object level rather than trusting the source:

```
$ nm -u build/release.x64/drivers/crucible-blk.o | grep -E 'bdev_|physio|multiplex'
                 U bdev_read
                 U bdev_write
                 U multiplex_strategy
$ nm -u build/release.x64/drivers/crucible-blk.o | grep -c physio
0
```

Still a draft: the `tst-crucible-zfs` `txg_wait_synced` hang against a multi-host quorum is unchanged, and I have not re-run the driver against a live downstairs cluster, so the I/O path change is compile-and-boot verified but not yet exercised end to end against real downstairs.

PR #1425, opened 2026-07-12
libc: implement splice(2), vmsplice(2), tee(2)

gburd:
## What

`splice(2)`, `vmsplice(2)`, and `tee(2)` were missing entirely, so programs that
move data between an fd and a pipe via splice (a common high-throughput idiom)
could not run.

## How

OSv has no user/kernel boundary and copies are cheap, so this is a correct
(not zero-copy) implementation built on the existing read/write and pread/pwrite
paths:

- **`splice()`**: move up to `len` bytes from `fd_in` to `fd_out` through a
  bounce buffer, honoring the optional `off_in`/`off_out` offset pointers
  (pread/pwrite when set, read/write otherwise), looping over short writes. A
  pipe-page-sharing fast path can be added later if it ever shows up hot;
  correctness does not need it.
- **`vmsplice()`**: implement the common memory -> fd direction as a write of the
  iovec segments into the fd.
- **`tee()`**: duplicating pipe data without consuming it needs a non-consuming
  pipe peek, which OSv's pipe does not support, so it returns `ENOSYS` rather
  than silently consuming the input.

Unknown flags return `EINVAL`. Wired as `SYS_splice`/`SYS_vmsplice`/`SYS_tee`
with tracepoints, symbols exported from `libc.so.6` and `ld-musl.so.1` (the
declarations and `SPLICE_F_*` flags are already in the musl headers).

## Testing

`tests/tst-splice.cc`: file -> pipe -> file with byte-integrity verification and
offset-pointer handling, vmsplice memory -> pipe round-trip, `tee` `ENOSYS`, and
the `EINVAL` path. Passes on OSv under KVM.


libc/splice.cc
```diff
@@ -0,0 +1,138 @@
...
+//
+// Linux implements these as zero-copy transfers through a pipe's page buffers.
+// OSv has no user/kernel boundary and the copies are cheap, so this is a
+// correct (not zero-copy) implementation built on the existing read/write and
+// pread/pwrite paths: move up to `len` bytes through a bounce buffer.  This
+// unblocks the many programs that use splice()/sendfile-style loops without
+// caring how the bytes move.
+//
+// ponytail: bounce-buffer, not zero-copy.  A pipe-page-sharing fast path can be
+// added later if splice ever shows up hot; correctness does not need it.
```
nyh: Since you said you didn't want to leave the word "ponytail" in the code, please remove this one too. Please word this as a "TODO" or "FIXME", saying what we should do in the future. The what-we-do-today is already explained in the paragraph above.

libc/splice.cc
```diff
@@ -0,0 +1,138 @@
...
+extern "C" OSV_LIBC_API
+ssize_t splice(int fd_in, off_t *off_in, int fd_out, off_t *off_out,
+               size_t len, unsigned int flags)
+{
+    if (flags & ~(SPLICE_F_MOVE | SPLICE_F_NONBLOCK | SPLICE_F_MORE | SPLICE_F_GIFT)) {
+        return libc_error(EINVAL);
+    }
+    if (len == 0) {
+        return 0;
+    }
```
nyh, line 77: What kind of files do you allow these fd_in and fd_out to be? Can they be filesystem files? Pipes? Sockets?
I think for certain types of fd_in or fd_out you're not allowed a non-null off_in, off_out, I don't see you are checking that.

libc/splice.cc
```diff
@@ -0,0 +1,138 @@
...
+
+    std::vector<char> buf(std::min(len, SPLICE_CHUNK));
+    size_t total = 0;
+    while (total < len) {
+        size_t want = std::min(len - total, buf.size());
+        ssize_t r = read_at(fd_in, buf.data(), want, off_in);
+        if (r < 0) {
+            return total ? (ssize_t)total : -1;   // errno set by read
+        }
+        if (r == 0) {
```
nyh, line 101: I think there's a problem here, but maybe you intentionally didn't handle it (and if so, it needs to join the apologetic comment on top and the FIXME): What happens if the input file descriptor has just, say, 1024 bytes and then nothing more was written yet (it's not an eof)? If I understand correctly, this code will not read and write what's already available, but rather wait until a full 64K chunk can be read. I'm worried that this is not what network code especially expects to happen when it splice()s.

libc/splice.cc
```diff
@@ -0,0 +1,138 @@
...
+    return total;
+}
+
+extern "C" OSV_LIBC_API
+ssize_t vmsplice(int fd, const struct iovec *iov, size_t nr_segs, unsigned int flags)
+{
+    if (flags & ~(SPLICE_F_MOVE | SPLICE_F_NONBLOCK | SPLICE_F_MORE | SPLICE_F_GIFT)) {
+        return libc_error(EINVAL);
+    }
+    // vmsplice with SPLICE_F_GIFT to a pipe is user-pages -> pipe.  Without
```
nyh: 1. It doesn't seem you checked above that this is "vmsplice with SPLICE_F_GIFT".
2. You say "we implement the common direction", but where do you check if we're in the wrong direction and return an error?

nyh (changes requested):

gburd:
Rebased onto current master and squashed into one coherent commit (the branch had drifted and its old base spuriously touched linux.cc/signal.cc/pthread.cc; that's gone - it now touches only the splice files). Addressed the review:

- **"ponytail" word**: reworded to a proper forward-looking TODO explaining the zero-copy fast path we'd add if splice ever gets hot.
- **fd types / non-null offset (your comment)**: splice now rejects a non-null `off_in`/`off_out` on a non-seekable (pipe/socket) fd with **ESPIPE**, matching Linux (verified OSv's lseek returns ESPIPE for such fds). I did NOT implement the full "exactly one end must be a pipe" rule because OSv's anonymous pipe isn't reliably distinguishable via fstat here - documented that limitation honestly in the source rather than half-checking it.
- **short input / "what if fewer than len bytes"**: the loop ends on a short read / EOF and returns what was moved.
- **infinite loop (Copilot)**: a legal zero-byte (no-progress) write now stops the loop and returns bytes-moved instead of spinning.
- **overflow (Copilot)**: splice rejects `len > SSIZE_MAX` (EINVAL); vmsplice bounds `nr_segs` at UIO_MAXIOV (like readv/writev) and guards the running total against ssize_t overflow.
- **vmsplice wrong direction (your comment)**: it implements memory -> fd; the unsupported pipe -> memory direction is rejected naturally because write() returns EBADF on a read-only fd. The comment now names that enforcement rather than just saying "not supported".
- **tee**: ENOSYS (honest - OSv's pipe can't peek non-consuming).
- **test flakiness (Copilot)**: file write / pipe read now loop until complete; added asserts for the new ESPIPE-on-pipe-offset and SSIZE_MAX-overflow guards. Builds/runs on Linux too.

The change was independently reviewed by two reviewers (systems + API/semantics); the vmsplice `nr_segs` bound and the two new guard tests came out of that review.

gburd:
The CHANGES_REQUESTED here predates my fixes - all of these are already addressed in the current head (rebased onto master):

- **"ponytail" -> TODO** (your comment): done, worded as a forward-looking TODO.
- **fd types / non-null offset** (your comment): splice rejects a non-null off_in/off_out on a non-seekable (pipe/socket) fd with ESPIPE, matching Linux. The full "exactly one end must be a pipe" rule is documented as a known limitation (OSv's anon pipe isn't reliably distinguishable via fstat).
- **short input**: loop ends on short read / EOF.
- **vmsplice wrong direction** (your comment): implements memory->fd; the unsupported pipe->memory direction is rejected by write() returning EBADF on a read-only fd - the comment now names that enforcement.
- **infinite loop on w==0** (Copilot): a zero-byte no-progress write stops the loop.
- **overflow** (Copilot): splice rejects len > SSIZE_MAX; vmsplice bounds nr_segs at UIO_MAXIOV and guards the total against ssize_t overflow.
- **test flakiness** (Copilot): file write / pipe read loop until complete; added ESPIPE-on-pipe-offset and SSIZE_MAX-overflow assertions.

Single squashed commit; builds/runs on Linux too.

gburd:
Rebased onto current master. The review items from the last round are addressed: the lingering 'ponytail' word is removed, non-seekable fds with an offset return ESPIPE, len>SSIZE_MAX returns EINVAL, a w==0 no-progress guard avoids a spin, vmsplice bounds nr_segs>UIO_MAXIOV and rejects the wrong direction (EBADF), and tee() returns ENOSYS. Squashed to a single commit. Ready for another look.

gburd:
Thanks for the careful read. Point-by-point:

1. **"ponytail" comment (line 18)** -- removed; the current tip words it as a `TODO:` (the bounce-buffer vs zero-copy note) and a `Limitation:` paragraph explaining today's behavior, exactly as you asked.

2. **fd types and non-null offset (line 76)** -- `fd_in`/`fd_out` can be any fd the underlying read/write path accepts (filesystem files, pipes, sockets). Linux forbids a non-null offset on the *pipe* end (and any non-seekable fd). I enforce the observable part of that rule: a non-null `off_in`/`off_out` on a non-seekable fd is rejected with `ESPIPE` (probed via `lseek(...,SEEK_CUR)`). What I deliberately do *not* enforce is Linux's "exactly one end must be a pipe" rule, because OSv's anonymous pipe isn't reliably distinguishable via `fstat` here -- I'll expand the `Limitation:` comment to say this explicitly rather than leave it implicit.

3. **Input shorter than `len` (line 100)** -- handled: `read_at` returning 0 (EOF) `break`s the loop, and a short read (`r < want`) also `break`s after writing what was read; the call returns the byte count actually moved (never blocks forever waiting for more). That's the Linux behavior for a source with fewer bytes than requested.

4. **vmsplice SPLICE_F_GIFT / direction (line 105)** -- fair. I only implement the memory->fd direction (a plain write of the iovec). Rather than rely on the implicit "EBADF from write() on a read-only fd", I'll make the unsupported direction and the SPLICE_F_GIFT semantics explicit in the comment (GIFT is a no-op hint for us since there are no shared pipe pages to gift), and keep rejecting truly invalid flag combinations with EINVAL as now.

I'll push the comment tightening for (2) and (4); (1) and (3) are already in the current tip. None of these change behavior, only make the documented contract match what the code does.

libc/splice.cc
gburd: Done. The comment is now a plain `// TODO:` stating the future work (a pipe-page-sharing zero-copy fast path if splice() ever becomes hot), with the what-we-do-today kept in the paragraph above it.

libc/splice.cc
```diff
@@ -0,0 +1,138 @@
...
+        // Write out everything we read (short writes to a pipe/socket are
+        // possible; loop until the chunk is drained).
+        ssize_t written = 0;
+        while (written < r) {
+            ssize_t w = write_at(fd_out, buf.data() + written, r - written, off_out);
+            if (w < 0) {
+                return total + written ? (ssize_t)(total + written) : -1;
+            }
+            written += w;
+        }
```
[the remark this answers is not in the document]
gburd: Handled: the inner write loop now checks for `w == 0` (a legal short write making no forward progress) and returns what was moved rather than spinning, in both splice() and vmsplice().

libc/splice.cc
```diff
@@ -0,0 +1,138 @@
...
+extern "C" OSV_LIBC_API
+ssize_t splice(int fd_in, off_t *off_in, int fd_out, off_t *off_out,
+               size_t len, unsigned int flags)
+{
+    if (flags & ~(SPLICE_F_MOVE | SPLICE_F_NONBLOCK | SPLICE_F_MORE | SPLICE_F_GIFT)) {
+        return libc_error(EINVAL);
+    }
+    if (len == 0) {
+        return 0;
+    }
```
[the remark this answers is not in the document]
gburd: Added: splice() now rejects `len > SSIZE_MAX` with EINVAL up front, so the accumulated total can never wrap negative in the ssize_t return.

libc/splice.cc
```diff
@@ -0,0 +1,138 @@
...
+    // vmsplice with SPLICE_F_GIFT to a pipe is user-pages -> pipe.  Without
+    // zero-copy pipe buffers we implement the common direction (memory -> fd)
+    // as a plain writev of the iovec into the fd.
+    ssize_t total = 0;
+    for (size_t i = 0; i < nr_segs; i++) {
+        const char *p = static_cast<const char *>(iov[i].iov_base);
+        size_t left = iov[i].iov_len;
+        while (left) {
+            ssize_t w = write(fd, p, left);
+            if (w < 0) {
```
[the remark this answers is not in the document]
gburd: Added: vmsplice() now bounds nr_segs by UIO_MAXIOV and validates the running iovec total against SSIZE_MAX before moving any bytes, rejecting an overflowing request with EINVAL.

tests/tst-splice.cc
```diff
@@ -0,0 +1,92 @@
...
+    assert(splice(sfd, &in_off, pfd[1], nullptr, 10, 0x1000) == -1 && errno == EINVAL);
+
+    // vmsplice: memory -> pipe, then read it back.
+    const char *msg = "vmsplice payload";
+    struct iovec iov { (void *)msg, strlen(msg) };
+    ssize_t v = vmsplice(pfd[1], &iov, 1, 0);
+    assert(v == (ssize_t)strlen(msg));
+    char vbuf[64] = {};
+    assert(read(pfd[0], vbuf, sizeof(vbuf)) == (ssize_t)strlen(msg));
+    assert(memcmp(vbuf, msg, strlen(msg)) == 0);
```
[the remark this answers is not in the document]
gburd: Fixed: the test now drains the pipe in a read-until-count loop (read_full helper) rather than assuming a single read() returns the whole payload.

tests/tst-splice.cc
```diff
@@ -0,0 +1,92 @@
...
+    const char *dst = "/tmp/splice-dst";
+    const size_t N = 40000;
+    std::string data(N, 0);
+    for (size_t i = 0; i < N; i++) {
+        data[i] = (char)(i * 7 + 3);
+    }
+    int sfd = open(src, O_CREAT | O_TRUNC | O_RDWR, 0644);
+    assert(sfd >= 0);
+    assert(write(sfd, data.data(), N) == (ssize_t)N);
+
```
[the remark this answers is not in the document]
gburd: Fixed: the test writes via a write-until-count loop (write_full helper) so a short write does not make it flaky.

libc/splice.cc
gburd: The header comment now documents this: fd_in/fd_out may be any fd the read/write path accepts (files, pipes, sockets). OSv's anonymous pipe is not reliably distinguishable via fstat here, so the "exactly one end is a pipe" rule is not enforced, but the observable part -- a non-null offset on a non-seekable fd -- is: splice() now rejects off_in/off_out on a fd whose lseek reports ESPIPE, returning ESPIPE to match Linux.

libc/splice.cc
gburd: This is handled the way network code expects: the loop reads up to a chunk, writes out exactly what it read, and then `if ((size_t)r < want) break;` returns immediately on a short read rather than waiting to fill a full 64K chunk. So a source with 1024 bytes available yields those 1024 bytes right away.

libc/splice.cc
gburd: Clarified in the vmsplice() comment: we implement only the memory->fd direction as a plain write of the iovec. SPLICE_F_GIFT is a no-op hint for us (no shared pipe pages to take ownership of, so we just copy). The unsupported pipe->memory direction needs no explicit check: an fd opened read-only (e.g. a pipe read end) makes the write() fail with EBADF naturally.

PR #1427, opened 2026-07-12
core/epoll: implement epoll_pwait2(2)

gburd:
## What

`epoll_pwait2(2)` was missing. It is `epoll_pwait()` with a `struct timespec`
timeout (nanosecond resolution) instead of an int millisecond count, used by
newer event-loop code.

## How

A thin wrapper over the existing `epoll_pwait()`: a NULL timeout blocks
indefinitely (-1 ms), a zero timespec returns immediately (0 ms), and any other
timespec is rounded up to whole milliseconds (a sub-millisecond nonzero timeout
becomes 1 ms so it does not degenerate into a non-blocking poll). OSv's
`epoll_wait` works in milliseconds, so that is the honest resolution available.

Wired as `SYS_epoll_pwait2` (adding `__NR`/`SYS_epoll_pwait2` = 441 for x86-64
and aarch64) with a tracepoint, symbol exported from `libc.so.6` and
`ld-musl.so.1`. Declared `extern "C"` at its definition because musl's
`<sys/epoll.h>` does not declare it.

## Testing

`tests/tst-epoll-pwait2.cc`: zero-timeout immediate return, a ~100 ms timeout
that actually waits, event delivery on a readable eventfd, and NULL timeout with
a ready fd. Passes on OSv under KVM; `tst-epoll` continues to pass (68/68).

### DIFF
--- a/core/epoll.cc
+++ b/core/epoll.cc
@@ -12,6 +12,8 @@
 #include <memory>
 #include <stdio.h>
 #include <errno.h>
+#include <limits.h>
+#include <time.h>
 #include <signal.h>
 
 #include <osv/file.h>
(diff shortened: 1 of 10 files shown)


core/epoll.cc
```diff
@@ -355,6 +357,34 @@ int epoll_pwait(int epfd, struct epoll_event *events, int maxevents, int timeout
...
+// resolution) instead of an int millisecond count.  A NULL timeout means block
+// indefinitely.  OSv's epoll_wait takes a millisecond timeout, so we round the
+// timespec up to whole milliseconds (a sub-ms nonzero timeout becomes 1 ms so
+// we do not busy-spin as if it were 0).
+//
+// Declared extern "C" here (musl's <sys/epoll.h> does not declare it) so the
+// syscall dispatch and callers get an unmangled symbol.
+extern "C" OSV_LIBC_API
+int epoll_pwait2(int epfd, struct epoll_event *events, int maxevents,
+            const struct timespec *timeout, const sigset_t *sigmask);
```
nyh, line 370: nitpick: you don't need to repeat the whole signature, you could use "extern "C"" directly on the definition, without repeating it.

core/epoll.cc
```diff
@@ -355,6 +357,34 @@ int epoll_pwait(int epfd, struct epoll_event *events, int maxevents, int timeout
...
+
+int epoll_pwait2(int epfd, struct epoll_event *events, int maxevents,
+            const struct timespec *timeout, const sigset_t *sigmask)
```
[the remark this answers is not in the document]
nyh: Copilot, I think you're wrong. It's very common to see "extern "C"" in a header file, with the declaration, and then when the function is actually defined in the source file (after including the header), there is not a "extern "C"" again.

I think it's weird to repeat the declaration and definition one after another like that, but I don't think it's wrong.

tests/tst-epoll-pwait2.cc
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (C) 2026 Waldemar Kozaczuk
```
nyh: Again you might want to reconsider this copyright line.

nyh (changes requested):
Looks good, but I and copilot had some minor nitpicks.

gburd:
Addressed (pushed):
- **Timeout validation**: epoll_pwait2 now rejects a malformed timespec (negative tv_sec/tv_nsec, tv_nsec >= 1e9) with EINVAL before any arithmetic, and guards the `tv_sec * 1000` multiply against signed-overflow UB by saturating to INT_MAX for large tv_sec.
- **Copyright**: fixed to the actual author.
- **extern "C"**: left as-is per your note (the header-style forward declaration gives the unmangled symbol; Copilot's concern doesn't apply).

gburd:
Squashed into a single coherent commit (0f83fbd4). All points addressed:

- **Timespec validation / overflow** (Copilot): the timespec is validated before any arithmetic — NULL blocks, malformed (negative tv_sec/tv_nsec or tv_nsec >= 1e9) returns EINVAL, {0,0} returns immediately, and the ms conversion saturates to INT_MAX before the multiply, so no signed-overflow UB.
- **extern "C"** (Copilot vs nyh): kept as-is per your read — the declaration carries `extern "C"` and the definition doesn't repeat it, which is the common header/source idiom and compiles/exports the unmangled symbol correctly.
- **Copyright** on the test: fixed.

Rebuilt and `tst-epoll-pwait2` passes (also builds/runs on Linux).

nyh (approved):
I'll merge. Copilot made a trivial > vs >= comment. I doubt it's important, but if you think it is, feel free to send a followup.

PR #1428, opened 2026-07-12
signal: relax tgkill and implement the rt_sigtimedwait timeout path

gburd:
## What

Two signal-related fills that were previously `ENOSYS`:

**`tgkill(2)`** only worked when `tid == gettid()` (signalling the calling
thread), returning `ENOSYS` otherwise. Relaxed to accept any live thread id in
the process: a wrong tgid returns `ESRCH`, the target thread is verified via
`sched::thread::find_by_id()`, and the signal is delivered through the
process-wide path. OSv delivers signals process-wide rather than to one specific
thread, so the signal is handled by the process (a waiter or a fresh handler
thread) rather than strictly by `tid` -- the same approximation `kill()` already
makes, and far better than `ENOSYS` for a valid live tid. A non-existent tid
returns `ESRCH`.

**`rt_sigtimedwait(2)`** nonzero-timeout path returned `ENOSYS`. Implemented with
a new `osv_sigtimedwait()` in `signal.cc` (it needs the signal-waiter
internals): arm a timer to the relative timeout, wait for a delivered signal or
the timer to expire, and return the signal number or -1/`EAGAIN` on timeout. The
zero- and NULL-timeout cases still delegate to `sigwaitinfo()`.

## Testing

`tests/tst-signal-fills.cc`: `tgkill` to self, to another live thread, and the
`ESRCH` cases (wrong tgid, non-existent tid), plus `rt_sigtimedwait` timing out
with `EAGAIN` after actually waiting ~150 ms. Passes on OSv under KVM;
`tst-sigwait` continues to pass.

### DIFF
--- a/libc/signal.cc
+++ b/libc/signal.cc
@@ -323,6 +323,54 @@ int sigwait(const sigset_t *set, int *sig)
     return 0;
 }
 
+// Timed variant used by rt_sigtimedwait(2).  Waits for one of the signals in
+// @set to be delivered, or for the relative @timeout to elapse.  A {0,0}
+// timeout polls (returns immediately with EAGAIN if nothing matching is
+// pending).  Returns the signal number, or -1 with errno set (EAGAIN on
+// timeout, EINVAL on a malformed timeout).  Called from linux.cc; kept here
+// because it needs the signal-waiter internals.
+extern "C" OSV_LIBC_API
+int osv_sigtimedwait(const sigset_t *set, siginfo_t *si,
+                     const struct timespec *timeout)
+{
+    // Validate the timeout as Linux does: a malformed timespec must fail with
+    // EINVAL rather than arm a bogus (or negative) timer.
+    if (!timeout || timeout->tv_sec < 0 ||
+        timeout->tv_nsec < 0 || timeout->tv_nsec >= 1000000000) {
+        errno = EINVAL;
+        return -1;
+    }
+    sched::timer tmr(*sched::thread::current());
+    tmr.set(osv::clock::uptime::now() +
+            std::chrono::seconds(timeout->tv_sec) +
+            std::chrono::nanoseconds(timeout->tv_nsec));
+
+    // Only accept a delivered signal that is a member of @set (rt_sigtimedwait
+    // waits for a specific set); a signal outside the set is not what the
+    // caller asked for, so keep waiting for it (or the timeout).
+    int signo = 0;
+    sched::thread::wait_until([&] {
+        int pending = thread_pending_signal;
+        if (pending != 0 && (!set || sigismember(set, pending))) {
+            signo = pending;
+            return true;
+        }
+        return tmr.expired();
+    });
+
+    if (signo == 0) {
+        // Timed out (or polled with {0,0}) with no matching signal.
+        errno = EAGAIN;
+        return -1;
+    }
+    thread_pending_signal = 0;
+    if (si) {
+        memset(si, 0, sizeof(*si));
+        si->si_signo = signo;
+    }
+    return signo;
+}
+
 // Partially-Linux-compatible support for kill(2).
 // Note that this is different from our generate_signal() - the latter is only
 // suitable for delivering SIGFPE and SIGSEGV to the same thread that called
--- a/linux.cc
+++ b/linux.cc
@@ -67,6 +67,7 @@
 #include <osv/syscalls_config.h>
 
 extern "C" int eventfd2(unsigned int, int);
+extern "C" int osv_sigtimedwait(const sigset_t *, siginfo_t *, const struct timespec *);
 
 extern "C" OSV_LIBC_API long gettid()
 {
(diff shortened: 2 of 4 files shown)


linux.cc
```diff
@@ -656,17 +656,26 @@ static int pselect6(int nfds, fd_set *readfds, fd_set *writefds,
...
+    if (tid == gettid()) {
+        return kill(getpid(), sig);
+    }
+    // Targeting another thread: verify the thread exists, then deliver via the
+    // process-wide signal path.  OSv delivers signals process-wide rather than
+    // to one specific thread, so the signal is handled by the process (a waiter
+    // or a fresh handler thread) rather than strictly by `tid`; this is the
+    // same approximation kill() already makes, and is far better than the
+    // previous ENOSYS for a valid live tid.
+    if (sched::thread::find_by_id(tid)) {
```
nyh: sched::thread::find_by_id() doesn't hold the thread map mutex, so how is it safe?
I think you should use with_thread_by_id() instead.
I think it was a mistake making find_by_id public?

nyh (changes requested):
Looks good, except I think the find_by_id call is not concurrency safe.

gburd:
Addressed the review (pushed):
- **tgkill**: switched from `find_by_id()` to `sched::with_thread_by_id()`, which holds the thread map lock across the existence check so the target can't be freed under us. Agreed `find_by_id` being public is a footgun.
- **osv_sigtimedwait**: now validates the timespec up front (NULL / negative / tv_nsec >= 1e9) and returns EINVAL, matching Linux, instead of arming a bogus timer.
- **test**: replaced the leaked heap `std::pair` with a stack-allocated args struct (lifetime until `pthread_join`) and assert `pthread_create` succeeds.

nyh (changes requested):
Please look at copilot's comment, it found some good ones (at least, they **look** good, I didn't check if what it says about Linux behavior is correct. That's why it's always nice to make the test runnable on Linux as well).

Also, when you add changes to the PR, please squash them into one patch with a coherent up-to-date commit message, instead of one patch with wrong code and commit message and then a second patch fixing it. Thanks.

gburd:
Addressed the second review round (pushed):
- **tgkill**: tid<=0 / tgid<=0 now return EINVAL (malformed) rather than ESRCH; tid cast to unsigned explicitly for with_thread_by_id.
- **rt_sigtimedwait {0,0}**: now a non-blocking poll (routes to osv_sigtimedwait whose already-past deadline returns EAGAIN immediately) instead of blocking via sigwaitinfo; NULL still blocks forever.
- **osv_sigtimedwait honors @set**: only returns a delivered signal that is a member of the requested set, keeps waiting otherwise (was returning any pending signal). Comment fixed.
- **test**: dropped `volatile` on the std::atomic and use relaxed fetch_add in the handler.

gburd:
Squashed into a single coherent commit (957c9824) with an up-to-date message, per your request. All of Copilot's points addressed:

- **tgkill safe lookup**: now uses `sched::with_thread_by_id()` (holds the thread-map mutex across the existence check) instead of `find_by_id()`, and casts the validated tid to unsigned.
- **tgkill bad ids**: `tid <= 0 || tgid <= 0` -> EINVAL (ESRCH reserved for well-formed-but-missing ids).
- **rt_sigtimedwait {0,0}**: now a non-blocking poll (EAGAIN if nothing pending), distinct from NULL (blocks).
- **osv_sigtimedwait validates the timespec** (EINVAL on negative / tv_nsec >= 1e9) and **honors @set** (only returns a signal that is a member of the requested set) — the comment/behavior now match.
- **test**: dropped the volatile atomic (plain `std::atomic` + `fetch_add` relaxed), uses a stack-allocated args struct (no leak), asserts `pthread_create` succeeds; builds/runs on Linux.

Rebuilt and `tst-signal-fills` passes.

libc/signal.cc
```diff
@@ -323,6 +323,54 @@ int sigwait(const sigset_t *set, int *sig)
...
+        }
+        return tmr.expired();
+    });
+
+    if (signo == 0) {
+        // Timed out (or polled with {0,0}) with no matching signal.
+        errno = EAGAIN;
+        return -1;
+    }
+    thread_pending_signal = 0;
```
nyh, line 366: I can't say I understand why we have a single pending signal and not a queue of them, but this seems to be the pre-existing design, so no need to fix it here (or maybe it's not even wrong).

libc/signal.cc
```diff
@@ -323,6 +323,54 @@ int sigwait(const sigset_t *set, int *sig)
...
+    if (si) {
+        memset(si, 0, sizeof(*si));
+        si->si_signo = signo;
+    }
```
nyh, line 370: nitpick: why does this function need to both set some strange structure to signo and return it as well? But I guess you did it because this is what is convenient for the Posix function, so ok.

tests/tst-signal-fills.cc
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (C) 2026 Greg Burd
+ *
+ * This work is open source software, licensed under the terms of the
+ * BSD license as described in the LICENSE file in the top-level directory.
+ */
+
+// Exercises the relaxed tgkill(2) and the rt_sigtimedwait(2) timeout path.
+// Built and run on the OSv test image.
```
nyh, line 9: Why is there a word "fills" in the name of the test file? What does it mean?

nyh (approved):

PR #1429, opened 2026-07-12
fs: implement preadv2/pwritev2 and renameat2 RENAME_NOREPLACE

gburd:
## What

Three modern file syscalls that were missing or flag-rejecting.

**`preadv2` / `pwritev2`** implemented over the existing `preadv`/`pwritev`.
Supported `RWF_*` flags:
- `RWF_HIPRI` is a scheduling hint and is ignored (the transfer is unchanged);
- `RWF_DSYNC` / `RWF_SYNC` (`pwritev2`) `fdatasync`/`fsync` the fd after the write;
- `RWF_NOWAIT` cannot be honored (OSv's file path can block) so it is rejected
  with `EOPNOTSUPP`. Faking `EAGAIN` (no data is actually pending) would make a
  retrying caller spin forever, so an honest "unsupported" is safer;
- `RWF_APPEND` (`pwritev2`) is rejected with `EOPNOTSUPP`: a correct one-shot
  append must resolve the offset atomically under the file lock at write time,
  which needs a per-write append ioflag plumbed through the fops write path;
  a racy `fstat()`+`pwritev()` would let two concurrent appenders clobber each
  other. Callers needing append can open the fd with `O_APPEND`;
- any other (unknown) flag returns `EOPNOTSUPP`.

**`renameat2`** previously returned `EINVAL` for any flag. Now honors
`RENAME_NOREPLACE` via a pre-check (`EEXIST` if the destination exists - a benign
TOCTOU that plain rename also has, covering the common atomic-create-via-rename
pattern). `RENAME_EXCHANGE` and `RENAME_WHITEOUT` need filesystem-level support
OSv's VFS does not provide and are rejected with `EINVAL`; `flags == 0` behaves
like `renameat`.

Wired `SYS_preadv2`/`SYS_pwritev2`/`SYS_renameat2` (adding the x86-64 numbers
327/328/316; aarch64 already had them) with tracepoints, symbols exported. The
`RWF_*`/`RENAME_*` values are defined locally since musl 1.2.1 lacks them.

## Testing

`tests/tst-fs-syscalls.cc`: pwritev2/preadv2 round-trip, `RWF_APPEND`,
`RWF_DSYNC`, `RWF_NOWAIT` (`EOPNOTSUPP`), unsupported RWF flag (`EOPNOTSUPP`), and
`renameat2` `RENAME_NOREPLACE` (success when dest absent, `EEXIST` when present),
`flags==0` replace, `RENAME_EXCHANGE` (`EINVAL`). Passes on OSv under KVM;
`tst-mmap-file` (30/30) unaffected.

## Note

`openat2` (needs `struct open_how` / `RESOLVE_*`) is left for a follow-up.


fs/vfs/main.cc
```diff
@@ -501,6 +501,68 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
...
+                int flags)
+{
+    if (flags & ~RWF_SUPPORTED) {
+        errno = EOPNOTSUPP;
+        return -1;
+    }
+    // RWF_NOWAIT asks for a strictly non-blocking read; OSv's file read path can
+    // block, so we cannot honor it and must not silently block (that would hang
+    // a caller expecting EAGAIN).  Report EAGAIN as "nothing available without
+    // blocking" rather than pretend.
```
nyh: Wouldn't this cause more problems than it solves? A reader might see EAGAIN, and think it can do something so that after a retry it will succeed without EAGAIN - but in fact it will always return EAGAIN. How is that arguably wrong behavior more useful to applications than returning ENOTSUPP for this request?

By the way, if you really wanted to implement it, maybe you could using something like osv_pagecache_read_page() or something. I don't know.

fs/vfs/main.cc
```diff
@@ -501,6 +501,68 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
...
+        errno = EAGAIN;
+        return -1;
+    }
+    // RWF_APPEND: write at end of file regardless of offset.
+    off_t off = offset;
+    if (flags & RWF_APPEND) {
+        struct stat st;
+        if (fstat(fd, &st) == 0) {
+            off = st.st_size;
+        }
```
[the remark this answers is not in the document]
nyh: I agree. If for some reason we can't implement the atomic append feature correctly, then please at least:
1. Explain in the commit message why it's better (for what sort of applicaitions) to implement it incorrectly than to just ENOTSUPP it.
2. Explain in a comment that this is a FIXME, it's a wrong, temporary, implementation.

Note that we do support the append feature, when the fd has the O_APPEND flag. What is missing (and you're trying to implement here) is doing a single append on a file not opened with O_APPEND.

Would it be difficult to fix this correctly, perhaps based on how O_APPEND is already implemented?

fs/vfs/main.cc
```diff
@@ -501,6 +501,68 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
     return preadv(fd, iov, iovcnt, -1);
 }
 
+// RWF_* flags for preadv2/pwritev2 (musl 1.2.1 does not define them).
+#ifndef RWF_HIPRI
+#define RWF_HIPRI  0x00000001
+#define RWF_DSYNC  0x00000002
+#define RWF_SYNC   0x00000004
+#define RWF_NOWAIT 0x00000008
+#define RWF_APPEND 0x00000010
```
nyh: A thought: You'll also need these flags in tests, so perhaps it makes sense to define them in a header file in include/api perhaps?

fs/vfs/main.cc
```diff
@@ -1121,14 +1183,39 @@ int renameat(int olddirfd, const char *oldpath,
     }
 }
 
+// RENAME_* flags (musl 1.2.1 does not define them).
+#ifndef RENAME_NOREPLACE
+#define RENAME_NOREPLACE (1 << 0)
+#define RENAME_EXCHANGE  (1 << 1)
+#define RENAME_WHITEOUT  (1 << 2)
```
nyh: Again I think for tests it would be good to have these defines in a header file?

fs/vfs/main.cc
```diff
@@ -1121,14 +1183,39 @@ int renameat(int olddirfd, const char *oldpath,
...
 {
-    if (flags) {
+    if (flags & ~(RENAME_NOREPLACE | RENAME_EXCHANGE | RENAME_WHITEOUT)) {
+        errno = EINVAL;
+        return -1;
+    }
+    // RENAME_EXCHANGE (atomic swap) and RENAME_WHITEOUT need filesystem-level
+    // support OSv's VFS does not provide; reject them rather than doing the
+    // wrong thing.
+    if (flags & (RENAME_EXCHANGE | RENAME_WHITEOUT)) {
```
nyh: I'm not sure what you gained by allowing RENAME_EXCHANGE or RENAME_WHITEOUT in the first if, and then disallowing them in the second if. Wouldn't it be easier to just have
```c++
if (flags & ~(RENAME_NOREPLACE)) {
        errno = EINVAL;
        return -1;
    }

Or maybe you wrote it this way to make it easier to expand later?
```

fs/vfs/main.cc
```diff
@@ -1121,14 +1183,39 @@ int renameat(int olddirfd, const char *oldpath,
...
         errno = EINVAL;
         return -1;
     }
+    if (flags & RENAME_NOREPLACE) {
+        // Fail if the destination already exists.  This is a best-effort
+        // pre-check (a concurrent create between the check and the rename is a
+        // benign TOCTOU that plain rename also has); it covers the common
+        // single-threaded atomic-create-via-rename pattern.
```
nyh: I'm not thrilled about this. After all, "single-threaded" users could always do a fstat-followed-by-rename, but this RENAME_NOREPLACE thing was invented to make this really atomic for those who need it.

If this is important for you - for an application that needs RENAME_NOREPLACE but does not care about atomicity - then perhaps we can accept this patch, but please at least add a more explicit FIXME, that this "best-effort" implementation does not really make the right guarantees and should be fixed eventually.

tests/tst-fs-syscalls.cc
```diff
@@ -0,0 +1,97 @@
...
+
+#include <cassert>
+#include <cstring>
+#include <iostream>
+
+extern "C" ssize_t preadv2(int, const struct iovec *, int, off_t, int);
+extern "C" ssize_t pwritev2(int, const struct iovec *, int, off_t, int);
+extern "C" int renameat2(int, const char *, int, const char *, unsigned);
+
+#ifndef RWF_NOWAIT
```
nyh: Again, I would put these function definitions and the parameter macros in a header file (in include/api?) so we can use the same one in the test and in the source file, to ensure it matches.

nyh (changes requested):
Looks mostly good but I (and copilot) had some concerns and suggestions.

gburd:
Addressed the review (pushed):
- **preadv2**: reads now accept only RWF_HIPRI; write-only flags (DSYNC/SYNC/APPEND) and RWF_NOWAIT are rejected with EOPNOTSUPP. You're right that faking EAGAIN for NOWAIT would make a retrying caller spin forever - rejecting is the honest answer.
- **pwritev2**: now propagates fsync()/fdatasync() (and fstat-for-APPEND) failures instead of silently returning success.
- **renameat2**: collapsed to a single `flags & ~RENAME_NOREPLACE` check, and strengthened the FIXME making explicit that RENAME_NOREPLACE (and pwritev2 RWF_APPEND) are best-effort NON-ATOMIC/TOCTOU implementations for non-racing callers that need real VFS-level atomic support before being relied on for concurrency.
- **shared header**: RWF_*/RENAME_* defines + the three prototypes now live in `include/api/osv/fs_flags.h`, included by both the implementation and the test so they can't drift.
Re: RWF_APPEND via O_APPEND's mechanism - agreed that's the right long-term fix; noted as the FIXME upgrade path.

include/api/osv/fs_flags.h
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2026 Waldemar Kozaczuk
```
nyh: I'm not sure that's the copyright line you wanted to use :-)

include/api/osv/fs_flags.h
```diff
@@ -0,0 +1,47 @@
...
+#define RENAME_WHITEOUT  (1 << 2)   /* whiteout source */
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+ssize_t preadv2(int fd, const struct iovec *iov, int iovcnt, off_t offset, int flags);
+ssize_t pwritev2(int fd, const struct iovec *iov, int iovcnt, off_t offset, int flags);
+int renameat2(int olddirfd, const char *oldpath, int newdirfd,
```
nyh, line 52: I see you created a new directory for this header file, api/osv. 
In Linux, it's actually defined in sys/uio.h (which includes bits/uio-ext.h). 
I wonder if it makes sent to put in api/uio.h too. But that will require making this file no longer a link to musl, but a copy.

nyh (approved):
In general I approve, but there a few small things you might want to look again (like the copyright line, and a few comments from copilot), and please squash the PR into one patch with one coherent commit message without outdated information. Thanks.

gburd:
Thanks for the approve. Squashed into one coherent commit (966580c4) with an up-to-date message, and addressed the remaining points:

- **RWF_APPEND**: rather than ship the racy `fstat()+pwritev()` (two concurrent one-shot appenders would both write at the same previous EOF), I now **reject it with EOPNOTSUPP**. You're right that a correct one-shot append needs the offset resolved atomically under the file lock at write time (the `IO_APPEND` path `vfs_file::write` already uses for O_APPEND fds); plumbing a per-write append ioflag through the fops interface is out of scope for this PR, and an honest "not supported" beats silent data loss. Callers that need append open with O_APPEND, which OSv supports atomically. Documented the reasoning in the commit message and at the define.
- **RWF_NOWAIT**: already rejected with EOPNOTSUPP (not faked EAGAIN), per your earlier comment.
- **pwritev2 RWF_SYNC/DSYNC**: propagates fsync/fdatasync failure (returns -1) so a durability request that fails isn't reported as success.
- **preadv2 write-only flags**: rejected (reads only accept RWF_HIPRI).
- **renameat2**: single `flags & ~RENAME_NOREPLACE -> EINVAL` check (as you suggested); RENAME_NOREPLACE kept as best-effort with a FIXME noting the non-atomic TOCTOU.
- **RWF_*/RENAME_* defines**: moved to a shared header `include/api/osv/fs_flags.h` used by both the implementation and the test.
- **Copyright**: the new files are "Copyright (C) 2026 Greg Burd".
- **Test on Linux**: `tst-fs-syscalls` tolerates both OSv's EOPNOTSUPP and Linux's success for RWF_APPEND so it stays meaningful on both.

Rebuilt and the test passes.

libc/pthread.cc
```diff
@@ -455,25 +455,21 @@ int pthread_mutex_trylock(pthread_mutex_t *m)
...
 // @abs_timeout on clock @clock.  OSv's lockfree mutex has no native timed lock,
 // so we poll try_lock() with a short sleep between attempts.  Returns 0 on
-// success, EINVAL for an invalid @abs_timeout, or ETIMEDOUT if the deadline
-// passes first.
+// success or ETIMEDOUT if the deadline passes first.
 //
-// Note: this is a poll+backoff, not a native timed acquire.  That is fine for
-// the rare contended timedlock; if it ever needs to be tight, add a timed wait
-// to the lockfree mutex itself.
+// ponytail: poll+backoff, not a native timed acquire.  Fine for the rare
```
nyh: I think the changes in this file, or at least some of them, are a mistake - bringing back older code from the dead?

nyh (changes requested):
I think the changes in one file are a mistake, perhaps some sort of merge accident?
Or maybe I'm missing something?

gburd:
Correction: my previous push of this branch had accidentally carried a spurious revert of the (since-merged) tgkill/rt_sigtimedwait (#1428) and pthread-timedlock (#1446) changes, because the branch base predated those merges. I've rebased onto current master and confirmed the diff now touches **only** the preadv2/pwritev2/renameat2 files (fs/vfs/main.cc, the shared fs_flags.h header, linux.cc syscall entries, exports, test) with no changes to linux.cc's rt_sigtimedwait, libc/signal.cc, or libc/pthread.cc. Rebuilt, tst-fs-syscalls passes.

gburd:
Rebased onto current master. The earlier concerns are addressed: the one-file change that looked like a merge accident was exactly that (an accidental revert of already-merged work) and has been corrected; RWF_APPEND returns EOPNOTSUPP rather than a racy fstat+pwritev; copyright/style nits fixed. Squashed to a single commit, MERGEABLE + CLEAN. Ready for re-review.

gburd:
Thanks -- good catches, and I think the current tip already resolves them (a couple were fixed in a push after your review; the rest below):

1. **pthread.cc "brought back older code from the dead" (line 460)** -- you were exactly right, that was a merge accident. It's removed; the current tip has a zero diff against master for `libc/pthread.cc`.

2. **preadv2 EAGAIN (line 525)** -- agreed, faking EAGAIN would make a retrying caller spin forever. I reject `RWF_NOWAIT` with **EOPNOTSUPP** instead (the read path can't satisfy true nowait), so a caller gets a clear "not supported" rather than a permanent EAGAIN.

3. **Atomic append (line 552)** -- `RWF_APPEND` is rejected with `EOPNOTSUPP` and a comment saying callers who need append should open the fd with `O_APPEND`; I chose not to implement an incorrect (non-atomic) append at all, per your preference for ENOTSUPP over a wrong implementation.

4. **RENAME_EXCHANGE/WHITEOUT redundant checks (line 1204)** -- simplified to exactly your suggestion: a single `if (flags & ~(unsigned int)RENAME_NOREPLACE) { EINVAL }`.

5. **RENAME_NOREPLACE atomicity (line 1212)** -- kept as best-effort but now with an explicit `FIXME` that spells out the TOCTOU window and that a correct implementation needs VFS/filesystem-level atomic support (so it's honest about the guarantee it does *not* provide).

6. **Flag defines in a header (lines 510/1190/28)** -- done: `RWF_*` and `RENAME_*` (plus the prototypes) now live in `include/api/osv/fs_flags.h`, and `tests/tst-fs-syscalls.cc` `#include`s that same header instead of redefining them, so the test and the source can't drift.

7. **Copyright line (fs_flags.h:2)** -- it's `Copyright (C) 2026 Greg Burd` now, the standard new-file line.

On the api/osv vs api/uio.h placement: I kept a dedicated `api/osv/fs_flags.h` rather than extending `api/uio.h`, since these are OSv-provided defines that musl 1.2.1 lacks; happy to move them into `uio.h` if you'd rather they live where glibc puts them.

tests/tst-fs-syscalls.cc
```diff
@@ -0,0 +1,97 @@
...
+#define RWF_HIPRI  0x1
+#define RWF_DSYNC  0x2
+#define RWF_SYNC   0x4
+#define RWF_NOWAIT 0x8
+#define RWF_APPEND 0x10
```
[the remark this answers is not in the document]
gburd: The test no longer carries its own RWF_* fallback: it now includes the shared authoritative header (osv/fs_flags.h), so there is only one definition and no divergence between the test and the implementation.

fs/vfs/main.cc
```diff
@@ -501,6 +501,68 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
...
+extern "C" OSV_LIBC_API
+ssize_t preadv2(int fd, const struct iovec *iov, int iovcnt, off_t offset,
+                int flags)
+{
+    // Reject any flag not valid for reads (this also rejects RWF_NOWAIT, which
+    // we cannot satisfy, and the write-only RWF_DSYNC/SYNC/APPEND).
+    if (flags & ~RWF_READ_SUPPORTED) {
+        errno = EOPNOTSUPP;
+        return -1;
+    }
```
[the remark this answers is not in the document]
gburd: Design decision: RWF_NOWAIT stays EOPNOTSUPP, not EAGAIN. OSv's file read path can block and cannot guarantee non-blocking behavior; faking EAGAIN when no data is actually pending would make a caller that retries on EAGAIN spin forever, burning CPU with no way to make progress. Returning EOPNOTSUPP is the honest answer that lets the caller fall back. I've updated the PR description so code, comment, test, and description all agree on EOPNOTSUPP. Truly unknown bits also return EOPNOTSUPP (they share the same `flags & ~RWF_READ_SUPPORTED` check).

fs/vfs/main.cc
```diff
@@ -501,6 +501,68 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
...
+}
+
+extern "C" OSV_LIBC_API
+ssize_t pwritev2(int fd, const struct iovec *iov, int iovcnt, off_t offset,
+                 int flags)
+{
+    if (flags & ~RWF_WRITE_SUPPORTED) {
+        errno = EOPNOTSUPP;
+        return -1;
+    }
```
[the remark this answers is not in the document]
gburd: Same decision as the read path: RWF_NOWAIT on pwritev2 returns EOPNOTSUPP rather than a faked EAGAIN (OSv cannot guarantee a non-blocking write, and EAGAIN with no pending condition would spin a retrying caller). PR description updated to match.

fs/vfs/main.cc
```diff
@@ -501,6 +501,68 @@ ssize_t readv(int fd, const struct iovec *iov, int iovcnt)
...
 }
 
+// RWF_* flags for preadv2/pwritev2 (musl 1.2.1 does not define them).
+#include <osv/fs_flags.h>
+// Flags valid for the write path.
+#define RWF_WRITE_SUPPORTED (RWF_HIPRI | RWF_DSYNC | RWF_SYNC | RWF_APPEND)
+// Flags valid for the read path.  DSYNC/SYNC/APPEND are write-only per Linux;
+// NOWAIT we cannot honor (OSv's read path can block) so it is rejected rather
+// than faked with EAGAIN, which would make a retrying caller spin forever.
+#define RWF_READ_SUPPORTED (RWF_HIPRI)
```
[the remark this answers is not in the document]
gburd: The comment is correct and now matches the (updated) PR description: RWF_NOWAIT is rejected with EOPNOTSUPP rather than faked with EAGAIN. The description previously disagreed; it has been fixed so all four (code, comment, test, description) are consistent on EOPNOTSUPP.

tests/tst-fs-syscalls.cc
```diff
@@ -0,0 +1,85 @@
...
+    off_t sz = lseek(fd, 0, SEEK_END);
+    assert(sz == (off_t)(strlen(msg) + strlen(tail)));
+
+    // RWF_DSYNC just forces a flush; still succeeds.
+    assert(pwritev2(fd, &wio, 1, 0, RWF_DSYNC) == (ssize_t)strlen(msg));
+
+    // RWF_NOWAIT can't be honored on a blocking file path.  Rather than fake
+    // EAGAIN (which would make a retrying caller spin forever), reject it.
+    errno = 0;
+    assert(preadv2(fd, &rio, 1, 0, RWF_NOWAIT) == -1 && errno == EOPNOTSUPP);
```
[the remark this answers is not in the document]
gburd: The test intentionally expects EOPNOTSUPP, which is the chosen behavior (see the RWF_NOWAIT thread above). The PR description has been corrected to say EOPNOTSUPP, so the test now matches the documented and implemented behavior and will catch regressions correctly.

include/api/osv/fs_flags.h
gburd: The copyright line is the intended one -- Copyright (C) 2026 Greg Burd, BSD as in the top-level LICENSE, matching the other new files in this PR.

include/api/osv/fs_flags.h
gburd: Good point. I kept it as a single osv/ header shared by the implementation and the test to guarantee they cannot drift, rather than copying it into api/uio.h (which, as you note, would break the musl symlink). If you'd prefer it surfaced at api/uio.h for source compatibility with Linux's <sys/uio.h>, I'm happy to add a thin api/uio.h that includes this one -- just say the word.

libc/pthread.cc
gburd: You were right -- that was a stale hunk from an earlier push. The current head's diff is feature-only (fs/vfs/main.cc, the header, tests, syscall wiring, exported symbols); libc/pthread.cc is no longer touched.

fs/vfs/main.cc
```diff
@@ -1121,14 +1183,33 @@ int renameat(int olddirfd, const char *oldpath,
...
+        // fstatat()+renameat() has a TOCTOU window: a concurrent create between
+        // the check and the rename can clobber the destination.  A correct
+        // implementation needs VFS/filesystem-level atomic support and should
+        // replace this.  It is offered meanwhile for the common non-racing case.
+        struct stat st;
+        if (fstatat(newdirfd, newpath, &st, AT_SYMLINK_NOFOLLOW) == 0) {
+            errno = EEXIST;
+            return -1;
+        }
+    }
```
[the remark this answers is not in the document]
gburd: Fixed. renameat2(RENAME_NOREPLACE) now only proceeds to the rename when fstatat() fails with ENOENT (destination truly absent). Any other fstatat() error (EOVERFLOW, EACCES, a transient failure) is propagated instead of being treated as "absent", so we can no longer overwrite an existing destination on an ambiguous stat error.

include/api/osv/fs_flags.h
```diff
@@ -0,0 +1,47 @@
...
+#include <sys/types.h>
+#include <sys/uio.h>
+
+#ifndef RWF_HIPRI
+#define RWF_HIPRI  0x00000001   /* high priority request */
+#define RWF_DSYNC  0x00000002   /* per-io O_DSYNC */
+#define RWF_SYNC   0x00000004   /* per-io O_SYNC */
+#define RWF_NOWAIT 0x00000008   /* per-io nonblocking mode */
+#define RWF_APPEND 0x00000010   /* per-io O_APPEND */
+#endif
```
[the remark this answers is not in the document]
gburd: Done: each RWF_* macro now has its own #ifndef guard instead of one #ifndef RWF_HIPRI covering the block, so a libc that defines a subset still compiles.

include/api/osv/fs_flags.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef RENAME_NOREPLACE
+#define RENAME_NOREPLACE (1 << 0)   /* don't overwrite target */
+#define RENAME_EXCHANGE  (1 << 1)   /* exchange source and dest */
+#define RENAME_WHITEOUT  (1 << 2)   /* whiteout source */
+#endif
```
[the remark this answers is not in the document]
gburd: Done: each RENAME_* macro is now guarded independently, so a toolchain that defines RENAME_NOREPLACE but not the newer RENAME_EXCHANGE/RENAME_WHITEOUT still compiles.

fs/vfs/main.cc
```diff
@@ -1121,14 +1172,41 @@ int renameat(int olddirfd, const char *oldpath,
     }
 }
 
+// RENAME_* flags (musl 1.2.1 does not define them).
```
nyh, line 1175: This comment seems to be misplaced?

fs/vfs/main.cc
```diff
@@ -1121,14 +1172,41 @@ int renameat(int olddirfd, const char *oldpath,
...
 extern "C" OSV_LIBC_API
 int renameat2(int olddirfd, const char *oldpath,
               int newdirfd, const char *newpath, unsigned int flags)
 {
-    if (flags) {
+    // Only RENAME_NOREPLACE is supported.  RENAME_EXCHANGE (atomic swap) and
+    // RENAME_WHITEOUT need filesystem-level support OSv's VFS does not provide,
+    // so reject anything other than RENAME_NOREPLACE with EINVAL in one check.
+    if (flags & ~(unsigned int)RENAME_NOREPLACE) {
```
nyh, line 1183: Why is the (unsigned int) case needed here, but not below?

include/api/osv/fs_flags.h
```diff
@@ -0,0 +1,59 @@
...
+#ifndef RWF_NOWAIT
+#define RWF_NOWAIT 0x00000008   /* per-io nonblocking mode */
+#endif
+#ifndef RWF_APPEND
+#define RWF_APPEND 0x00000010   /* per-io O_APPEND */
+#endif
+
+#ifndef RENAME_NOREPLACE
+#define RENAME_NOREPLACE (1 << 0)   /* don't overwrite target */
```
nyh, line 37: Should this be 1u or something, to make this unsigned int? How is it defined on Linux?

nyh (approved):
Looks mostly good, only left some trivial nitpicks.
However, the patch needs to be rebased because it has conflicts.

PR #1430, opened 2026-07-12
mm: add node-preferring page allocation (alloc_page_on_node)

gburd:
## What

First incremental step of a node-aware allocator, on top of the NUMA topology
discovery (#1418) and getcpu/get_mempolicy (#1419) work. Adds
`memory::alloc_page_on_node(node)`, which prefers a page whose physical memory
lies on the requested NUMA node, falling back to a normal allocation when that
node has no free memory in the global page-range allocator (or when NUMA is
unavailable / the node is out of range).

## How

Layers on top of the existing allocator without restructuring it:
`page_range_allocator::alloc_page_from_node()` walks the free lists, finds a
range whose physical address resolves (via `numa::node_of_phys()`) to the
requested node, and carves one page the same way `alloc()` does; if none is
found it returns `nullptr` and `alloc_page_on_node()` falls back.

This is deliberately a best-effort hint, **not** full per-node pools: OSv drains
much of physical memory into per-CPU L1/L2 pools that are not yet
node-partitioned, so once a node's global ranges are exhausted the allocator
falls back. It gives correct placement while node-local global memory is
available and never fails or misplaces a page. Full per-node L1/L2 pools are
future work; the scheduler affinity and `mbind`/`set_mempolicy` enforcement
build on this hint.

## Testing

`tests/tst-numa-alloc.cc`: basic allocation and writability, out-of-range and
negative node fall back cleanly, and (with QEMU `-numa`) node-0 allocations land
on node 0 while other nodes fall back cleanly with no crash or misplacement.
Passes on OSv under KVM single-node and with a 2-node `-numa` config; `tst-mmap`
(the shared allocator path) unaffected.

## Note

Depends on #1418 and #1419 (this branch stacks on them); until they merge the
diff shows their commits too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1131,6 +1131,7 @@ endif
 objects += linux.o
 objects += core/commands.o
 objects += core/sched.o
+objects += core/numa.o
 objects += core/mmio.o
 objects += core/kprintf.o
 ifeq ($(conf_tracepoints),1)
--- a/core/mempool.cc
+++ b/core/mempool.cc
@@ -23,6 +23,7 @@
 #endif
 #include <atomic>
 #include <osv/mmu.hh>
+#include <osv/numa.hh>
 #include <osv/trace.hh>
 #include <lockfree/ring.hh>
 #include <osv/percpu-worker.hh>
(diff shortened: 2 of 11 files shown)


gburd:
Superseded by #1441 (rebased onto current master; GitHub auto-closed this one when the branch was updated). Please review #1441 instead.

PR #1442, opened 2026-07-13
net: forward-port FreeBSD IPv6 stack with SLAAC and DHCPv6 autoconfiguration

gburd:
This forward-ports the FreeBSD IPv6 stack to current OSv and adds the boot-time
autoconfiguration a unikernel needs (SLAAC and a DHCPv6 client), so an
IPv6-only or dual-stack application works out of the box. All IPv6 code is
compile-time gated so IPv6-less builds stay small.

## What this does

- **netinet6 stack** -- imports the FreeBSD 9 `netinet6` sources (ip6, icmp6,
  nd6, in6_*, udp6, raw_ip6, mld6, frag6, dest6, route6, scope6) adapted for
  OSv. 45 new files under `bsd/sys/netinet6/`, plus the shared eth/tcp/udp code
  taught about IPv6.
- **OSv fast paths for IPv6** -- the net-channel classifier, LRO, and RCU
  connection lookup are extended to IPv6/TCP so v6 gets the same in-kernel fast
  path as v4 (mirrors the existing IPv4 optimizations).
- **Linux socket compatibility** -- 43 `IPV6_*` socket options translated in the
  Linux compat layer (V6ONLY, PKTINFO, HOPLIMIT, TCLASS, MULTICAST_*,
  JOIN/LEAVE_GROUP, etc.), plus `getifaddrs()`/`if_nameindex()` over a NETLINK
  socket and AAAA-aware `getaddrinfo`.
- **SLAAC** -- OSv has no `rtsold`, so the kernel now sends its own Router
  Solicitation (`nd6_rs_output`) at interface bring-up and processes the Router
  Advertisement to form a global address. Also fixes an ordering bug where
  `ND6_IFF_ACCEPT_RTADV` was never set on interfaces that attached before the
  accept-RA flag was enabled at boot.
- **DHCPv6 client** -- a minimal RFC 8415 stateful client
  (`bsd/porting/dhcp6.cc`): SOLICIT/ADVERTISE/REQUEST/REPLY with DUID-LLT and
  IA_NA, plus Rapid Commit (two-message SOLICIT/REPLY). The loader chooses SLAAC
  vs DHCPv6 from the RA flags: an Autonomous-prefix RA (radvd, QEMU SLIRP) uses
  SLAAC, a Managed-flag RA (many enterprise/cloud networks) uses DHCPv6. Works
  with standard servers (isc-dhcp, kea, dnsmasq).
- **Boot options** -- `--ipv6` selects the IPv6-only boot path (skips the
  blocking DHCPv4 wait); static `--ip`/`--defaultgw`/`--nameserver` already
  accept v6 addresses.

## Compile-time gating

Everything is behind `conf-INET6` (the FreeBSD `INET6` macro) and, for the
DHCPv6 client, `conf_networking_dhcp6`. A `conf-INET6=0` build compiles and
links with no IPv6 code and no dangling references (verified), saving ~2.7 MB.
The default remains `conf-INET6=1`.

## Testing

- `tst-tcp-v6-smoke` (new, no boost): IPv6 + IPv4 loopback TCP echo, concurrent
  dual-stack, and the socket-option/multicast/getaddrinfo API surface. Passes.
- `tst-ifaddr-dump` (new): dumps interface addresses to observe SLAAC/DHCPv6.
- `tst-ipv6-internet` (new, manual): DNS AAAA + IPv6 TCP + HTTP to a real host.
- Existing `tst-tcp-v6`, `tst-pktinfo`, `tst-socket-timestamp` build and run.

Validated on KVM/QEMU (SLIRP `ipv6=on` -> SLAAC) and on a bare-metal host:
- SLAAC forms a global address from a Router Advertisement.
- The DHCPv6 client obtains a global address from a standard dnsmasq DHCPv6
  server (SOLICIT with Rapid Commit -> REPLY -> address bound), confirming the
  client is standards-compliant and not tied to any specific network.
- IPv6-only boot (no IPv4 on the link) reaches the IPv6 internet.

## Notes for review

- The history preserves the original ipv6-branch commits (FreeBSD import ->
  OSv adaptation -> net-channel -> Linux compat -> getifaddrs/netlink -> tests) so
  authorship is intact, with the SLAAC + DHCPv6 work as the final commit. Happy
  to squash if preferred.
- The DHCPv6 client lives in `bsd/porting/` (not `core/`) because it needs the
  BSD-internal netinet6 headers, which cannot coexist with the musl/glibc
  `in6_addr` that `boost::asio` pulls in; the DNS setter is reached through a
  small C-linkage shim in `libc/network/__dns.cc`.


gburd:
Follow-up (deliberately not in this PR): IPv6 inherits OSv's high-value fast paths — net channels (`classify_ipv6_tcp`, `tcp_net_channel_ipv6_packet`), LRO (`ETHERTYPE_IPV6` in `tcp_lro.cc`), the RCU connection classifier, and the AF-agnostic TCP header-prediction path. What it does not yet inherit is **virtio-net hardware checksum + TSO offload for IPv6**: `drivers/virtio-net.cc` hard-codes the offload switch to `ETHERTYPE_IP` (TX builds only `GSO_TCPV4`, RX has a literal `// How come - no support for IPv6?!`), and `if_hwassist`/`if_capabilities` advertise `IFCAP_TSO4` but not `IFCAP_TSO6` or the `CSUM_*_IPV6` bits.

Net effect: IPv6 is correct but computes checksums in software and doesn't get large-segment TSO, so high-throughput v6 sends/receives use more CPU than v4. This is a self-contained, driver-local performance follow-up (extend the TX/RX offload switch to `ETHERTYPE_IPV6` → `CSUM_TCP_IPV6`/`GSO_TCPV6`, add `IFCAP_TSO6`), best done and benchmarked as its own PR rather than expanding this correctness-focused one.

gburd:
Heads-up before merging: this branch was last synced against an older master (via a merge commit), so although GitHub reports it mergeable, its tree predates several tests that have since landed (tst-iovcnt-guard, tst-splice, tst-fs-syscalls, tst-membarrier, tst-mremap) and it still edits the `tests :=` literal in modules/tests/Makefile directly. A clean rebase onto current master hits genuine net-stack conflicts (net_channel.cc, tcp_input.cc, routecache.hh) from the merged IPv6 net-channel/classifier changes, so this needs a real forward-port rather than a mechanical Makefile fixup. I will rebase it onto current master and convert the test entries to standalone `tests += ` lines (per #1469) before it is ready; please hold this one until then. The smaller syscall/mm/ext4 PRs are already converted and ready.

PR #1444, opened 2026-07-13
[STACKED on #1442] virtio-net: enable IPv6 checksum and TSO offload

gburd:
> [!NOTE]
> **Stacked on #1442** (the IPv6 forward-port). This branch is based on
> `pr/ipv6-forward-port`, so the diff below is cumulative and re-includes the
> IPv6 stack. Once #1442 merges I will rebase onto master and this reduces to
> the two-file `drivers/virtio-net.{cc,hh}` change. Please review #1442 first.

Performance follow-up to #1442. IPv6 already inherits OSv's high-value fast
paths (net channels, LRO, RCU connection classifier, AF-agnostic TCP header
prediction), but the virtio-net **hardware offload** paths were hardcoded to
`ETHERTYPE_IP`:

- TX built only `VIRTIO_NET_HDR_GSO_TCPV4`;
- RX rejected everything but IPv4 (with a literal `// How come - no support for
  IPv6?!` comment);
- `if_hwassist`/`if_capabilities` advertised `IFCAP_TSO4` but not `IFCAP_TSO6`
  or the `CSUM_*_IPV6` bits.

So IPv6 computed checksums in software and never used TSO, burning more CPU than
IPv4 on high-throughput sends/receives even though `tcp_output` already emits
`CSUM_TCP_IPV6`.

This change:
- negotiates `VIRTIO_NET_F_HOST_TSO6` / `VIRTIO_NET_F_GUEST_TSO6`;
- advertises `IFCAP_TSO6` and `CSUM_TCP_IPV6 | CSUM_UDP_IPV6`;
- TX: adds an `ETHERTYPE_IPV6` case (`csum_start` past the 40-byte base header,
  `gso_type = VIRTIO_NET_HDR_GSO_TCPV6`); packets with extension headers have
  `ip6_nxt != TCP/UDP` and fall back to software checksums;
- RX: accepts `ETHERTYPE_IPV6` in the checksum-offload path (the TCP/UDP
  checksum-offset logic is identical for v4 and v6);
- uses large receive buffers when `GUEST_TSO6` is negotiated.

Boot-tested with IPv6 SLAAC over virtio-net (loopback + concurrent dual-stack
TCP pass, global address configured through the modified TX/RX paths). A
throughput A/B (v6 with vs without offload) is the natural benchmark once the
base merges.


gburd:
This branch is stacked on the full IPv6 forward-port (#1442), whose base predates several merged commits and needs a real net-stack forward-port (net_channel.cc / tcp_input.cc / routecache.hh conflicts against the merged IPv6 net-channel classifier changes), not just a mechanical Makefile fixup. I will rebase this on top of the forward-ported #1442 and convert the test entries to standalone `tests += ` lines (per #1469) at that point. Please hold until #1442 is forward-ported.

PR #1445, opened 2026-07-13
[STACKED on #1441] mm: per-NUMA-node L2 page pools

gburd:
> [!NOTE]
> **Stacked on #1441** (node-preferring page allocation). This branch builds on
> `alloc_page_on_node`/`alloc_page_from_node` from #1441, so the diff is
> cumulative until #1441 merges; then it reduces to the single `core/mempool.cc`
> commit. Review #1441 first.

Extends the node-preferring allocator into the hot page-pool path. Previously a
single global L2 pool fed every CPU's per-CPU L1 cache and refilled from the
global `free_page_ranges` with no node awareness, so on a multi-socket machine
a CPU routinely ran on pages from a remote NUMA node.

Now, when the system has more than one NUMA node, each node gets its own L2 pool
whose `refill()` prefers pages physically local to that node
(`alloc_page_from_node`), and each CPU draws from its own node's L2 via
`get_l2()` (keyed on `numa::node_of_cpu`). A drained node falls back to any page
so allocation always makes progress. Single-node / non-NUMA systems keep the
original single `global_l2` with zero behavior change.

## Validation

- Single-node boot: unchanged (fallback path), `tst-numa-alloc` passes.
- **A real 2-socket NUMA bare-metal host (node0 ~193 GB / node1 ~193 GB, 96
  cores)**, OSv under KVM with two NUMA nodes bound to the real host nodes:
  ```
  NUMA: 2 node(s), 8 CPU(s) mapped, 4 memory range(s)
  node 0: 32 on-node, 0 fell back
  node 1: 32 on-node, 0 fell back
  ```
  100% node-local allocation on both nodes.
- QEMU-emulated NUMA (no host-node binding) shows a higher node falling back,
  because emulated memory is concentrated on node 0 -- an emulation artifact;
  real hardware with genuinely node-distributed memory allocates locally.

`tst-numa-alloc` reports per-node on-node/fell-back counts as a regression
signal.

Known follow-up (separate): OSv's early memory init can still concentrate
`free_page_ranges` on lower nodes on some configs; this PR is the allocator-side
half and is correct on real hardware. Boot-time physical-range balancing across
nodes would be a further step if a workload needs it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1133,6 +1133,7 @@ endif
 objects += linux.o
 objects += core/commands.o
 objects += core/sched.o
+objects += core/numa.o
 objects += core/mmio.o
 objects += core/kprintf.o
 ifeq ($(conf_tracepoints),1)
--- a/core/mempool.cc
+++ b/core/mempool.cc
@@ -802,18 +802,56 @@ page_range* page_range_allocator::alloc(size_t size, bool contiguous)
     return &pr;
 }
 
+// Return the byte offset of the first page-aligned page within the physical
+// range [base, base+size) that lies on NUMA node `node`, or (size_t)-1 if the
+// range has no memory on that node.  Uses the SRAT memory ranges to intersect
+// directly rather than probing page by page, so a multi-hundred-MB free range
+// costs a handful of comparisons.
+static size_t node_offset_within(uint64_t base, size_t size, int node)
+{
+    uint64_t end = base + size;
+    uint64_t best = end;   // lowest on-node start found so far
+    for (auto& r : numa::memory_ranges()) {
+        if ((int)r.node != node) {
+            continue;
+        }
+        uint64_t lo = std::max(base, r.base);
+        uint64_t hi = std::min(end, r.base + r.length);
+        if (lo < hi) {
+            // Align the intersection start up to a page; must still leave room
+            // for one page inside both the free range and the node's range.
+            uint64_t start = align_up(lo, (uint64_t)page_size);
+            if (start + page_size <= hi && start < best) {
+                best = start;
+            }
+        }
+    }
+    return best == end ? (size_t)-1 : (size_t)(best - base);
+}
+
 page_range* page_range_allocator::alloc_page_from_node(int node)
 {
-    // Find a free range whose memory lies on the requested NUMA node, then
-    // carve a single page from its front (mirroring alloc()).  This is a
-    // best-effort preference: a linear scan over the free lists, returning the
-    // first range on the node.  Returns nullptr if none is found, so the caller
-    // can fall back to a node-agnostic allocation.
+    // Carve a single page physically on NUMA node `node` from the free lists.
+    //
+    // A free range may straddle a node boundary: OSv's boot allocator coalesces
+    // physically adjacent memory, and node N and node N+1 are usually adjacent
+    // in physical space, so one free range can begin on a low node and extend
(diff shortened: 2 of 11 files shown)


gburd:
Follow-up commit 425c77c2 on this branch closes the emulated-NUMA node-fallback that the earlier version had (real 2-socket hardware was already 32/32 on both nodes; this fixes the emulated case and any config where the boot allocator coalesces across node boundaries).

**Root cause:** `arch_setup_free_memory()` frees all E820 ranges into `free_page_ranges` at `.init` priority `mempool`, *before* `acpi::init()`/`numa::init()` run — so NUMA topology is unknown when memory is registered. The page-range allocator then coalesces physically-adjacent free memory, and since node N and N+1 are adjacent in physical space the whole machine's RAM collapses into essentially one giant free range whose `page_range` header sits at its low base. `alloc_page_from_node()` classified a range solely by `node_of_phys(base)`, so a higher node's memory living *inside* a low-node-based range was invisible and every allocation for it fell back. It was never a draining problem (L2 caps at ~8 MB/node) — it was range classification.

**Fix:** `alloc_page_from_node` is now boundary-aware — for each free range it finds the first page-aligned offset intersecting the requested node's SRAT range and carves there, splitting into up-to-three pieces (prefix / carved page / suffix), mirroring the existing `alloc_aligned` 3-way split. Base-on-node reduces to the old front-carve. Only mutates after the scan completes (no iterator invalidation); single-node is unaffected (`alloc_page_from_node` is unreachable when nr_nodes<=1).

**Emulated QEMU NUMA:** 2-node node-1 went 0/32→32/32; 4-node all four nodes 32/32. Single-node regression + tst-mmap still pass.

gburd:
Rebased onto master; converted the test entries to standalone `tests += tst-numa.so` / `tst-numa-mempolicy.so` / `tst-numa-alloc.so` lines so they no longer conflict in modules/tests/Makefile (see #1469). (Stacked on #1437/#1438/#1441.)

gburd:
## Rebased onto master (`0e34e4dc2`) - conflict resolved, still a draft

Same add/add cause as #1438 and #1441: this branch carried its own copy of
`numa: discover NUMA topology from ACPI SRAT/SLIT`, which is now on master as
`a92c44f9b`. Dropped that copy, rebased the rest.

```
def93eb13 copyright: correct new-file attribution to the actual author
d0d33f818 mm: split free ranges at NUMA node boundary in alloc_page_from_node
7cb83b9ff mm: per-NUMA-node L2 page pools
b12695c91 mm: add node-preferring page allocation (alloc_page_on_node)
fc000757c numa: report real topology from getcpu and get_mempolicy
```

The bottom two commits belong to #1438 and #1441, so the stacking is unchanged.

Verification (no CI on this repo, these are my own runs):

| check | result |
|---|---|
| compiles | `make build/release.x64/core/{numa,mempool}.o build/release.x64/linux.o` -> all TUs build, no errors, no warnings |
| change is live, not a no-op | `mempool.o` `4ec67e66` on a master-only build -> `a4fa2c73` here |
| merges clean | `merge-tree --write-tree --messages upstream/master HEAD` -> 0 conflicts, from a master-base worktree |
| signed | all five commits `%G? == G` |

**Staying a draft.** The rebase makes it mergeable and reviewable, but the
per-node L2 pool sizing still wants a NUMA-hardware measurement before I would
ask anyone to take it. This was a sync, not a readiness claim.

PR #1447, opened 2026-07-13
libext: ext4 sequential-read performance (read-ahead + async prefetch)

gburd:
> [!IMPORTANT]
> **Draft: one previously cited number has been withdrawn as invalid.** See "Withdrawn" below. The read results are retained but were measured on a base predating af27ba4b7, so they are pending re-validation. Not ready to merge until re-measured.

#1431 is now merged upstream as af27ba4b7, so this no longer depends on it and stands alone. Rebased onto current master (0e34e4dc2); the two fsync/dtime commits that used to sit at the base of this branch are gone, since upstream now carries them as af27ba4b7 and bf16817c9.

Rebased branch, since force-push is not available to me on this PR's head branch:

- branch: `pr/ext4-perf-v2`
- head: `8f95be01e2ecf9192b279aacf0370fd2e1157379`
- merges clean into master: `git merge-tree --write-tree --messages upstream/master pr/ext4-perf-v2` exits 0 with no conflicts

## The gap this addresses

`ext_read` issued one synchronous bio per `read()` with no prefetch (`ext4_blocks_get_direct` bypasses lwext4's block cache), so a stream of sequential reads left the NVMe queue idle during application compute. Linux keeps the device busy with multi-window async read-ahead. Baseline ext4 sequential read was roughly 560 MB/s against Linux's ~2300 MB/s on the same device, while 4K random read was already at or above Linux.

## Changes (6 commits)

1. **benchmark** (`tst-ext4-bench`): sequential write+fsync / sequential read / 4K random, reported in MB/s, filesystem-agnostic. Writes an absolute-offset pattern and verifies every byte on read, so it doubles as a read-path correctness test.
2. **direct user-buffer read/write**: skip the per-op bounce allocation and extra `uiomove` copy on the single-iovec fast path; also fixes a latent `free()`/`free_contiguous_aligned()` mismatch. Correctness cleanup, not the bottleneck.
3. **synchronous 1 MiB read-ahead**: on a sequential pattern, read one window and serve subsequent in-window reads from a memcpy.
4. **async double-buffered prefetch**: two windows plus a per-vnode prefetch worker that fills `next` while the application consumes `cur`. A worker thread (not raw `bio_done`) reuses `ext_internal_read` verbatim; writes and truncation invalidate, `~ext_vdata` joins the worker and frees.
5. **deep multi-window prefetch ring**: generalises to an N-window ring with M workers.
6. **ring tuning** to N=4/M=4, with `RA_WINDOWS_N`/`RA_WORKERS_M` overridable at build time.

## Read results (pre-af27ba4b7 base, retained but pending re-validation)

Measured A/B on a 2-socket x86-64 bare-metal host booting OSv under KVM against local NVMe, ext4 4K/no-journal, median of 3, no VERIFY FAIL:

| block size | sync read-ahead | async prefetch | speedup |
|---|---|---|---|
| 64 KiB  | ~505 MB/s | ~1248 MB/s | 2.5x |
| 128 KiB | ~534 MB/s | ~1260 MB/s | 2.4x |
| 256 KiB | ~600 MB/s | ~1250 MB/s | 2.1x |

Reference on the same NVMe: Linux ext4 buffered read ~2300 MB/s; NVMe single-stream O_DIRECT ceiling 1.5 GB/s at 128K, 1.9 GB/s at 1M. The remaining gap to Linux is queue depth: Linux prefetches many windows, this prefetches a bounded ring, and virtio-blk's `make_request()` serialises submission under a single lock, which caps effective device queue depth at roughly 2 to 4.

**Why these read numbers are probably still valid on the current base**, stated as an argument a reviewer can check rather than something to take on trust:

- af27ba4b7 changed `ext_fsync` from `vop_nullop` to a flush of lwext4's block cache, and added `vop_cache` for mmap page-cache warming.
- The read path does not go through that block cache. `ext_internal_read` reaches the device via `ext4_blocks_get_direct` -> `ext4_bdif_bread`, which addresses the block device directly and never consults the cache `ext_fsync` flushes.
- The benchmark's read phases use `read()`/`pread()`, not mmap, so `vop_cache` is not on their path either.
- `fsync()` is called only inside `bench_write`'s timing window (`tst-ext4-bench.cc:56`); neither read phase calls it.

**Caveat that keeps this "probably" rather than "confirmed":** the benchmark writes a file and then reads that same file. With a real fsync the dirty state at read time is different, so the read number could shift second-order even though the mechanism above does not touch the read path. Likely intact is not measured.

## Withdrawn

The earlier revision of this PR cited roughly 620 MB/s for sequential write with fsync, and used it to claim write throughput was at parity with Linux. **Both the number and the parity claim are withdrawn.**

That measurement was taken on a base where `ext_fsync` was `#define ext_fsync ((vnop_fsync_t)vop_nullop)`, so `fsync()` returned without flushing anything and the benchmark's timer stopped before the data was durable. It was not measuring durable write throughput. Since `ext_fsync` is now a real block-cache flush (af27ba4b7), the figure is invalid on the current base and I am not restating it with a caveat. There is currently no supported claim about write throughput in this PR.

Nothing in this series targets the write path, so this is a withdrawn measurement rather than a regression, but the honest position is that write performance on the current base is unmeasured.

### The withdrawn claim still appears in a commit message

Disclosing the current state of the branch rather than only the description,
because the two disagree. The withdrawal above is in this description only. One
commit message on the head still contains it:

- `f20b91d75` ("libext: sequential read-ahead cache for ext4 reads") opens with
  "The local-NVMe A/B showed ext4 sequential reads at ~33% of Linux while writes
  were at parity and random reads ~86%." The withdrawn write-parity claim is
  load-bearing there: it is the stated motivation for the commit.
- The same text is present in the rebased `pr/ext4-perf-v2` as `4ca6429df`, so
  it is not an artifact of this branch being stale.

This is stated as a disclosure of what the branch contains today, not as a
promise to fix it on some later date. A commit message can only be changed by
rewriting the commit, and this branch is already blocked on re-measuring the
read path on a base containing af27ba4b7 (see "To re-validate"), which will
rebuild these commits anyway. Correcting the message before that measurement
exists would mean rewriting it twice. So the sequence is: re-measure first, then
rewrite these commits with both the corrected numbers and the corrected
motivation. Until that happens the description is authoritative and the commit
message is not.

For a reviewer who wants to check this rather than take it on trust:

```
git log --format='%h %s%n%b' upstream/master..06210dfcf | grep -n 'were at parity'
```

## To re-validate

On a 2-socket x86-64 bare-metal host (KVM, local NVMe, ext4 4K/no-journal), with a base containing af27ba4b7:

1. The 64K / 128K / 256K sequential-read A/B sweep, median of 3, confirming the read speedup holds now that fsync is real.
2. A fresh sequential-write-with-real-fsync number, to replace the withdrawn one and settle whether write is at, above, or below Linux parity.

Staying in draft until both are done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ modules/dl_tests/usr.manifest
 compile_commands.json
 downloaded_packages
 .firecracker
+ext_images/
--- a/modules/libext/Makefile
+++ b/modules/libext/Makefile
@@ -3,7 +3,7 @@ include ../common.gmk
 module_out := $(out)/modules/libext
 
 CXXFLAGS = -fPIC -std=gnu++11 $(INCLUDES) -I../lwext4/upstream/lwext4/include -I../lwext4/upstream/lwext4/build_lib_only/include \
-	 -D_KERNEL -D_GNU_SOURCE -Wall -fno-exceptions -fno-rtti -O2
+	 -D_KERNEL -D_GNU_SOURCE -Wall -fno-exceptions -fno-rtti -O2 $(RA_FLAGS)
 
 # the build target executable:
 TARGET = libext
--- a/modules/libext/ext_vnops.cc
+++ b/modules/libext/ext_vnops.cc
@@ -68,6 +68,16 @@ void free_contiguous_aligned(void* p);
 #define ext_debug(...)
 #endif
 
+// Mark an inode as deleted with a non-zero deletion time before it is freed, so
+// Linux e2fsck does not flag "deleted inode has zero dtime".  lwext4's own
+// delete path uses ext4_inode_set_del_time(inode, -1L), but that symbol is not
+// exported from liblwext4.so, so set the field directly.  0xffffffff is
+// byte-order invariant, so no to_le32() is needed.
+static inline void ext_mark_inode_deleted(struct ext4_inode *inode)
+{
+    inode->deletion_time = 0xffffffff;
+}
+
 //Simple RAII struct to automate release of i-node reference
 //when it goes out of scope.
 struct auto_inode_ref {
@@ -168,6 +178,10 @@ void ext_delete_outstanding_inodes(struct ext4_fs *fs)
         if (inode._r != EOK) {
             continue;
         }
+        // Set a non-zero deletion time before freeing so Linux e2fsck does not
+        // flag "deleted inode has zero dtime" (lwext4's own delete path does the
+        // same with -1L).
+        ext_mark_inode_deleted(inode._ref.inode);
         ext4_fs_free_inode(&inode._ref);
         ext_debug("delete: i-node=%ld when unmounting\n", inode_no);
     }
@@ -195,6 +209,7 @@ ext_close(vnode_t *vp, file_t *fp)
         if (inode._r != EOK) {
             return inode._r;
         }
+        ext_mark_inode_deleted(inode._ref.inode);
         ext4_fs_free_inode(&inode._ref);
 
         remove_inode_from_being_deleted(vp->v_ino);
@@ -885,6 +900,7 @@ ext_dir_link(struct vnode *dvp, char *name, int file_type, mode_t mode, uint32_t
         ext_debug("dir_link: created %s under i-node=%li with filetype:%d\n", name, dvp->v_ino, file_type);
     } else {
         if (!inode_no) {
+            ext_mark_inode_deleted(child_ref.inode);
             ext4_fs_free_inode(&child_ref);
         }
         //We do not want to write new inode. But block has to be released.
@@ -1065,6 +1081,7 @@ ext_dir_remove_entry(struct vnode *dvp, struct vnode *vp, char *name)
 
             if (links_cnt == 1) {//Zero now
                 if (vdata->ref_count == 0) {
+                    ext_mark_inode_deleted(child._ref.inode);
                     ext4_fs_free_inode(&child._ref);
                 } else {
                     ext_debug("dir_remove_entry: should remove i-node=%ld of %s on last close\n", vp->v_ino, name);
@@ -1075,6 +1092,7 @@ ext_dir_remove_entry(struct vnode *dvp, struct vnode *vp, char *name)
         }
     } else {
         if (vdata->ref_count == 0) {
+            ext_mark_inode_deleted(child._ref.inode);
             ext4_fs_free_inode(&child._ref);
         } else {
             ext_debug("dir_remove_entry: should remove i-node=%ld of %s on last close\n", vp->v_ino, name);
--- a/modules/libext/ext_vnops.cc
+++ b/modules/libext/ext_vnops.cc
@@ -19,6 +19,18 @@
 //The libext does not implement journal (we can integrate it later and make it optional)
 //nor xattr which is not even supported by OSv VFS layer.
 
+// The libext module is built with -fno-rtti, but <osv/pagecache.hh> pulls in
+// <osv/mmu.hh> -> <osv/trace.hh> which uses typeid.  We only need a couple of
+// symbols from the page cache to warm it in ext_map_cached_page(), so declare
+// them minimally here instead of including the heavy headers.  The uio carries
+// the pagecache hashkey opaquely, so we never need its layout.
+#include <osv/pagealloc.hh>   // memory::alloc_page / free_page (light header)
+namespace pagecache {
+    struct hashkey;
+    void map_read_cached_page(hashkey *key, void *page);
+}
+static const unsigned long EXT_PAGE_SIZE = 4096;
+
 extern "C" {
 #define USE_C_INTERFACE 1
 #include <osv/device.h>
@@ -37,6 +49,7 @@ void free_contiguous_aligned(void* p);
 #include <ext4_dir.h>
 #include <ext4_inode.h>
 #include <ext4_fs.h>
+#include <ext4_blockdev.h>
 #include <ext4_dir_idx.h>
 #include <ext4_trans.h>
 
@@ -596,7 +609,21 @@ ext_ioctl(vnode_t *vp, file_t *fp, u_long com, void *data)
     return (EINVAL);
 }
 
-#define ext_fsync     ((vnop_fsync_t)vop_nullop)
+static int
+ext_fsync(struct vnode *vp, struct file *fp)
+{
+    // lwext4 mounts with block-cache write-back enabled (ext_vfsops.cc), so a
+    // write only reaches the disk when the block cache is flushed.  fsync(2)/
+    // fdatasync(2) must make the file's data durable, so flush the device's
+    // block cache here (the cache is shared per device, so this persists this
+    // file's dirty buffers along with any others -- correct, if slightly more
+    // than the minimum a per-inode flush would do).
+    struct ext4_fs *fs = (struct ext4_fs *)vp->v_mount->m_data;
+    fs->bcache_lock();
+    int r = ext4_block_cache_flush(fs->bdev);
+    fs->bcache_unlock();
+    return r;
+}
 
 static int
 ext_readdir(struct vnode *dvp, struct file *fp, struct dirent *dir)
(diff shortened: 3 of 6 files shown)


wkozaczuk:
I see it is a draft that depends on #1431, which I hope to merge soon. I am really surprised by the performance results you list here - look very, very good compared to Linux. I am interested in what exactly your setup is; you mention nmve - emulated one? Do you compare to a Linux guest?

On my machine, when I run the fio read test with ext and compare it to zfs on OSv, I get pretty terrible results:
- zfs
```
./scripts/build image=fio fs=zfs 

./scripts/run.py -e '/fio --thread --name=fiotest --filename=/tmp/test --numjobs=4 --rw=read --size=128M' 
OSv v0.57.0-383-g894d5879
eth0: 192.168.122.15
Booted up in 264.45 ms
Cmdline: /fio --thread --name=fiotest --filename=/tmp/test --numjobs=4 --rw=read --size=128M
fiotest: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
...
fio-3.29-dirty
Starting 4 threads
fiotest: Laying out IO file (1 file / 128MiB)
...
Run status group 0 (all jobs):
   READ: bw=1463MiB/s (1534MB/s), 366MiB/s-444MiB/s (383MB/s-466MB/s), io=512MiB (537MB), run=288-350msec
```

- ext
```
./scripts/build image=fio fs=ext
 
./scripts/run.py -e '/fio --thread --name=fiotest --filename=/tmp/test --numjobs=4 --rw=read --size=128M' 
OSv v0.57.0-383-g894d5879
eth0: 192.168.122.15
Booted up in 189.80 ms
Cmdline: /fio --thread --name=fiotest --filename=/tmp/test --numjobs=4 --rw=read --size=128M
fiotest: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
...
fio-3.29-dirty
Starting 4 threads
fiotest: Laying out IO file (1 file / 128MiB)
...
Run status group 0 (all jobs):
   READ: bw=1113KiB/s (1140kB/s), 278KiB/s-278KiB/s (285kB/s-285kB/s), io=512MiB (537MB), run=471055-471062msec
```
1.5 K slower

When I increase the test block size to 128K from the default 4K, it gets a little better:
- zfs
```
READ: bw=7314MiB/s (7670MB/s), 1829MiB/s-3200MiB/s (1917MB/s-3355MB/s), io=512MiB (537MB), run=40-70msec
```

-ext
```
READ: bw=40.4MiB/s (42.3MB/s), 10.1MiB/s-10.1MiB/s (10.6MB/s-10.6MB/s), io=512MiB (537MB), run=12667-12677msec
```
150-200 times slower.

You mention the baseline. Is it what you see before your changes or after?

gburd:
Thanks for digging in, and the reconciliation is worth doing carefully because I think we are measuring two different things.

## The setup

The numbers in the PR body are an A/B of **ext-before vs ext-after this branch's four perf commits**, single-stream, on the same disk:

- Host: a 2-socket x86-64 bare-metal machine, OSv booting under KVM/QEMU with a **local NVMe** backing (no throughput cap), virtio-blk.
- Guest FS: ext4 formatted `-b 4096` with the journal off, mounted via lwext4.
- Workload: `tst-ext4-bench` (in commit 1 of this branch), which does **one** sequential `read()` stream front-to-back over a single file at a fixed block size (64K/128K/256K), and verifies every byte. Median of 3 runs.
- Linux reference: the identical file/pattern read on a Linux guest ext4 on the same NVMe, for the ~2300 MB/s buffered-read number.

So "baseline" in the table means **ext on this branch's base (#1431), before the read-ahead/prefetch commits** vs **ext with them**. It is not ext-vs-zfs.

## Why your fio run shows the opposite

I think there are three separate reasons your `fio` result and this PR's result diverge, and none of them contradict each other:

1. **The build you tested has none of this PR's changes.** Your console shows `OSv v0.57.0-383-g894d5879`, and `894d5879` is exactly the HEAD of #1431 (`libext: set inode deletion time...`), i.e. the **base** PR. The four read-ahead/prefetch commits live only on top of that, in #1447. So your `fs=ext` run measured the *un-prefetched* path -- which is precisely the ~40 MB/s @128K "before" case this PR is trying to fix. In that sense your number corroborates the PR: baseline ext seq read is bad, and that is the whole motivation. To see the after-number you would need to build with this branch's tip, not #1431's.

2. **`--numjobs=4` on one file is a very different access pattern from the bench.** `fio ... --numjobs=4 --filename=/tmp/test` runs four threads all reading the *same* vnode. The read-ahead in this PR is a **per-vnode single prefetch window** (`cur`/`next`), so four concurrent streams on one vnode contend for and repeatedly reposition that one window instead of each getting a clean sequential run -- worst case for a single-window prefetcher, especially at the 4K default block size. `tst-ext4-bench` is single-stream, which is what the window is designed for. A single-`numjobs`, 128K `psync` fio would be the apples-to-apples read of this branch's tip.

3. **ext != zfs; the fix is lwext4-specific.** Your fast number is zfs, which has the ARC plus its own multi-stream prefetch, so it is not surprising it beats un-prefetched lwext4 by a large factor. This PR does not touch the zfs path at all -- it is entirely in `ext_read`/lwext4. So "zfs is much faster than ext on OSv" is true and expected, and it does not bear on the ext-before/ext-after delta this PR claims. The remaining ext-vs-zfs gap is real and separate work (deeper prefetch pipeline, routing ext reads through the page-cache read path, a zero-copy borrow-and-pin bridge like the ARC one).

## Suggested apples-to-apples

If you build this branch's tip and run `tst-ext4-bench` on your NVMe I would expect the ~1250 MB/s single-stream after-number to reproduce (or tell us where the host differs). A single-stream `fio --numjobs=1 --bs=128k --ioengine=psync` on the branch tip should land in the same ballpark; the `--numjobs=4`-same-file case will stay low until the prefetch pipeline is per-stream/deeper, which is the documented upgrade path here.

I do not want to overclaim: this closes most of the **single-stream sequential** gap for lwext4, it exceeds the 128K O_DIRECT single-stream ceiling on this host, and it is still ~within 2x of Linux buffered read and behind zfs. Multi-stream and the ext-vs-zfs gap are follow-ups.

gburd:
Rebased onto master; converted the test entries to standalone `ext-only-tests += tst-ext4-rw.so` / `ext-only-tests += tst-ext4-bench.so` lines so they no longer conflict in modules/tests/Makefile (see #1469). (Stacked on #1431.)

gburd:
Rebased onto current master (0e34e4dc2) and the body rewritten. Summary of what changed and why one number is gone.

**Standalone now.** #1431 is merged upstream as af27ba4b7, so the two fsync/dtime commits that used to sit at the base of this branch are dropped and this reduces to the six performance commits. Title prefix removed.

**Rebased branch.** Force-push is not available to me on this PR's head branch, so the rebase is on a new branch rather than an update in place:

- branch: `pr/ext4-perf-v2`
- head: `8f95be01e2ecf9192b279aacf0370fd2e1157379` (signed)
- verified clean: `git merge-tree --write-tree --messages upstream/master pr/ext4-perf-v2` exits 0, no conflicts

Retarget this PR to that branch, or say the word and I will open a fresh one. Three things conflicted against current master and are resolved:

- `.gitignore` and `tests/tst-ext4-rw.cc` (add/add): both came from the two #1431 commits, which are dropped because upstream now carries them. `tst-ext4-rw.cc` is upstream's file now and this branch does not re-add it.
- `modules/libext/ext_vnops.cc`: resolved so upstream's merged work coexists with the prefetch ring. The `vnops` table is byte-identical to master's, and `ext_fsync`, `ext_map_cached_page`, the inode refcounting from c08544641 and the dtime marking from bf16817c9 are all intact on top of the ring. Compiles clean under the module's own flags (`-std=gnu++11 -Wall -fno-exceptions -fno-rtti -O2`) against current OSv and lwext4 headers.

**One number withdrawn.** The earlier revision cited roughly 620 MB/s for sequential write with fsync and used it to claim write parity with Linux. That was measured when `ext_fsync` was `vop_nullop`, so `fsync()` flushed nothing and the timer stopped before the data was durable. It was not measuring durable write throughput, and af27ba4b7 makes `ext_fsync` a real block-cache flush. I have deleted the number and the parity claim rather than restating them with a caveat. Write performance on the current base is unmeasured.

**Read numbers retained, explicitly pending re-validation.** They were taken on a pre-af27ba4b7 base. I checked whether af27ba4b7 could have flattered them and believe it cannot, for reasons a reviewer can verify: `ext_internal_read` reaches the device through `ext4_blocks_get_direct` -> `ext4_bdif_bread`, which never consults the block cache that `ext_fsync` flushes; the read phases use `read()`/`pread()` rather than mmap, so `vop_cache` is not on their path; and `fsync()` is called only inside `bench_write`'s timing window. The residual caveat is that the benchmark reads back the same file it just wrote, so real-fsync dirty state at read time could move the read number second-order. Likely intact is not measured, so the label stays.

**Staying in draft** until both are re-run on a 2-socket bare-metal host with local NVMe on a base containing af27ba4b7: the 64K/128K/256K read sweep, and a fresh write-with-real-fsync number. I would rather leave this draft with an honest gap than mark it ready on a number I know was measured against a no-op fsync.

gburd:
Status note so this draft is not mistaken for current.

This PR's head branch (`pr/ext4-perf`) predates upstream `af27ba4b7` ("ext: implement fsync"), and it now conflicts with master in 3 files. More importantly, every performance number originally posted here was measured when `ext_fsync` was still `vop_nullop`, i.e. a no-op, so the write-side figures measured nothing durable and have already been withdrawn from the description.

A rebased branch exists that does contain `af27ba4b7` and merges cleanly against master: `pr/ext4-perf-v2`. It carries the same five changes (direct read/write into the user buffer, sequential read-ahead, async double buffering, the multi-window prefetch ring, and the N=4/M=4 tuning). The two branches are divergent rebases rather than ancestor/descendant, so this PR's head cannot be fast-forwarded onto it.

Plan: leave this draft in place for now and re-measure the read path on a base containing `af27ba4b7` before re-proposing. The read-side argument does not depend on the fsync change (libext never places file data in the lwext4 block cache in either direction: reads go through `ext4_blocks_get_direct`/`ext4_block_readbytes` and writes through the `_direct`/`_writebytes` pair, so the cache holds only metadata), but the numbers still need to be re-taken rather than carried over.

Keeping it as a draft rather than closing it so the review context is preserved.

gburd:
Superseded by #1510. This PR's head sits on a base that predates the ext4 fsync/page-cache bridge (`af27ba4b7`) these read-ahead commits depend on, and a PR head cannot be repointed (the API accepts a new `head` and silently discards it). #1510 is the same six commits recreated on a base that contains `af27ba4b7`, merge-tree clean. Closing this one.

PR #1456, opened 2026-07-22
mm: per-child copy-on-write address space for fork() (opt-in, stacked on #1455)

gburd:
## What

Follow-up to #1455 (base thread-backed `fork()`). Gives a forked child its **own address space** with **copy-on-write** of private mappings, so a forked child gets real memory isolation like Linux `fork()` -- its private heap/globals/stack writes stay private, while `MAP_SHARED`/shm stays shared between parent and child. Like the base PR, this is gated entirely behind `CONFIG_fork` (default **off**): with the flag off, none of the address-space/COW/context-switch code is compiled and OSv's single-address-space model and context-switch path are byte-for-byte unchanged.

Stacked on #1455 -- please review/merge that first.

## How (with `CONFIG_fork` enabled)

- **`mmu::address_space`** object (page-table root + vma_list); the previous global becomes "address space 0" (kernel + init app).
- **Per-thread current address space**; the context switch loads the target CR3 only when the address space differs (a no-op for AS0-only workloads). Kernel PML4 entries are shared across all address spaces, so OSv code and the kernel heap work identically after a switch.
- **`fork()` clones the parent's vmas** into a child address space: private writable mappings are write-protected in both and copied on first write (reusing OSv's existing COW fault machinery); `MAP_SHARED`/shm map the same physical pages (truly shared). `execve()` returns the thread to AS0, and the child address space is destroyed exactly once (guarded against the exec-path also tearing it down).
- **Kernel-stack coherence:** lock/condvar `wait_record`s for fork-child (non-AS0) threads are allocated from the AS-shared kernel heap rather than the thread stack, so a `wait_record` queued on a shared kernel mutex resolves to the same physical page for any waker across address spaces.

## Validation (x86-64 / KVM, both flag states)

- `CONFIG_fork` **off** (default): builds `EXIT=0`, `nm loader.elf` shows zero COW/CR3 symbols (context switch has no address-space check compiled in), non-fork tests healthy (`tst-mmap` passes).
- `CONFIG_fork` **on**: builds `EXIT=0`; **`tst-fork` 10/10**; **`tst-fork-cow` 6/6** -- the proof: a forked child's private global is unchanged in the parent after the child writes it (**private stays private**), and the child's `MAP_SHARED` write is visible in the parent (**shared stays shared**); **`tst-execve` 3/3**; clean shutdown on every test.

## Known limitation (documented, non-blocking)

The child's stack is currently relocated+copied rather than same-VA COW, so a child that unwinds a very deep call chain (e.g. a multi-process server forking backends) can still hit a stack-fidelity fault -- `tst-fork-deep` fails by design here. The same-VA-stack fix (which requires addressing OSv's lack of a separate kernel stack) is a further follow-up. This PR delivers memory-isolated `fork()` with COW proven for the common cases.

Off by default, so no risk to existing OSv builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1057,6 +1057,9 @@ objects += arch/$(arch)/firmware.o
 objects += arch/$(arch)/hypervisor.o
 objects += arch/$(arch)/interrupt.o
 objects += arch/$(arch)/clone.o
+ifeq ($(conf_fork),1)
+objects += arch/$(arch)/fork.o
+endif
 ifeq ($(conf_drivers_pci),1)
 objects += arch/$(arch)/pci.o
 objects += arch/$(arch)/msi.o
@@ -1660,11 +1663,20 @@ musl += prng/srand48.o
 libc += random.o
 
 libc += process/execve.o
+ifeq ($(conf_fork),1)
+libc += process/fork.o
+endif
 musl += process/execle.o
 musl += process/execv.o
 musl += process/execl.o
 libc += process/waitpid.o
+# When fork() is enabled, our waitpid.cc provides wait()/waitpid()/wait4()
+# backed by the fork() child registry, superseding musl's process/wait.o.
+# When fork() is disabled, waitpid.cc provides only the ECHILD waitpid() stub,
+# so keep musl's wait.o for wait()/wait4().
+ifneq ($(conf_fork),1)
 musl += process/wait.o
+endif
 
 musl += setjmp/$(musl_arch)/setjmp.o
 musl += setjmp/$(musl_arch)/longjmp.o
--- a/arch/aarch64/fork.cc
+++ b/arch/aarch64/fork.cc
@@ -22,8 +22,9 @@
 extern "C" void __osv_run_atfork_child();
 
 sched::thread *fork_thread(void *caller_ret, void *caller_sp,
-                           void **out_stack_to_free)
+                           void *resume_ctx, void **out_stack_to_free)
 {
+    (void)resume_ctx;   // aarch64 uses caller_sp bias-copy (same-VA is x86-64)
     auto parent = sched::thread::current();
     auto parent_pinned_cpu = parent->pinned() ? sched::cpu::current() : nullptr;
 
--- a/arch/aarch64/fork.cc
+++ b/arch/aarch64/fork.cc
@@ -69,7 +69,15 @@ sched::thread *fork_thread(void *caller_ret, void *caller_sp,
            "br %1 \n\t"         // resume in fork()'s caller
            : : "r"(resume_sp), "r"(resume_pc) : "x0", "memory");
     }, sched::thread::attr().
-        stack(4096 * 4),
+        stack(4096 * 4).
+        // Detached: nobody join()s the fork child (the parent reaps it via the
+        // pid registry / waitpid, not sched::thread::join).  A detached thread
+        // is handed to the reaper on completion, which runs our set_cleanup()
+        // (freeing the copied stack and disposing the thread object, releasing
+        // its application_runtime reference).  Without this the thread object
+        // (and its app_runtime shared_ptr) would leak and OSv would hang at
+        // shutdown -- see the cleanup comment in libc/process/fork.cc.
+        detached(),
         false,
         true);
     t->set_app_tcb(parent->get_app_tcb());
--- a/arch/aarch64/fork.cc
+++ b/arch/aarch64/fork.cc
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2026 Greg Burd
+ *
+ * This work is open source software, licensed under the terms of the
+ * BSD license as described in the LICENSE file in the top-level directory.
+ *
+ * fork_thread() for aarch64 -- mirrors arch/x64/fork.cc.  fork() (fork.cc)
+ * passes the caller's return address and stack pointer; we copy the parent's
+ * user stack, bias the SP into the copy, and start a child thread that installs
+ * the copied stack, sets x0=0 (fork()'s return value in the child), and returns
+ * to fork()'s caller.
+ */
+
+#include "arch.hh"
+#include <errno.h>
+#include <string.h>
+#include <cstdlib>
+#include <osv/sched.hh>
+
+// pthread_atfork child-handler chain (defined in libc/pthread.cc), run in the
+// child's context before it resumes user code.
+extern "C" void __osv_run_atfork_child();
+
+sched::thread *fork_thread(void *caller_ret, void *caller_sp,
+                           void **out_stack_to_free)
+{
+    auto parent = sched::thread::current();
+    auto parent_pinned_cpu = parent->pinned() ? sched::cpu::current() : nullptr;
+
+    auto si = parent->get_stack_info();
+    char *stack_base = static_cast<char*>(si.begin) + si.size;
+    char *sp = static_cast<char*>(caller_sp);
+    if (sp < static_cast<char*>(si.begin) || sp > stack_base) {
+        return nullptr;
+    }
+    size_t stack_size = si.size;
+
+    char *child_stack_mem = static_cast<char*>(malloc(stack_size));
+    if (!child_stack_mem) {
+        return nullptr;
+    }
+    // Copy ONLY the live top [caller_sp .. stack_base) into the top of the
+    // child buffer (app stacks are demand-paged; copying from si.begin faults).
+    char *child_base = child_stack_mem + stack_size;
+    ptrdiff_t bias = child_base - stack_base;
+    size_t live = static_cast<size_t>(stack_base - sp);
+    memcpy(child_base - live, sp, live);
+    char *child_sp = sp + bias;
+
+    volatile u64 resume_sp = reinterpret_cast<u64>(child_sp);
+    volatile u64 resume_pc = reinterpret_cast<u64>(caller_ret);
+    char *stack_to_free = child_stack_mem;
+
+    // TLS: the child is a real OSv thread with its own fresh setup_tcb() block.
+    // Only override tpidr_el0 if the parent had installed its own app TCB via
+    // arch_prctl (parent_app_tcb != 0); otherwise keep the child's private OSv
+    // TLS (the clean case for a musl app built against OSv's libc).
+    u64 parent_app_tcb = parent->get_app_tcb();
+    auto t = sched::thread::make([resume_sp, resume_pc, parent_app_tcb] {
+        if (parent_app_tcb) {
+            asm volatile ("msr tpidr_el0, %0; isb" :: "r"(parent_app_tcb) : "memory");
+        }
+        // Run pthread_atfork child handlers in the child's context before
+        // resuming user code.
+        __osv_run_atfork_child();
+        asm volatile
+          ("mov sp, %0 \n\t"    // install the private copied stack
+           "mov x0, #0 \n\t"    // fork() returns 0 in the child
+           "br %1 \n\t"         // resume in fork()'s caller
+           : : "r"(resume_sp), "r"(resume_pc) : "x0", "memory");
+    }, sched::thread::attr().
+        stack(4096 * 4),
+        false,
+        true);
+    t->set_app_tcb(parent->get_app_tcb());
+    if (parent_pinned_cpu) {
+        t->pin(parent_pinned_cpu);
+    }
+    if (out_stack_to_free) {
+        *out_stack_to_free = stack_to_free;
+    }
+    return t;
+}
--- a/arch/x64/arch-switch.hh
+++ b/arch/x64/arch-switch.hh
@@ -95,14 +95,24 @@ void thread::switch_to()
     set_fsbase(reinterpret_cast<u64>(_tcb));
     barrier();
 #if CONF_fork
-    // Address-space (CR3) switch for fork COW.  Only touch CR3 when the target
+    // Address-space (CR3) switch for fork COW.  Only switch when the target
     // thread lives in a different address space than the outgoing one, so the
-    // common single-address-space case pays nothing (and no TLB flush).  The
-    // kernel half of every AS is identically mapped, so the switch code, kernel
-    // stacks and kernel heap remain valid across the write.
-    if (_current_as != old->_current_as) {
-        processor::write_cr3(mmu::pt_root_phys(_current_as));
-    }
+    // common single-address-space case pays nothing (and no TLB flush).
+    //
+    // CRITICAL (same-VA fork stacks): the CR3 write must happen EXACTLY at the
+    // rsp/rbp swap below, NOT here.  OSv runs kernel code (this switch, the
+    // fpu-cw/mxcsr scratch saves, the register-save asm) on the app thread's
+    // stack.  A forked child resumes on the parent's EXACT stack VAs, so parent
+    // and child alias the same stack virtual addresses (backed by different
+    // physical pages per address space).  If we loaded the incoming thread's
+    // CR3 here, the subsequent scratch writes to the OUTGOING thread's stack
+    // (fnstcw/stmxcsr at -0x34(%rbp), etc., with rsp/rbp still the old thread's)
+    // would resolve through the WRONG address space and clobber the incoming
+    // thread's physical page at that shared VA -- corrupting a blocked thread's
+    // live stack.  So we defer the CR3 load into the asm, coincident with the
+    // rsp/rbp swap: old-stack writes use the old AS, new-stack writes the new.
+    bool switch_as = (_current_as != old->_current_as);
+    u64 new_cr3 = switch_as ? mmu::pt_root_phys(_current_as) : 0;
 #endif
     auto c = _detached_state->_cpu;
     old->_state.exception_stack = c->arch.get_exception_stack();
@@ -119,6 +129,34 @@ void thread::switch_to()
     c->arch._current_thread_kernel_tcb = reinterpret_cast<u64>(_tcb);
     auto fpucw = processor::fnstcw();
     auto mxcsr = processor::stmxcsr();
+#if CONF_fork
+    if (switch_as) {
+        // Same-VA-safe switch: swap rsp/rbp to the incoming thread, THEN load
+        // its CR3 (mov %rax,%cr3), THEN jump.  All old-stack writes above ran
+        // under the old AS; every access after the rsp/rbp swap is to the new
+        // thread's stack, now resolved through the new AS -- so a shared stack
+        // VA never resolves through the wrong address space.  Kernel .text and
+        // the thread_state (heap) are identically mapped in every AS, so the
+        // instruction fetch across the mov-to-cr3 and the %c[rip] load are
+        // valid before and after the CR3 write.
+        asm volatile
+            ("mov %%rbp, %c[rbp](%0) \n\t"
+             "movq $1f, %c[rip](%0) \n\t"
+             "mov %%rsp, %c[rsp](%0) \n\t"
+             "mov %c[rsp](%1), %%rsp \n\t"
+             "mov %c[rbp](%1), %%rbp \n\t"
+             "mov %2, %%cr3 \n\t"
+             "jmpq *%c[rip](%1) \n\t"
+             "1: \n\t"
+             :
+             : "a"(&old->_state), "c"(&this->_state), "d"(new_cr3),
+               [rsp]"i"(offsetof(thread_state, rsp)),
+               [rbp]"i"(offsetof(thread_state, rbp)),
+               [rip]"i"(offsetof(thread_state, rip))
+             : "rbx", "rsi", "rdi", "r8", "r9",
+               "r10", "r11", "r12", "r13", "r14", "r15", "memory");
+    } else
+#endif
     asm volatile
         ("mov %%rbp, %c[rbp](%0) \n\t"
          "movq $1f, %c[rip](%0) \n\t"
--- a/arch/x64/arch-switch.hh
+++ b/arch/x64/arch-switch.hh
@@ -13,6 +13,7 @@
 #include <osv/kernel_config_preempt.h>
 #include <osv/kernel_config_threads_default_kernel_stack_size.h>
 #include <osv/kernel_config_syscall_stack_size.h>
+#include <osv/kernel_config_fork.h>
 #include <string.h>
 #include "tls-switch.hh"
 
@@ -93,6 +94,16 @@ void thread::switch_to()
     barrier();
     set_fsbase(reinterpret_cast<u64>(_tcb));
     barrier();
+#if CONF_fork
+    // Address-space (CR3) switch for fork COW.  Only touch CR3 when the target
+    // thread lives in a different address space than the outgoing one, so the
+    // common single-address-space case pays nothing (and no TLB flush).  The
+    // kernel half of every AS is identically mapped, so the switch code, kernel
+    // stacks and kernel heap remain valid across the write.
+    if (_current_as != old->_current_as) {
+        processor::write_cr3(mmu::pt_root_phys(_current_as));
+    }
+#endif
     auto c = _detached_state->_cpu;
     old->_state.exception_stack = c->arch.get_exception_stack();
     // save the old thread SYSCALL caller stack pointer in the syscall stack descriptor
--- a/arch/x64/fork.cc
+++ b/arch/x64/fork.cc
@@ -4,17 +4,26 @@
  * This work is open source software, licensed under the terms of the
  * BSD license as described in the LICENSE file in the top-level directory.
  *
- * fork_thread(): create a child thread that resumes in fork()'s CALLER, on a
- * private copy of the parent's user stack, returning 0 from fork() in the child.
- * The x86-64 arch half of the fork() emulation (see documentation/fork.md).
+ * fork_thread(): create a child thread that resumes in fork()'s CALLER on the
+ * SAME user stack virtual addresses the parent had -- no stack relocation, no
+ * SP bias -- and with the caller's FULL callee-saved register context restored,
+ * exactly as a normal `ret` from fork() would leave it.  The child's address
+ * space (built by clone_address_space) maps the parent's live stack VA range to
+ * freshly-allocated PRIVATE physical pages that byte-copy the parent's stack,
+ * so the child sees identical stack contents at identical addresses but writes
+ * to its own private frames.
  *
- * fork() (libc/process/fork.cc) passes us the caller's resume point:
- *   caller_ret  = the address fork() would return to  (__builtin_return_address)
- *   caller_sp   = the parent's SP at fork()'s return   (fork()'s frame base)
- * We copy the parent stack region [caller_sp .. stack_base) into a fresh stack,
- * bias caller_sp into the copy, and start a child thread whose trampoline sets
- * rsp=child_sp, rax=0, and jumps to caller_ret -- i.e. the child returns from
- * fork() with value 0 on its own private stack, in the caller.
+ * Why same-VA + full register restore: OSv/x86-64 code addresses saved frame
+ * pointers (rbp), return addresses, and &local pointers as absolute stack VAs,
+ * and keeps live values in callee-saved registers (rbx, r12-r15) across calls.
+ * Relocating the child stack to a different VA and biasing rsp leaves every
+ * saved rbp/&local pointing at the PARENT's stack (off by the bias), and simply
+ * jmp-ing to the return address skips fork()'s epilogue so the caller resumes
+ * with fork()'s internal register values.  Either corrupts a deep unwind.
+ * Keeping the child on the parent's exact VAs (private phys) AND restoring the
+ * full caller register context (osv::fork_resume_ctx, captured at fork() entry)
+ * makes the child continue in fork()'s caller as if fork() had returned 0 -- the
+ * only correct way to fork a deep stack (see tst-fork-deep, documentation/fork.md).
  */
 
 #include "arch.hh"
@@ -23,13 +32,16 @@
 #include <string.h>
 #include <cstdlib>
 #include <osv/sched.hh>
+#include <osv/fork.hh>
 
 // pthread_atfork child-handler chain (defined in libc/pthread.cc), run in the
 // child's context before it resumes user code.
 extern "C" void __osv_run_atfork_child();
 
-sched::thread *fork_thread(void *caller_ret, void *caller_sp, void **out_stack_to_free)
+sched::thread *fork_thread(void *caller_ret, void *caller_sp,
+                           void *resume_ctx, void **out_stack_to_free)
 {
+    auto ctx = static_cast<osv::fork_resume_ctx*>(resume_ctx);
     auto parent = sched::thread::current();
     auto parent_pinned_cpu = parent->pinned() ? sched::cpu::current() : nullptr;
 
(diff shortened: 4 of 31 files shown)


gburd:
Update: the known deep-call-chain limitation is now **fixed** (the child stack is same-VA COW rather than relocated+copied), so this PR delivers memory-isolated fork() including deep-call-chain children.

Root cause of the deep-unwind fault (found via hardware watchpoint): `thread::switch_to()` loaded the incoming thread's CR3 *too early*. OSv runs the scheduler on the app thread's stack; with same-VA stacks, a CHILD->PARENT switch loaded the parent's CR3 but then executed `fnstcw`/`stmxcsr` scratch saves while rsp still pointed at the child's (same-VA) stack, so those writes landed on the parent's physical stack page and clobbered its live state.

Fixes (all gated CONF_fork): defer the CR3 switch into the switch_to asm coincident with the rsp/rbp swap; privatize the forking thread's stack VA range into the child address space (same VA, private COW pages); restore the child's full callee-saved register context on resume; heap-allocate lock/condvar wait_records while any child address space is live (cross-AS coherence); and, on execve of a fork child, tear down the child address space from the kernel stack so destroy_address_space runs exactly once whether the child exits or execs (no leak, no double-free).

Validated x86-64/KVM: tst-fork 10/10, tst-fork-cow 6/6, **tst-fork-deep PASS**, tst-execve 3/3, clean shutdown across 5x repeat runs and smp1/smp4; conf_fork=0 still builds with all fork code excluded and tst-mmap/tst-pthread passing (no scheduler regression).

gburd:
Added a correctness fix to `clone_address_space` (commit on top of the stack): **preserve `file_vma` type when cloning a child address space on fork**.

Previously the child-AS rebuild recreated every parent VMA as an `anon_vma`. For an anonymous or COW-private mapping that's fine, but a **file-backed** mapping (`file_vma`, e.g. an ELF segment mapped by the loader via `map_file`) would become anonymous in the child. A demand fault on such a page in the child then dispatched to the base zero-fill path instead of `file_vma::fault`, so the child read zeros where file contents should be (for an executable text segment, it would execute zeros → wild branch). This doesn't bite a purely-anonymous fork workload, which is why the existing tests passed, but it's a latent correctness bug for any fork child that demand-faults a file-backed page it inherited.

Fix (entirely inside `#if CONF_fork`):
- In the rebuild, `dynamic_cast<file_vma*>` each parent VMA; a file-backed one is recreated via `fv->file()->mmap(range, flags, perm, offset)` (the same call `map_file`/`file_vma::split` use, so the child gets the correct page allocator and its faults read the file); anonymous VMAs still become `anon_vma`.
- `destroy_address_space` now disposes the child's owned VMA objects so `~file_vma` releases the fileref the child took (no leak).

New test `tst-fork-file-mmap`: the child reads a page of a MAP_PRIVATE file mapping the parent hadn't faulted in — without the fix it reads `0x00` (fails), with it reads the real bytes. `tst-fork`, `tst-fork-cow`, `tst-fork-deep`, `tst-execve` remain 0 failures; `conf_fork=0` is byte-identical (the whole change is under `#if CONF_fork`).

gburd:
Pushed `e9a31b1ab` here: the fork page-table clone now uses a sorted, coalesced range lookup with a last-match memo instead of two linear scans.

### What was wrong

`addr_is_shared()` and `addr_is_privatize()` are called **once per leaf PTE** by `clone_pt_level0`, and each was a linear scan of its `cow_range` vector. `cow_share_ranges` holds one entry per MAP_SHARED/stack vma **plus one per live thread** (the `with_all_threads` loop in `clone_address_space`). OSv pins roughly two threads per CPU (`idle_thread`, `page_pool_l1_<id>`) and a server application adds more per connection, so the scan length grows with both vCPU count and application concurrency, making the clone walk O(PTEs x threads).

On a real PostgreSQL fork the shape was 2,300,347 leaf PTEs against a 695-entry range vector: about 800M comparisons per fork.

### Where the cost actually was

An rdtsc phase breakdown of `clone_address_space` at 32 vCPUs, which is worth recording because it refutes the obvious suspect:

| phase | ms | share |
|---|---|---|
| ptwalk | 45.5 | **99.8%** |
| ranges setup | 0.06 | 0.1% |
| `flush_tlb_all` | 0.0013 | **0.003%** |

`flush_tlb_all` is 1.3 microseconds. The TLB broadcast is not the problem; the per-PTE predicate is.

### Effect of the fix

A/B in one image, `K` concurrent forks, only the lookup strategy differing:

| K concurrent forks | 1 | 2 | 4 | 8 | 16 | 32 |
|---|---|---|---|---|---|---|
| linear | 146 | 286 | 552 | 1179 | 2322 | 4631 ms |
| sorted+memo | 79 | 145 | 279 | 550 | 1172 | 2094 ms |

2.21x fork throughput (6.9 -> 15.3 forks/s), 144.7 -> 65.4 ms per fork.

vCPU sensitivity, which is the discriminator for the diagnosis:

| smp | 2 | 8 | 32 | spread |
|---|---|---|---|---|
| linear | 75 | 89 | 138 ms | +84% |
| sorted+memo | 62 | 62 | 69 ms | **+11%** |

And fork cost against live backend count at fixed vCPUs, which is the case that matters for a forking server:

| backends | n_threads | range_len | linear | sorted+memo |
|---|---|---|---|---|
| 8 | 667 | 701 | 150 ms | **73 ms** |
| 64 | 721 | 755 | 161 ms | **75 ms** |
| 256 | 913 | 947 | **191 ms** | **79 ms** |

The linear scan gets more expensive the more threads exist (+27% from 8 to 256 backends); the sorted lookup is nearly flat (+8%).

**Provenance, stated plainly:** those four tables were measured on a fuller fork stack that also carries a fork arena and an rdtsc instrumentation harness not present on this branch. This commit ports **only** the lookup change. The numbers are the motivation for it and have not been re-measured on this branch, so please read them as motivation rather than as a measurement of `e9a31b1ab` itself.

### Correctness

Only **how** the "is this VA in the set?" predicate is computed changes. Not which VAs are selected, not any PTE value, not any lock, not any flush semantics. Sorted, coalesced, half-open `[start, end)` intervals give an identical predicate, because the union of the intervals is exactly what the linear scan tested. The memo is a pure-function cache of the last matching range: a hit returns what the search would have returned, and it is invalidated whenever the range sets are rebuilt.

The sorted views have the same scope and lifetime as the `cow_share_ranges` / `cow_privatize_ranges` pointers they cache: written only between publishing and clearing those pointers, under the parent's `vmas_mutex` for_write. Nothing mutates the underlying vectors in that window, so rebuilding at publish time is the only invalidation point needed. Nothing reads the original vectors positionally, so sorting into separate vectors changes no other consumer.

### Validation

**Predicate equivalence, host-side.** The new lookup was extracted verbatim and differentially tested against the original linear scan over 3000 randomly generated range sets, including empty, zero-length, duplicate, adjacent and overlapping ranges (the cases where a coalescing rewrite breaks). Every VA in range was checked in both ascending order, which is the clone walk's real access pattern and the one the memo is tuned for, and in random order, so a stale memo cannot be hidden by monotonic access. Result: **1,560,000 lookups, identical to the linear scan at every VA**, with the coalesced view strictly ordered and disjoint. This is stronger evidence for the predicate than a boot gives, since a boot only exercises the ranges that one run happens to produce.

**Guest boot and fork tests**, x86_64 release, `conf_fork=1`, against the unpatched parent commit `e23e2829c`:

| test | unpatched | with this commit |
|---|---|---|
| tst-fork | 9 PASS / 1 FAIL | 9 PASS / 1 FAIL |
| tst-fork-cow | 6 PASS / 0 FAIL | 6 PASS / 0 FAIL |
| tst-fork-deep | 3 PASS / 0 FAIL | 3 PASS / 0 FAIL |
| tst-fork-file-mmap | 0 PASS / 1 FAIL | 0 PASS / 1 FAIL |

Byte-identical between arms. Both remaining failures are artifacts of the reduced test manifest I built (a missing exec payload and a missing `/rofs` fixture) and reproduce identically on the unpatched build, so neither is a regression.

`tst-fork-cow` is the direct exercise of the changed code: it asserts that a COW private page stays private to the parent and that a MAP_SHARED write in the child is visible in the parent. That is precisely the predicate being modified, and it passes with zero failures.

gburd:
## Defect: the ELF object cache is process-global while PT_LOAD mappings are per-address-space

Found while benchmarking PostgreSQL on this branch. It is a defect in the per-child address
space design rather than in pre-existing code, so I am reporting it here instead of opening a
PR against master (which has no `fork()`).

**Reproduces on both x86_64 and aarch64, with no PostgreSQL involved, in under a second after
boot.**

### The invariant that is violated

`elf::program`'s object cache (`_files`) and module list are **process-global**, but an
object's PT_LOAD segments are mapped into whichever **address space** was current when it was
first loaded: `file::load_segment` calls `mmu::map_file`, which inserts the vma into
`cur_vma_list()`, i.e. the current address space's own `vma_list`.

Under `conf_fork=1` those two disagree by construction. `program_base` (`0x100000000000`) is
application PML4 slot 32, and `clone_address_space` deep-copies slots 1..127 per child, giving
each child a private `vma_list` and private page-table subtree for exactly the region shared
objects live in.

So when an address space `dlopen()`s a library that a **different** address space loaded
earlier, `program::load_object` returns the cached object straight from `_files` and maps
nothing. The object's `_base`, `_dynamic_table` and symbol addresses all point at virtual
addresses that are unmapped in the caller. The first dereference dies:

```
page fault outside application, addr: 0x000010000008f000
PC: <elf::object::dynamic_tag(unsigned int)+8>
mmu::vm_fault -> page_fault -> elf::object::lookup_symbol_deep -> dlsym
ESR: 0x0000000096000007
```

Two properties of that message are worth spelling out, because both are diagnostic:

* The address is **page-aligned** because `vm_fault` aligns down *before* the vma lookup, and
  `vm_sigsegv` prints the aligned value. `fast_sigsegv_check` returns `false` on aarch64, and
  `ESR=0x96000007` has bit 6 clear so it is a **read**, meaning `access_fault()` cannot be
  true for a readable segment. Both alternative routes to that message are therefore excluded:
  it means literally `find_intersecting_vma_in(*as->vmas, addr) == end()`, i.e. **no vma
  covers the address in the faulting address space**.
* The address is **identical on every occurrence** because the object's `_base` is fixed at
  first load, so every later caller is a cache hit on the same object. This is why the failure
  looks deterministic rather than race-shaped, and it is the detail that rules out the
  concurrency and cache-reclaim theories we initially pursued: the first `dlopen` of a library
  succeeds and every subsequent one from another address space fails at exactly the same
  address, at one connection, with no concurrency at all.

### Measured, not inferred

Instrumenting the fault slow path itself at the `find_intersecting_vma_in(...) == end()`
branch, on aarch64:

```
=== round 0: fork
OSVDIAG loadseg /libplug.so base=0x100000050000 va=[0x100000050000,0x100000051000) as=0x6000ffe35580 as0=0
OSVDIAG loadseg /libplug.so base=0x100000050000 va=[0x10000008f000,0x100000091000) as=0x6000ffe35580 as0=0
[plug] constructor ran
[child0] RESULT plug_answer()=42 (expect 42)          <- works in the AS that mapped it
=== round 1: fork
OSVDIAG cachehit /libplug.so base=0x100000050000 as=0x600041116300 as0=0   <- maps NOTHING
OSVDIAG novma addr=0x000010000008f000 as=0x600041116300 as0=0 err=0x96000007
OSVDIAG  prev [0x100000040000,0x100000041000) perm=3 fl=0xa3
OSVDIAG  NOT-in-AS0 either
page fault outside application, addr: 0x000010000008f000
```

Same `_base` in both address spaces, mapped in the loader and absent from the cache-hit
address space, and absent from AS0 as well.

The same mechanism measured independently on **x86_64** via `mincore()` (which consults
`ismapped()` -> `find_intersecting_vmas()` -> `cur_vma_list()`, the same per-AS list the fault
path searches):

```
xas/child0-after-dlopen: base=0x10000002a000 PT_DYNAMIC=0x10000002edd8 VMA-in-this-AS=YES
xas/child0: dlopen+dlsym OK in the AS that mapped it, fn_0(7)=7
xas/child1:              base=0x10000002a000 PT_DYNAMIC=0x10000002edd8 VMA-in-this-AS=NO
page fault outside application, addr: 0x000010000002e000
RIP: <elf::object::dynamic_tag(unsigned int)+12>  RAX: 0x000010000002edd8
```

`RAX`/`X00` is `_dynamic_table`, the walk cursor in
`for (auto p = _dynamic_table; p->d_tag != DT_NULL; ++p)`.

Controls, on x86_64:

* A **probe-only** mode that reports the identical `VMA-in-this-AS=NO` state but never calls
  the faulting `dlsym` does **not** fault, isolating the fault to the `_dynamic_table`
  dereference itself.
* The first loader always survives its own `dlopen`+`dlsym`+call, so the library is fine; only
  the cross-address-space cache hit is broken.
* Page-cache reclaim is excluded structurally, not by argument: the image is `fs=rofs` with no
  ZFS, and ROFS sets `vop_cache = nullptr`, so library pages are private anonymous pages via
  `map_file_page_read` and are never in any page cache.

**Both directions fail**, so a fix must not special-case the forked child:

* a later forked child is the second consumer (the PostgreSQL backend shape), and
* **AS0 itself is the second consumer** when a forked child loaded the library first.

### Why this matters beyond PostgreSQL

Any program that forks and then `dlopen()`s the same library in more than one address space
hits this. The PostgreSQL symptom was that every `plpgsql` call faulted at one connection,
which made all six HammerDB TPROC-C stored procedures unusable on this branch;
`shared_preload_libraries=plpgsql` does not work around it, it just moves the failure earlier
into postmaster startup by loading the library pre-fork.

### Reproducer

`tests/tst-fork-dlopen-as.cc` plus `tests/libdlopenas.cc` on the branch below. It forks
repeatedly, and in each child `dlopen`s, `dlsym`s and calls into the same library, then repeats
with the parent as the second consumer. The library's function reads a writable global that a
constructor sets, so the test also catches the weaker failure where the segments are mapped but
relocations or constructors were not re-established for the new address space.

Before the fix, aarch64:

```
PASS: round 0: child dlopen + dlsym + call (status 0)
page fault outside application, addr: 0x00001000000ef000
PC: <elf::object::dynamic_tag(unsigned int)+8>
X07: 0x6e65706f6c646269  X12: 0x6f6c6462696c2f73  X13: 0x6f732e73616e6570
```

(those registers spell `ibdlopen` / `s/libdlo` / `penas.so`, the same register signature the
production `plpgsql` fault had.)

After:

```
PASS: round 0..11: child dlopen + dlsym + call (status 0)
PASS: parent (the address space that did not load it) dlopen + dlsym + call
tst-fork-dlopen-as: 0 failures
```

### Proposed fix

Branch: `fix/elf-object-cache-address-space` on https://github.com/gburd/osv-1

On a `_files` cache hit, ask whether the object is mapped in the **calling** address space and,
if not, re-establish it there at its existing base. Re-mapping the segments alone is not
enough, and this cost me a build to learn: a writable segment is mapped `MAP_PRIVATE` from the
file, so it comes back with its **unrelocated** on-disk contents (zero GOT entries), and the
next call through the GOT faults at address 0 immediately after a *successful* `dlsym`. So the
remap also re-runs the object's relocations, re-applies the RELRO and text protections, and
runs its constructors for the new address space. Relocations are recorded relative to `_base`,
which does not change, so they rebuild a byte-identical GOT.

Two further details that are easy to get wrong:

* The per-object fork-shared range registration in `load_segment` must become one-shot. A
  remap covers virtual addresses that are already registered, and re-registering grows the
  process-global vector without bound while its realloc/free lands on the **calling** address
  space's fork arena rather than the shared kernel heap. Without this guard the test aborts at
  `fork_arena.cc: recover: h->magic == chunk_magic`.
* Objects with no backing file (the kernel's own `memory_image`) must be excluded; they are
  mapped by the boot path in every address space and cannot be re-read. Without this,
  `memory_image::read` throws `cannot load from Elf memory image` at boot.

**Options considered.** Making `_files`/`_modules` fully per-address-space is the most faithful
model, but it changes a process-global structure that `dl_iterate_phdr`, the unwinder and
`program::lookup` all read cross-AS, and it would give each address space a different `_base`
for the same library, which breaks function pointers passed between them. Having
`clone_address_space` also clone segment mappings for the slots it copies fixes only the
fork-child direction and does nothing for the AS0-as-second-consumer case that x86_64
demonstrates, and it makes every fork pay for libraries the child may never touch. Refusing the
cache hit and loading a second copy would give the same library two bases in one process and
break `RTLD_GLOBAL` symbol identity. Re-mapping lazily on the cache hit keeps one object, one
base and one set of symbol addresses process-wide, and costs nothing for address spaces that
never touch the library.

**Memory cost, stated explicitly.** Read-only and executable segments cost no additional
physical memory: the mapping is file-backed at the same file offsets, so the same pages are
used. Writable segments (`.data`/`.got`/`.bss`) do become a private per-address-space copy,
which is the semantically correct outcome for a forked process anyway (library globals are
per-process), and is the same cost a normal `dlopen` in that address space would have incurred.
For a PostgreSQL backend loading `plpgsql.so` that is a few pages per backend.

### Caveat on the fix

I have validated the mechanism, the fix and the regression test on aarch64 (Graviton3, KVM,
`conf_fork=1 fs=rofs`), and the fix compiles for both architectures. The x86_64 evidence above
is the *bug*, reproduced independently; I have not yet run the *fix* on x86_64. Given that the
fix is architecture-neutral and the x86_64 reproducer is standalone, that should be quick to
confirm, but it is untested there and I would rather say so than imply otherwise.

gburd:
One scope caveat on the branch above that I should have stated in the previous comment.

**The fix is validated at runtime on aarch64 only.** The pre-fix fault and the post-fix pass (rounds 0..11 plus the AS0-as-second-consumer case, 0 failures) were both produced on a Graviton3 guest. The x86_64 side of this report is the *mechanism* confirmation, measured independently with `mincore()` and its own reproducer, not a validation of the patch: the patched tree has not been booted on x86_64.

There is no reason to expect an ISA-specific problem here, since the change is in `core/elf.cc` and the address-space bookkeeping rather than in any arch layer, and the mechanism reproduces identically on both. But "confirmed on both ISAs" applies to the bug, and "validated" applies only to aarch64. An x86_64 boot plus the committed test is the outstanding item before anyone treats this as verified on that ISA.

The committed test provably fails without the fix, so it is a real gate rather than a smoke test, and running it on x86_64 is the cheap way to close this out.

gburd:
Lifting the aarch64-only caveat: **the fix is now runtime-validated on x86_64 as well.**

Committed test `tests/tst-fork-dlopen-as.cc`, same setup both directions:

Pre-fix, round 0 passes and the first cross-AS cache hit faults with the production signature:

```
page fault outside application, addr: 0x10000002d000
RIP: <elf::object::dynamic_tag(unsigned int)+12>
dlsym+268 -> elf::object::lookup_symbol_deep+29
```

Post-fix, rounds 0..11 PASS plus `PASS: parent (the address space that did not load it)`, 0 failures. That last line is the AS0-as-second-consumer case, which a fork-child-only fix would leave broken, so it is the one worth checking.

So both the **bug** and the **patch** are confirmed on x86_64 and aarch64.

Two further data points from putting the fixed tree under real load, which bear on this PR:

**A second instance of the same rule, in the rofs cache.** With dlopen working, a TPROC-C sweep reached VU32 and hit:

```
Assertion failed: h->magic == chunk_magic (core/fork_arena.cc: recover: 257)
  fork_arena::free+120 -> std::_Hashtable<rofs::rofs_cache_key, ...>::_M_rehash+168
  -> rofs::cache_read+28 -> vfs_file::read+104 -> __uflow -> fgets -> getnameinfo+336
```

`fs/rofs/rofs_cache.cc:152` holds a process-global `unordered_map` populated lazily by whichever address space reads a file first, so under fork those nodes land in that child's COW arena and are garbage elsewhere. The trigger is ordinary: every backend calls `getnameinfo()` to resolve the client address, which `fgets`es files off the rofs root, so dozens of address spaces grow and rehash one shared cache. This is the **third** instance of the rule already documented in `core/epoll.cc:63-81` ("the same rule as struct file, f_epolls and thread objects"). Fixed with one `kernel_heap_scope` at `cache_read()`, the single allocation entry point: branch `fix/rofs-cache-fork-arena-coherence`.

**A fourth candidate, not yet fixed.** `core/pagecache.cc:361-364` declares `arc_read_cache`, `read_cache` and `write_cache` as process-global maps with **zero** `kernel_heap_scope` in the file. The ARC one is unreachable under `conf_zfs=openzfs`, but that file's own comment at line 31 says ROFS and the OpenZFS port route reads through `read_cache`, which is used unconditionally at `:413`/`:415`/`:470`. Same shape, hotter path.

Both of those need `conf_fork=1`, so they are reportable here rather than against master.

gburd:
## Fixed a latent `conf_fork=0` build break in the elf address-space fix

The elf object-cache/address-space fix reported on this PR carried a
conditional-compilation defect. Recording it here because this is where that work
is tracked, and because the break would have landed on merge rather than on the
branch where it was written.

`include/osv/elf.hh` declared the base `object::has_backing_file()` **inside** the
`#if CONF_fork` block, but the `class file` override **outside** it. With
`conf_fork=0`, which is the default since `conf/base.mk` does not set `conf_fork`,
the base declaration is not compiled and the derived one is left marked `override`
with nothing to override:

    include/osv/elf.hh:548:18: error: 'virtual bool
    elf::file::has_backing_file() const' marked 'override', but does not override

Nothing published was affected: the hunk is not on this PR's head branch
(`wip/fork-stage2-pr`) and `da3041f5d` is not an ancestor of it, so no open PR
built the broken header. It lived only on the three private branches carrying the
elf fix, all now corrected by a two-line commit that moves the override inside the
same guard.

**Why it survived: the branch has only ever been built with `conf_fork=1`.** In
that configuration both declarations exist and the header is well-formed, so the
break is invisible to every check we routinely run. It passes review, since the
diff is self-consistent. It passes a mergeability check. It passes the tests we
run, because we never ran the configuration that fails.

Verified both directions rather than assuming, by compiling the real `core/elf.cc`
translation unit through the project's own build (not a reduced test case):

| header state | `conf_fork=0` | `conf_fork=1` |
|---|---|---|
| override unguarded (as written) | **fails, line 548** | compiles OK |
| override inside `#if CONF_fork` | **compiles OK** | **compiles OK** |

The middle cell is the one that matters: it confirms the fix does not simply
silence the error at the cost of the fork build, which would have been worse than
the bug.

Recommendation: exercise `conf_fork=0` as a build configuration before any of the
fork-related branches merge. A `conf_fork=0` compile of the affected TUs is cheap
and is the only check that would have caught this.

gburd:
## Fork is the PostgreSQL gap on OSv, and the root cause was one line in the page-table clone

Hypothesis **confirmed**, then the residual was found and fixed too. Connect median on aarch64 went **114.8 ms -> 3.98 ms**, which is at or slightly better than the Linux reference (4.0-4.3 ms). Two independent causes, both in `clone_address_space()`, both attributed by same-binary A/B.

Branch: **`gh-fork wip/forkroot-arm-base`** (head `5a8e946f0`). Not master-eligible: needs `conf_fork=1`, and `upstream/master` has no `fork()`.

### Identity tuple

| item | value |
|---|---|
| osv SHA | `5a8e946f04329583d55d17460159b5eb291a9b54` |
| apps SHA | `b38b99cd9302af01fb388e569bc0911c0e90bbc4` |
| ZFS submodule SHA | `6330a45b06d20125de679aae5f63ba14082671ef` |
| PG branch / version | `REL_18_STABLE` / PostgreSQL **18.6** |
| configure | `--prefix=... --without-icu --without-zlib --without-readline --without-libxml --without-lz4 --without-zstd CFLAGS="-O2 -g -fPIC -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0 -DWAIT_USE_SELF_PIPE -mno-outline-atomics -idirafter /usr/include"`, `CC=musl-gcc`, `LDFLAGS_EX=-pie` |
| build | `conf_fork=1 conf_zfs=openzfs fs=rofs image=zfs,zfs-tools,postgres18-musl` |
| box | `c7gd.metal` aarch64, 64 vCPU, guest `-smp 32 -m 64G`, qemu 8.1.3 in fedora:39 |

Image proof (build-time, not the banner): **298** `modules/open_zfs` objects vs **0** `bsd/.../fs/zfs` objects; all four OpenZFS-only libs present (`libzfs_core/libzutil/libshare/libtpool`); `strings disk.img` finds `PostgreSQL 18.6` and the forkbench strings, so this is not one of the PostgreSQL-free images.

### Method: same binary, env-var A/B

Every arm below is **one image, one binary, one dataset**; only an OSv `--env=` differs. Build variance is eliminated as a confound rather than argued about. Arms are **interleaved** (BS1,BS0,BS1,BS0...) so drift cannot be attributed to the switch, and each boot prints `FORKFLAG <name>=<v>` as proof the flag bound. **50/50 fork cells and every PG arm proven bound, 0 unproven.** That print earned its keep immediately: my first sweep had no FORKFLAG lines at all because I had edited `mmu.cc` without committing it, so the "A/B" was running the same code twice.

### 1. Fork cost, no PostgreSQL involved (`misc-forkbench`)

p50 ms/fork, median of 3 reps, 40 forks each:

| cell | shared | thr | window | BS=0 linear | BS=1 fixed | speedup | forks/s |
|---|---|---|---|---|---|---|---|
| base | 256M | 0 | 1 | 38.95 | **2.93** | 13.3x | 25.7 -> 342 |
| sh1g | 1G | 0 | 1 | 40.57 | **4.49** | 9.0x | 24.6 -> 214 |
| thr32 | 256M | 32 | 1 | 42.66 | **2.91** | 14.7x | 23.4 -> 333 |
| thr128 | 256M | 128 | 1 | 55.09 | **3.16** | 17.4x | 18.1 -> 307 |
| sh1g_thr128 | 1G | 128 | 1 | 56.89 | **4.86** | 11.7x | 17.6 -> 199 |
| win8 | 256M | 0 | 8 | 38.95 | **2.94** | 13.3x | 25.7 -> 338 |
| win32 | 256M | 0 | 32 | 38.97 | **2.81** | 13.9x | 25.6 -> 347 |
| sh1g_win32 | 1G | 0 | 32 | 40.53 | **4.84** | 8.4x | 24.6 -> 213 |

Reads: the linear scan cost grows with **thread count** (38.9 -> 55.1 ms from 0 to 128 threads) and the fix removes that sensitivity (2.93 -> 3.16). Cost does **not** grow with the number of live children (`win1` vs `win32` identical in both arms), which refutes the "cost scales with backend count" reading of the older data; what actually scaled was resident PTEs and threads.

### 2. Connect latency against real PostgreSQL

`OSV_FORK_RANGE_BS` A/B, 3 reps, interleaved, `shared_buffers=16GB`. Phases timed separately so the claim is falsifiable:

| conc | arm | total p50 | **startup p50** | tcp p50 | query p50 | p99 | >=1s |
|---|---|---|---|---|---|---|---|
| 1 | BS=0 | 114.81 | 114.43 | 0.07 | 0.27 | 131.22 | 0 |
| 1 | **BS=1** | **42.57** | 42.22 | 0.09 | 0.27 | 60.77 | 1 |
| 8 | BS=0 | 919.96 | 919.59 | 0.07 | 0.29 | 2949.57 | 214 |
| 8 | **BS=1** | **347.46** | 346.81 | 0.07 | 0.29 | 386.82 | 0 |
| 32 | BS=0 | 3899.35 | 3898.81 | 0.09 | 0.29 | 7491.47 | 3817 |
| 32 | **BS=1** | **1412.88** | 1412.52 | 0.08 | 0.29 | 3355.28 | 3773 |

**Your prediction, scored:**
- "connect median falls from ~100 ms" - **YES**, 114.8 -> 42.6 ms at conc=1. Your 100 ms figure reproduced (114.8).
- "toward a 40-65 ms residual floor" - **YES, exactly**: 42.6 ms, inside your predicted band.
- "improves most at high VU" - **NO.** The ratio is flat at 2.65-2.76x across conc 1/8/32. It is a constant-factor win on a per-fork cost, not a concurrency unlock.
- Mechanism confirmed hard: **all** of the movement is in `startup` (the phase containing the fork). `tcp` and `query` are unchanged to two decimals in every cell. Connect latency really is fork latency.

### 3. The residual, characterised and then fixed

`OSV_FORK_TIMING=1` on real PG backends, BS=1:

```
tot_kc  58,189   ptwalk_kc 58,123 (99.9%)   tblalloc_kc 1,820 (3.1%)
ranges_kc 48     snap_kc 3    flush_kc 0    vmareb_kc 10
n_ptes 4,445,585   n_lookups 4,445,329   n_tables 8,691
```

So the answer to "where does the remaining time go" is not TLB work, not the vma_snap rebuild, not intermediate table allocation: it is **99.9% the leaf-PTE walk**, 4.44M PTEs at ~13 cycles each. `flush_tlb_all()` measured **zero**.

`shared_buffers=16GB` is 4.19M 4K pages. Sweeping it (BS=1, fix on):

| shared_buffers | 4K pages | connect p50 | startup p50 |
|---|---|---|---|
| 16GB | 4,194,304 | 42.73 | 42.36 |
| 8GB | 2,097,152 | 23.38 | 23.03 |
| 4GB | 1,048,576 | 13.74 | 13.42 |
| 2GB | 524,288 | 8.80 | 8.51 |
| 1GB | 262,144 | 6.40 | 6.08 |

Fit `startup_ms ~ 3.66 + 9.23 ms per 1M PTEs`, **max error 0.6%** across a 16x range. The residual was not mysterious, it was `shared_buffers` walked one PTE at a time.

**Root cause of the residual:** `shared_buffers` is backed by 2 MB large pages, and `clone_pt_level<1>` **split every writable large page** into 512 4K PTEs, then decided each individually. For a MAP_SHARED range all 512 decisions reach the same branch and produce "writable, shared in both" - which is what the large PTE already said. The split allocated a page table, did 512 PTE writes and 512 range lookups per 2 MB, 8192 times per 16 GB, to reproduce the entry it started from.

**Fix** (`5a8e946f0`, gated `OSV_FORK_LARGE_SHARE`, default on): if the whole 2 MB is inside one coalesced share range and clear of every privatize range (thread stacks), share the large PTE verbatim. Teardown was already correct - `free_child_pt_level<1>` skips `large()` entries, so a dying child never frees the jointly-owned frame.

Measured mechanism, same binary: `n_ptes` **4,445,585 -> 132,497** (33x fewer), `n_lg_shared` **8,423** large pages shared, `ptwalk_kc` **58,123 -> 2,615**.

| conc | LARGE_SHARE=0 | **=1** | speedup | >=1s (0 -> 1) |
|---|---|---|---|---|
| 1 | 46.08 | **4.06** | 11.4x | 0 -> 0 |
| 8 | 378.17 | **18.11** | 20.9x | 1 -> 0 |
| 32 | 1538.40 | **126.30** | 12.2x | 1559 -> **96** |

### Both fixes together vs the briefed baseline

| | connect p50 |
|---|---|
| linear scan + split (the benchmarked trees) | **117.3 ms** |
| both fixes | **3.98 ms** (p90 4.04) |
| Linux/PG18.6 reference | 4.0-4.3 ms |

**29x, and the ~100 ms aarch64 connect latency is gone.** The 16x-worse-than-Linux residual you expected to remain does not remain; it was the large-page split.

### Correctness

All 14 fork tests, **both** arms of the large-page path, 28/28 pass, 0 crashes - including `tst-fork-shared-write`, `tst-fork-posix-shm`, `tst-fork-shared-catalog`, `tst-fork-cow`, `tst-pgfork`, i.e. exactly the tests that fail if a page is shared that should have diverged.

### Corrections to the brief

1. **`wip/fork-clone-scan` is 8 commits ahead of the base, not 2.** It also carries `core/mempool.cc` +62 (OOM retry-reclaim and the largest-free-range report), `drivers/virtio-blk.cc` +24, `include/osv/mempool.hh` +5. Any comparison to previously benchmarked trees is **8 commits**, not 2; only the env A/B attributes the scan fix.
2. **`git merge-tree` "0 conflicts" proves mergeability, not scope.** `git log base..branch` is the scope check.
3. **"cost scaling with connection count" is not reproducible.** Window 1 vs 32 is identical in both arms. The real axes are resident PTEs and live threads.
4. **Fork was 65-150 ms/fork before; it is now ~2.9 ms** in the microbenchmark and the PG backend fork is ~3.7 ms of the 3.98 ms connect.

### Not reproduced (so not investigated)

Neither open bug appeared in ~40 guest boots: **no silent hang** after `checkpoint starting`, and **no OOM abort** - so no gdb backtrace is owed from this run, because there was nothing wedged to attach to. The OOM is plausibly already addressed by `866bc8b6a` (retry reclaim across shrink passes), which was in the scan branch and in neither benchmarked tree. That is the second case today of a fix existing but not being where it was needed.

### Caveats

- HammerDB NOPM not run; the prediction that NOPM improves most at high VU is **untested**. Given the connect result is a flat-ratio per-fork win, I would expect NOPM to gain wherever connection churn matters and little in a steady-state VU8 run with persistent connections.
- aarch64 only. x86 untested, but the code path is shared.
- `>=1s` probes at conc=32 drop to 96 but are not zero, so something else still stalls under heavy concurrent connect. That is the next thread to pull.

gburd:
## x86 port of the large-page fork fix: measured, and the aarch64 result reproduces

Ported the missing half of the fork/connect fix to the x86 lineage and measured
the connect probe A/B on bare metal. Summary: the fix is **not** ISA-specific.
The mechanism and its magnitude reproduce on x86 almost exactly.

### Identity tuple

| item | value |
|---|---|
| osv branch | `wip/pg-x86-forkroot-largeshare` |
| osv SHA | `42aeb163b3b3a2b2660f68464f0e36d89bd7777f` |
| apps SHA | `5cb19033555d6d4370da905965016f0adb81e515` |
| ZFS submodule | `6330a45b06d20125de679aae5f63ba14082671ef` (`zfs-2.4.2`) |
| PG branch / version | `REL_18_STABLE` / PostgreSQL 18.6 |
| configure | `--prefix=... --without-icu --without-zlib --without-readline --without-libxml --without-lz4 --without-zstd CFLAGS="-O2 -g -fPIC -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0 -DWAIT_USE_SELF_PIPE -idirafter /usr/include"`, `CC=musl-gcc` |
| build line | `./scripts/build -j16 conf_fork=1 conf_zfs=openzfs fs=rofs image=zfs,zfs-tools,postgres18-musl` |
| host | `m5zn.metal`, Xeon Platinum 8252C @ 3.8 GHz (4.5 GHz max), 48 vCPU, KVM, qemu 8.1.3 `-M pc` |
| guest | 32 vCPU / 64 GiB hugepages, `--env=OSV_FORK_LARGE_SHARE=<n>` |

Branch is the four requested commits plus the three elf fixes, cherry-picked
onto `wip/pg-v26-faultfix-repinned` + a merge of `wip/fork-clone-scan`. 7/7
applied with no conflicts. All 11 new commits are `-S` signed, `%G?` == `G`.

Both flag halves are now present on x86, where before only one was:

```
branch                            OSV_FORK_RANGE_BS  OSV_FORK_LARGE_SHARE  FORKFLAG
wip/pg-v26-faultfix-repinned              0                   0               0
wip/fork-clone-scan                       2                   0               0
wip/pg-x86-forkroot-largeshare            2                   2               1   <- this
```

### Connect probe A/B

Same binary both arms, arms interleaved, fresh boot per arm, 8 boot pairs.
Each sample is one `PQconnectdb` + `select 1` + `PQfinish` over TCP from the
host, i.e. exactly one postmaster `fork()`. Flag binding is proved per boot by
the `FORKFLAG` line from `16149a08b`.

```
arm                             n    median       min       max       p90
OSV_FORK_LARGE_SHARE=1         53     6.607 ms   5.615     9.121     7.358
OSV_FORK_LARGE_SHARE=0         11    56.870 ms  55.785    58.212    57.316

speedup 8.61x   (56.87 -> 6.61 ms, delta 50.26 ms)
```

Per-boot detail, showing the two arms never overlap:

```
arm             flag  n   median    min      max      spread%
rep1_flag1       1    18   5.936    5.615    6.841    20.7
rep1_flag0       0     1  55.888   55.888   55.888     0.0
rep2_flag1       1    19   6.716    5.826    7.814    29.6
rep2_flag0       0     0  (no surviving sample)
rep3_flag1       1     2   7.254    6.996    7.511     7.1
rep3_flag0       0     5  56.808   55.785   58.212     4.3
rep4_flag1       1     2   6.992    6.431    7.553    16.0
rep4_flag0       0     1  57.705   57.705   57.705     0.0
rep5_flag1       1     2   6.732    6.558    6.907     5.2
rep5_flag0       0     1  56.929   56.929   56.929     0.0
rep6_flag1       1     1   9.121    9.121    9.121     0.0
rep6_flag0       0     0  (no surviving sample)
rep7_flag1       1     6   6.723    6.454    7.625    17.4
rep7_flag0       0     2  56.638   56.406   56.870     0.8
rep8_flag1       1     3   7.192    6.797    7.358     7.8
rep8_flag0       0     1  57.316   57.316   57.316     0.0
```

### The mechanism, measured directly

With `OSV_FORK_TIMING=1`, the instrumentation prints the PTE count and cycle
cost of each fork. This is the same counter reported for aarch64:

```
                     n_ptes cloned per fork      fork cost
OSV_FORK_LARGE_SHARE=0    4,445,538  (0x43d562)   186,027 kc
OSV_FORK_LARGE_SHARE=1      132,962  (0x20762)     12,343 kc
                          33.4x fewer            15.1x fewer
```

Against the previously reported aarch64 figures:

```
            n_ptes before -> after      ratio
aarch64     4,445,585 -> 132,497        33.6x
x86         4,445,538 -> 132,962        33.4x
```

The two architectures agree to within 0.1% on both endpoints. The fix does
what it says on x86: writable `MAP_SHARED` 2 MB large pages backing
`shared_buffers` are shared verbatim across `fork()` instead of being split
into 512 4K PTEs each.

`5a8e946f0` is 76 lines in one file (`core/mmu.cc`) with zero arch guards,
which is consistent with the result being arch-neutral.

### Caveat on absolute numbers

The `=1` median here is 6.61 ms, not the ~4 ms measured on aarch64, and the
`=0` median is 56.9 ms rather than the previously reported x86 78 ms. Two
differences are relevant and I have not separated them:

- this run is on `m5zn.metal` (8252C, 3.8 GHz base / 4.5 GHz max), a
  higher-clocked part than the `m5d`-class hardware used for the earlier x86
  and Linux numbers, which flatters OSv on per-core work;
- the probe here is a libpq `connect + select 1 + close` from the host across
  a bridged tap, so it includes a TCP round trip that a purely in-guest fork
  timing would not.

The **ratio** is the load-bearing result, and the `n_ptes` counters above are
hardware-independent.

### Open bug found while measuring: fork-count-bounded stack overflow

The guest reliably dies after a small number of cumulative `fork()` calls:

```
__stack_chk_fail(): Stack overflow detected. Aborting.
```

Characterised on this build:

- **Not caused by this port.** Both A/B arms crash identically, 6/6 boots,
  1 to 5 serial forks before death, with `OSV_FORK_LARGE_SHARE` on *and* off.
  The flag changes fork *cost*, not crash behaviour.
- **Not query work.** A single held connection ran 500k inserts, a full-table
  update, a delete of 71k rows, `VACUUM ANALYZE` and `CHECKPOINT` with zero
  crashes. The very next new connection then killed it.
- **Not concurrency per se.** From a fresh boot, 8/8, 16/16, 32/32 and 96/96
  simultaneous connections all succeeded. It is the cumulative fork count over
  a boot's lifetime that matters, not the instantaneous connection count.
- **Not `shared_buffers`.** 16GB / 1GB / 128MB gave 2 / 4 / 2 forks.
- **Not kernel stack depth**, or not only that: rebuilding with
  `threads_default_kernel_stack_size` 65536 -> 262144 (4x) moved the budget
  from 1-2 forks to 4. It helps slightly and does not fix it.
- Backtrace is empty on the stripped loader, so there is no frame data yet.

This is what blocked the HammerDB NOPM sweep: a TPC-C run needs to survive far
more than a handful of backend spawns. Reporting it here because it is
fork-gated and reproducible, and because the A/B above is unaffected by it (it
is visible in both arms equally and the surviving samples are consistent).

### Substrate notes that cost time, for the next run

- `scripts/build` sets `use_openzfs` **only** from the `--use-openzfs` flag,
  at line 176. `conf_zfs=openzfs` selects the OpenZFS *kernel objects* but does
  not change image assembly. Without the flag the root image is built by
  booting a guest and pushing the manifest over `cpiod`, which wedged 4/4 here
  (`BrokenPipeError` / `ECONNRESET` mid-upload) including on the supported
  `scripts/run.py` path. Separately, `upload_manifest.py` hardcodes `-m 4G`, so
  `zfs_builder_mem` has no effect.
- `fs=rofs` avoids that entirely: `gen-rofs-img.py` builds the root image
  host-side with no guest boot. PGDATA still goes on a real OpenZFS pool on a
  separate disk, so this does not compromise a no-tmpfs requirement.
- OpenZFS was verified at build time by object counts, not by a banner: **302**
  `open_zfs` objects, **0** `bsd/.../fs/zfs/spa*.o`, and the OpenZFS-only libs
  (`libzfs`, `libzfs_core`, `libuutil`, `libzutil`, `libshare`, `libtpool`).
- All required pool properties verified `local` from inside the guest:
  `ashift=12 recordsize=8K compression=off atime=off sync=standard
  logbias=latency primarycache=all`. Note `ashift` is a *pool* property, so it
  needs `zpool get`, not `zfs get`.
- `virtio-net` with `vhost=on` leaves the guest unreachable: it binds its IP
  and logs `eth0: <addr>` but never answers ARP. `vhost=off` works.
- Both pre-boot greps pass on the built tree after every submodule operation:
  `remap_segments_in_current_address_space` = 3, `.PHONY: module build` = 1.
- The `apps` pin needs care: `b38b99c` has `.PHONY: module build` = 0 and
  re-arms the shadowed-build trap; the fix `5cb1903` is its direct child. This
  branch pins `5cb1903`.
- For an `fs=rofs` PG image the size check is **119 MB** with 6 PG manifest
  entries (payload is 112 MB: postgres 48M, lib 15M, share 3M, data-seed 46M).
  A silently PG-free build is ~8 MB.

### Not delivered

The HammerDB NOPM sweep, blocked by the fork-count crash above. No NOPM
numbers are reported rather than reporting numbers from a guest that dies
mid-run.

PR #1458, opened 2026-07-28
[DRAFT / tracking - DO NOT MERGE as one unit] PostgreSQL-on-OSv fork+ZFS integration (will split into S1..S6 on #1455/#1456/#1457 + #1423)

gburd:
## Note (submodule pin was broken, now fixed)

If you tried to clone this branch before now, the `apps` submodule would not
initialize. The recorded gitlink was `57c607984393901ce556a75f81ea40f7c469d1f3`,
a commit that exists in no repository (not `cloudius-systems/osv-apps`, not my
fork, not any local clone). It was recorded from a working tree where
`postgres18-musl/` was staged but never committed, so the superproject captured a
SHA that was never created:

    git submodule update --init apps
    fatal: remote error: upload-pack: not our ref 57c607984393901ce556a75f81ea40f7c469d1f3

Fixed in `b01994d04`, which re-pins `apps` to
`b38b99cd9302af01fb388e569bc0911c0e90bbc4` (reachable as
`refs/pull/73/head`, osv-apps#73). Verified with a fresh clone rather than by
inspection: `git submodule update --init apps` now checks out `b38b99cd`, and the
other seven submodule pins on this head resolve too. No source change; only the
recorded submodule commit.

## Draft / tracking -- DO NOT MERGE AS ONE UNIT

This is the full **PostgreSQL-on-OSv integration branch** (`integ/pg-fork-zfs`, 67 commits). It is filed as a **draft tracking PR** to make the completed fork-completeness + ZFS-coherence work **visible with its dependency chain**, NOT for review as one 67-commit unit.

It **will be split** into the logical follow-on stack (S1..S6 below) once its bases merge. Until then this branch is the integration reference that proves stock, unmodified PostgreSQL boots, forks its aux processes, serves rows, and sustains concurrent load on OSv over OpenZFS-on-NVMe.

## Depends on (remaining blockers)

The fork trilogy (opt-in, `CONF_fork`, off by default):

- **#1455** -- `process: implement thread-backed fork()/vfork()/execve()/waitpid()` (OPEN -- still to merge)
- **#1456** -- `mm: per-child copy-on-write address space for fork()` (stacked on #1455; OPEN -- still to merge)
- **#1457** -- `libc: honor POSIX default-ignore disposition for SIGCHLD/SIGURG/SIGWINCH` (**MERGED** 2026-08-03)

And, for the ZFS-coherence + PG-on-ZFS fixes:

- **#1423** -- `zfs: selectable in-kernel ZFS (BSD or OpenZFS 2.4.2) via upstream submodule + patch series` (**MERGED** 2026-08-17)

So of the four bases, #1423 and #1457 have merged; the split is now blocked only on the two remaining fork PRs #1455 and #1456.

This branch **already contains** the #1423 commits and the fork-trilogy commits inline (it was built on top of them); the split PRs below will be rebased to sit *on top of* the merged bases so each is independently reviewable.

## The intended split (S1..S6), in dependency order

Every fork-side commit on this branch is prefixed by its destination
(`[fork-stack / CONF_fork]`, `fork:`, `fork arena:`, `mm/fork:`, `mmu:`,
`process:`), and every ZFS-side commit by `[#1423 / OpenZFS]` /
`zfs(openzfs):`. The split follows those groupings:

**S1 -- AS-aware mmap allocation path** (stacks on #1455 + #1456)
Thread `address_space*` through the mmu *allocation* path (allocate/map_anon/
find_hole/evacuate/protect/unmap/mprotect/msync + vma_range_set), defaulting to
the kernel AS so the non-fork path stays byte-identical; give a child AS its own
`vma_range_set`. Fixes fork-child mmap landing in the global vma_list invisibly
to the child fault handler (the W-mmap wall). Also `fork: preserve file_vma type
in clone_address_space`.
Key commits: `mmu: make the mmap allocation path address-space aware`,
`fork: preserve file_vma type in clone_address_space (fix wild-branch)`.

**S2 -- Per-child COW arena + same-VA COW stack** (stacks on #1456)
Private COW-able heap arena for fork children; COW 2 MB large pages; same-VA COW
stack for deep call chains; lock-free arena allocator; eager arena population;
identity-heap TLS/syscall stacks; intrusive zombie reaping (no alloc on the
lifecycle path).
Key commits: the `mm/fork:` trio + the nine `fork arena:` commits.

**S3 -- fd inheritance across fork** (stacks on S1/S2)
Child gets its own reference on inherited fds; keep the shared fd slot when the
owner closes an fd a live child still holds (backend connection-socket wall).
Key commits: `fork: give the child its own reference on inherited fds`,
`fork: keep the shared fd slot when the owner closes an fd a live child inherited`.

**S4 -- Per-process signals + timer parking** (stacks on #1457)
Per-process signal dispositions + deliver blocked SIGCHLD to the handler
(postmaster PM_RUN wall); park app-thread timers off the per-CPU list across
address spaces (IRQ-context COW wall).
Key commits: `fork: per-process signal dispositions + deliver blocked SIGCHLD`,
`fork: park app-thread timers off the per-CPU list across address spaces`.

**S5 -- Cross-AS coherence (kernel structures)** (stacks on S1..S4)
Route kernel wait-records / list nodes / latches that a forked backend and AS0
must both see onto the identity kernel heap: rwlock read-waiters, signal-waiters,
rcu_defer wait_records, epoll containers + watcher lists, semaphores, waitqueues,
renamed-dentry d_path, `kill(OSV_PID)` routing to AS0 for parallel workers,
anonymous MAP_SHARED coherence for `shared_buffers`, POSIX-shm/DSM/ramfs/net
RX+TX coherence, thread-stack coherence, application_runtime + atfork handlers.
Key commits: the `fork: cross-AS coherence ...` set + the `[fork-stack /
CONF_fork] route ... onto the identity heap` set.

**S6 -- ZFS-coherence + PG-on-ZFS durability fixes** (stacks on S5 **and #1423**)
Route ZFS taskqueue + large kmem allocations to the identity linear map/heap so
forked-backend writes don't deadlock; cross-AS coherence for the ZFS
I/O-completion path (bio, zio/SPL heap, libsolaris.so statics); virtio-blk req
coherence. Plus the `[#1423 / OpenZFS]` durability patches (0028 vnode v_size
refresh, 0029 zfs_write_simple/zfs_space for ZIL replay, 0030 single-threaded
inline ARC eviction, taskq_wait quiescence) -- these amend **#1423**, not the fork
stack, per the standing rule.

## Split & landing rule (from the program's standing rules)

- Fixes that are **arena-independent and correct on their own** fold into the
  shipping fork PRs (#1455/#1456/#1457) to strengthen them.
- **OpenZFS / OSv-OpenZFS glue** bugs amend **#1423** (`pr/openzfs-draft`).
- **Arena-only / fork-COW-divergence** fixes stay in the fork follow-on stack
  (S1..S6 above), gated `CONF_fork`; the non-fork OSv path stays byte-identical.

## Blocks / part of

Part of the PostgreSQL-on-OSv North Star (Milestone 1: stock unmodified PG at
parity with PG-on-Linux over ZFS). This branch is where the M1 walls were
root-caused and fixed; the app side is the osv-apps `postgres18-musl` demo
(held on this stack landing on master).

---

**Status:** WIP / DO-NOT-MERGE as one unit. Will be split into S1..S6 and each
sub-PR opened for review once the remaining fork bases #1455 / #1456 merge
(#1457 and #1423 have already merged). Filed as a draft so the dependency chain
and the completed work are visible.

---

## Banked fix: shared-gfdt fd-reap EMFILE leak (`bank/fork-fd-reap-leak`)

Root-caused while characterizing why stock PostgreSQL wedges under sustained
connection + autovacuum churn on this stack (the wedge that a reboot-per-cell
benchmark hides). The single global fd table (`gfdt[FDMAX]`) is shared across
fork children. On reap, `release_inherited_fds()` reclaimed the child's
inherited fds but only `erase()`d the tracking for the fds a backend opened
itself after fork - it never closed those slots. Every backend that opened an
fd post-fork and exited without closing it leaked the slot; gfdt fills to
FDMAX and `pipe()`/`accept()` fail EMFILE until reboot.

Fix (one clean commit, `CONF_fork`, gated `OSV_FORK_FD_REAP` default on): close
each still-open child-opened fd on reap (null the slot iff it still points at
that file, then drop the ref; these fds are private to the exited backend so
closing is unconditional). Lock order preserved (snapshot under `g_fd_lock`,
close after release).

Measured, 32-connection churn loop + autovacuum, stock PG on OSv/OpenZFS, 400
rounds:

| | before | after |
|---|---|---|
| gfdt occupancy | climbs to 65535 (FDMAX) | flat ~29-46 |
| first EMFILE | round ~224 (wedge) | none, 400 rounds ok |

Stacks on the fork-fd-table generation (the shared-gfdt model with
`g_inherited_fds` / `g_child_opened_fds` / `fork_child_owns_fd`), a three-commit
redesign of fork fd handling that itself depends on the fork arena
(`fork_arena::kernel_heap_scope`). That generation is NOT in #1455 / #1456 -
those carry the older, simpler fd model and no fork arena - so this fix cannot
be a standalone-vs-master PR and cannot be cherry-picked onto #1456 (it
conflicts on the missing `fork_arena.hh` and would not build). It lands as one
clean commit on top of whichever PR proposes the shared-gfdt fd-table
generation, above the fork arena, above #1455 / #1456.

Review-ready (probe-stripped, G-signed, compiles under `CONF_fork`, both
changed TUs build to `.o` and the new reap helper links) on gh-fork branch
`bank/fork-fd-reap-leak-reviewready`: the fork-stack base (carrying the
gfdt generation) + the pure two-file reap fix on top. Blocker: its minimal
correct base is ~134 commits ahead of master with no open PR yet, so it is
tracked here rather than opened as a PR against a base that does not exist
upstream.

A second, independent per-fork memory leak (unreclaimable, ~0.77 MB/backend,
surfaces once the fd leak is fixed) is root-caused and FIXED on gh-fork branch
`bank/fork-arena-real-reclaim` (two G-signed commits, probe-stripped, on the
same fork stack). This supersedes the `bank/fork-arena-grow-mitigation` branch,
which only delayed exhaustion; the grow branch stays as a fallback but the real
fix is preferred.

Mechanism (a two-layer bug). The fork arena carves each chunk from a single
global monotonic bump pointer, and `release_as()` never returns a dead child's
carved VA, so under sustained fork/reap churn the 512 MiB arena VA is exhausted
after ~700 backends. Post-exhaustion every backend's heap falls through to the
identity kernel heap, which `destroy_address_space()` never frees on reap, so
each reaped backend leaks ~0.77 MB and a long-running fork-heavy guest OOMs. An
app-vs-kernel-lifetime histogram of the fallback shows the leaked bytes are
99.9% the backend's own application heap (dies with the backend); essentially
no kernel-lifetime object reaches the fallback. The fix keeps post-exhaustion
allocations in a per-address-space COW-private overflow region (a dedicated app
slot, mapped only in the allocating child) reclaimed wholesale by the existing
`destroy_address_space()` page-table teardown on reap; `free()` of an overflow
pointer is a no-op recognized by a BSS range check, so a cross-AS free never
reads a header at a VA mapped only in the dead child. Gated by
`OSV_FORK_ARENA_RECLAIM` (default on; =0 restores the leak for A/B).

Because the per-AS overflow region is mapped in only one AS (unlike the arena,
which is COW-mapped in every AS and so silently serves a divergent copy on a
cross-AS read), it turned latent cross-AS fork-coherence bugs into hard faults.
An audit of the AS0/IRQ/RCU-touched object graph found a BOUNDED set of
per-connection kernel objects allocated with C++ `new`/`aligned_new` (bypassing
the wholesale `uma_stub` identity-heap scope): `socket`, `inpcb` (tcp/udp/raw),
the six `serial_timer_task` nodes, the RX `net_channel`, the zerocopy
`ztx_handle`, and the libc stdio `FILE` object (on the process-global
`ofl_head` list). The first commit scopes all of these onto the identity heap;
it is a generic fork-coherence fix valuable independently of the leak fix (each
was a latent divergent-COW-copy read the arena was masking). Boot/device-init
`new` sites are unreachable by a child app thread and left unscoped.

Evidence (m5d.metal, A/B on `OSV_FORK_ARENA_RECLAIM`):

| Gate | Fix ON (default) | Fix OFF (control) |
|------|------------------|-------------------|
| (a) mem_free over 400 rounds | FLAT (~22.6 GB) while 90 GiB overflow served, 26,181 backends reaped | monotonic drop, ~0.85 MB/backend |
| (b) wedge/crash to round 400 | none, ok=32 fail=0 | leak degrades the guest |
| (c) totalCORRUPT (8 KiB md5 + prefix-bleed) | 0 | 0 |
| (d) cross-AS kernel-object survival | 0 recover magic-mismatch, 0 mmap_fault in any vrele/drele/~file/RX backtrace across 26,181 reaped backends | n/a |
| (e) identity-heap fallback | peaks at 16 MB total (bounded map-miss) | 17,289 MB (the leak) |

Same fork-stack landing dependency as leak #1: it lands as clean commits on top
of whichever PR proposes the shared-gfdt fd-table generation, above the fork
arena.


gburd:
## Fork-only sched follow-on for the S-split: parked-timer unlink before idle-pull

Recording a small `CONF_fork`-only scheduler-correctness commit that belongs on this stack (it references machinery that exists only on the fork build, so it is not standalone against master).

The idle-CPU pull work-stealing path (`donate_to_puller()`, from #1472) migrates a runnable thread from the busiest CPU to an idle one. On a `CONF_fork` build a preempted app thread may be parked on the victim CPU's per-CPU timer list (park_timers runs on every app-thread switch-out to keep a forked child's on-stack timer nodes out of a foreign address space). `donate_to_puller()` called `suspend_timers()` but, unlike `load_balance()`'s source-side migration, did not first `cpu::unlink_parked()` the thread. A donated parked thread was then left linked on the source CPU's parked timer list, and a later timer IRQ on that CPU could walk the migrated thread's on-stack timer node through the wrong address space and fault in non-preemptable context.

The fix mirrors `load_balance()`'s existing `#if CONF_fork cpu::unlink_parked(mig)` exactly and is `conf_fork=0` byte-identical. `cpu::unlink_parked()` is a safe no-op if the thread is not parked, so it is harmless for the non-parked case and correct for the parked one.

Branch: `pr/parked-timer-unlink` (single commit on top of the idle-pull cherry-pick, so it has both `donate_to_puller` and the `unlink_parked` machinery). This is a follow-on to #1472; #1472 itself is correct for a stock build, and only the fork build needs this additional unlink. Filed here so it lands as an S-split item alongside the rest of the fork stack rather than as a doomed standalone master PR.

gburd:
Tracking update: the current authoritative integration tip is now `wip/pg-zfs-bench-v20` (fast-follow on the `integ/pg-fork-zfs` snapshot filed here). It carries the same intended S1..S6 split; the following fork-completeness / cross-address-space coherence follow-ons were added on top of the snapshot above and belong to the fork-stack groups (all `CONF_fork`-gated, byte-identical when fork is off):

- elf: share only read-only library segments across fork, not writable ones (S1/S3 group).
- elf: share dlopen'd library segments across fork address spaces.
- mm: allocate file-backed and split vmas on the identity heap under fork.
- fork(x86_64): allocate the child thread + detached_state on the identity heap.
- libc: allocate lazy_indirect backing objects on the identity heap under fork.
- fork: keep vma-range nodes and dlopen'd ELF bookkeeping on the identity heap.
- open_zfs: ARC-eviction waiter on the identity heap for fork (ZFS-coherence group).

Two of these coherence fixes are also broken out as standalone master-eligible follow-ons where they stand on their own:

- the wake-path null/invalid-thread guard is #1484 (`sched: skip waking a null or invalid thread pointer`).
- the ARC half-of-RAM default cap for app coexistence is #1487 (draft, follow-on to #1423/#1478).

These are recorded here so the completed work is captured in the tracking PR while the trilogy bases (#1455/#1456/#1457) and #1423 await review. No change requested on this PR; it remains DO-NOT-MERGE-as-one-unit.

gburd:
Head fast-forwarded by one commit to fix the `apps` submodule pin.

This branch pinned `apps` at `b38b99cd`, which predates the `postgres18-musl` Makefile fix for a shadowed `build` target. Without that fix, `module: build` names a target of the same string as a real `build/` directory the module creates, so make treats it as up to date, skips the recipe, and the module target succeeds having built nothing. The observable result is an image containing no PostgreSQL while the build exits zero: roughly 8 MB where a working image is about 119 MB.

The new pin `5cb1903` is `b38b99cd` plus exactly that one-line fix, so this is a strict fast-forward with no code change to the branch itself.

Since this branch exists as a tracking reference rather than something to merge, a bad pin here is worse than elsewhere: anyone using it as a starting point inherits a build that quietly omits the application under test. That is not hypothetical, it cost two separate benchmark runs before the cause was found.

gburd:
## Proposal: how to land the newer fork generation upstream (and where two validated leak fixes fit)

This tracking PR has grown into a large integration branch. This comment proposes a concrete decomposition to get the reviewable pieces upstream, and records two fork-teardown fixes that are complete and validated but currently have nowhere to land. It is a composition proposal for maintainers to accept or reshape, not a request to merge this branch.

### The situation
The open fork PRs are #1455 (base thread-backed fork/vfork/execve/wait) and #1456 (per-child COW address space), which carry the original, simpler fork model. The PostgreSQL-on-OSv work here evolved a newer, more complete fork implementation on top of that idea: a COW-private heap arena for fork children, a shared global fd table with correct inherited-fd semantics, lock-free COW fault handling, timer/IRQ/thread-lifecycle coherence across address spaces, ELF/dlopen segment sharing, and the OpenZFS glue. It is a larger body of work that is not an increment on #1456 (the fd model differs), and it currently only lives on this integration branch.

### Two validated fixes waiting on it
1. fd-table EMFILE leak on backend reap. On reap the child's inherited fds were released but the fds a backend opened itself after fork were only dropped from tracking, never closed in the shared table. Under connection churn the table fills to the fd limit and pipe()/accept() fail EMFILE, wedging the guest. A/B over 400 rounds: table occupancy flat with the fix vs climbing to the limit and failing around round 224 without it.
2. fork-arena memory leak. The arena's bump pointer never returns a reaped child's virtual addresses, so after a few hundred backends the arena exhausts and every backend's per-connection heap falls to the identity heap, which is never freed on reap (roughly 0.85 MB per backend, leading to OOM). The fix gives each child a private overflow region reclaimed wholesale on teardown, plus scopes eight cross-address-space-referenced kernel objects (socket, inpcb, tcp timer, net_channel, stdio FILE, and similar) to the identity heap. A/B over 400 rounds / 26,181 reaped backends: memory flat with the fix vs a monotonic drop without it, zero corruption, zero cross-AS fault at scale.

Both are generic fork/teardown correctness fixes; PostgreSQL is just the workload that exercises sustained fork churn. Neither is PG-specific.

### A finding worth surfacing regardless of composition
The fork arena is COW-mapped in every address space, so a cross-address-space raw-pointer read returns a divergent private copy instead of faulting. That silently masks latent cross-AS correctness bugs. Making per-connection memory per-AS (the second fix) turned those into hard faults and exposed eight pre-existing coherence gaps, for example a stdio FILE linked on the process-global open-file list, and a net_channel RX ring written by the classifier in IRQ context from another AS. An audit of the AS0/IRQ/RCU-touched object graph showed that set is bounded and enumerated it. The working invariant is that any object referenced across address spaces must live on the identity heap. This is relevant to how the fork model is reviewed, since the arena's map-everywhere property can hide such gaps.

### Proposed decomposition (a stacked series, not one merge)
The body groups into coherent, independently reviewable units:

- S1 fork-arena: COW-private heap arena for fork children
- S2 fork COW address-space and clone-range correctness and performance
- S3 timer / IRQ / thread-lifecycle coherence across address spaces
- S4 ELF/dlopen segment sharing across fork address spaces
- S5 shared fd-table: inherited-fd semantics and reap-close (fix 1 lands here)
- S6 signal routing to the correct address space
- S7 the OpenZFS glue the fork path needs (tracks with #1423)
- S8 fork-arena overflow reclaim and the cross-AS object scopings (fix 2 lands here)

The series is strictly stacked on #1455 and #1456, so nothing above them can open as a mergeable PR until those land. Suggested order once composition is settled: S1, then S2, then S3/S4/S6 in parallel, then S5, then S8.

### The one decision that unblocks this
Does the newer fork generation become the proposed fork implementation, superseding or extending #1455/#1456, in roughly the S1..S8 shape above? Once that is decided the pieces can open as reviewable PRs and the two validated fixes land as their natural commits in S5 and S8. Everything is built and the two fixes are A/B-validated at scale; the blocker is purely this composition decision.

### What this does not claim
This does not claim OSv-vs-Linux PostgreSQL parity. It removes the known wedges and leaks and shows the earlier apparent scaling decline was largely a benchmark run-order artifact, but a clean fresh-boot-per-cell comparison on the merged tree has not been run, and a separate per-commit wakeup-latency factor remains open. Parity is an open measurement.

gburd:
> **WITHDRAWN (see the later comment): the scaling table and the "flat across the range" conclusion in section 2 below do not reproduce.** Re-measured on a single harness, OSv VU32 is 1.73x higher than reported here while Linux reproduces within 6%, giving 0.87x rather than 0.53x. The cause is not yet proven. The memory-footprint result in section 1 is unaffected and did reproduce.

## Update: the fork per-backend memory footprint, and where the scaling limit actually is

Two results since the last update on this branch, both measured on a matched pair (one bare-metal host, one raw NVMe each, identical ZFS geometry, io_uring on both sides, fresh boot per benchmark cell). Both are generic fork/memory fixes; PostgreSQL is only the workload that made them visible.

### 1. A forked child's private memory tracked its allocation churn, not its live heap

The COW heap arena hands each address space memory from one global bump pointer. Once that global region is used up, every later allocation took a fresh eagerly-populated region, and `free()` on those was an unconditional no-op: the region was only reclaimed wholesale when the address space died. That is safe (a cross-address-space free must not read a chunk header at an address the freeing context may not have mapped) but it means a long-lived child's committed memory grows with its *lifetime* allocation churn rather than its live heap. Any process that allocates and frees steadily in a forked child pays this; a database backend that builds and tears down per-query memory contexts is simply the fastest way to hit it.

Measured at 73 live address spaces, attributing per-child private memory:

| term | per child | share |
|---|---|---|
| arena copy-on-write faults | 4.0 MB | 1.6% |
| page-table clone | 1.1 MB | 0.4% |
| privatized stacks | 1.0 MB | 0.4% |
| **post-exhaustion region commit** | **245 MB** | **99%** |

The fix carves the overflow area into one disjoint window per address-space slot, so the owning slot follows from the address arithmetically. `free()` then distinguishes same-address-space from foreign with a check that reads no arena page at all: the cross-address-space safety property is preserved exactly, and only the same-address-space case changes, which is where recycling was missing.

| | before | after |
|---|---|---|
| private per child | 247 MB | **10.0 MB** |
| live overflow | 17.5 GB | 283 MB |
| free memory | 29.3 GB | 46.8 GB |

10 MB per child is the same order as a native process. Under load the fix recycled 1.43 TB of cumulative churn out of 1.66 GB live, and per-child memory stayed flat at 9.4-12.5 MB from 42 up to 202 address spaces. The count that previously exhausted a 64 GiB guest now completes, and the guest reached 576 concurrent children while still holding 42 GB free, so that number is a floor set by a configuration limit rather than by memory.

Two incidental findings worth recording for anyone reading the arena code: a writable large page shared across fork is already kept whole rather than split, so that concern is closed; and the earlier "wedge" at high concurrency was this memory exhaustion, not a lost wakeup. Six separate wakeup-path experiments all still wedged, and the console showed an out-of-memory abort in the reclaimer. A reclaim stall presents exactly like a lost wakeup from the application's side: every worker blocked on one lock, no device traffic, idle CPU.

### 2. With memory fixed, the remaining limit is the asynchronous write-completion path

Re-running the sweep with the fix in place, equal cache ceiling on both sides, and client-side instrumentation to prove the load generator was never the constraint (it stayed under 10% CPU throughout):

| clients | OSv | Linux | ratio | OSv private/child | OSv CPU | Linux CPU | OSv device util |
|---|---|---|---|---|---|---|---|
| 32 | 151,527 | 285,228 | 0.53 | 9.6 MB | 53% | 62% | 53% |
| 64 | 166,217 | 284,552 | 0.58 | 10.1 MB | 54% | 84% | 95% |
| 128 | 153,299 | 361,752 | 0.42 | 12.0 MB | 59% | 86% | 97% |
| 192 | 149,905 | 204,343 | 0.73 | 12.5 MB | 58% | 88% | 95% |

The memory wall is gone and it revealed a throughput ceiling underneath: OSv is flat across the range while leaving roughly 40% of its CPU unused with the device pinned near saturation, where the comparison runs at 84-88% CPU with the device at under half utilization.

The decisive experiment was removing durability. If the limit were the synchronous commit path, dropping `fsync` should have closed the gap. It widened it: the comparison doubled to 600k and became CPU-bound as expected, while OSv fell to 37.7k with **6% CPU and an idle device** - waiting, not working. `LWLock/AioUringCompletion` appears in the top three waits of both OSv cells and neither comparison cell. So the limit is in asynchronous write completion (the io_uring and ZFS writeback path), not in commit durability.

Related and probably the same subsystem: three of eleven high-concurrency cells stopped making progress with 99.2-99.5% of time in `AioUringCompletion`, device idle, CPU idle, memory healthy. That is a liveness problem distinct from the memory fix, and it is the next thing being diagnosed.

### Scope

The memory fix is generic: it affects any forked child that allocates and frees over its lifetime. The measurements above are reproducible with the harness in #1512. Full evidence, including the hypotheses that were refuted along the way and retracted in place, is kept with the branch record rather than summarized selectively here.

gburd:
## Withdrawing the scaling table from my earlier comment

The comparison numbers I posted above do not reproduce, and the error is one-sided in a way that made OSv look worse than it is. Withdrawing them.

Re-measuring on a single harness, same box, same commit, VU32 durable:

| arm | reps | median NOPM | vs what I published |
|---|---|---|---|
| OSv | 5 | 261,760 | **1.73x higher** |
| Linux | 1 | 301,922 | 1.06x (reproduces) |

So the ratio on that harness is **0.87x** (range 0.84-0.94), where I published 0.53x. OSv rep spread was only 1.13x, so 1.73x is far outside run-to-run noise, and Linux landing within 6% of the published figure makes this one-sided: it is not the box, the driver, or the schema. Equivalence was checked rather than assumed - identical GUCs read back from the running server, byte-identical driver script, same commit, both PostgreSQL builds linking real liburing, and guest CPU nearly identical (1739 vs 1700 of 3200) while doing 1.7x the work.

What I can and cannot say:

- **Cannot say** the curve is 0.87x everywhere. Only VU32 was re-measured, so the *shape* across VU64/128/192 is untested. The honest statement is "not reproducible as published, withdrawn pending a re-run", not a replacement table.
- **Cannot yet say why.** I found and fixed a real harness bug (a `pkill -f "qemu-system..."` pattern that matched its own sudo wrapper, killing the wrapper while qemu survived holding the image write-locked, so the next boot failed), but I have not proven that is the cause of the 1.7x. Two other explanations I proposed were refuted by measurement: crash-recovery overhead (my own cells show it too) and cache starvation (the earlier run's ARC was in fact *larger*). The unexplained residual is that the earlier OSv cells show `AioUringCompletion 12.3%` and `AioIoUringExecution 6.6%` in their wait profiles and the new ones show neither. That is the open question.

The consequence for the conclusion I drew: "OSv is flat across the range while leaving 40% of its CPU unused" rests on the withdrawn numbers and I am withdrawing it too. The memory-footprint result in that comment is unaffected - it is a direct per-address-space measurement, not a throughput ratio, and it reproduced consistently (9.4-12.5 MB per child, flat from 42 to 202 address spaces).

Separately, the same run produced one real fix worth having on its own terms, and it is deliberately not a throughput claim. `vdev_disk_open()` never sets `vdev_nonrot`, so OpenZFS treats an NVMe device as rotating and picks the 1 MiB aggregation limit with gap-stitching instead of the 128 KiB non-rotating cap. Setting it eliminates every request above 128 KiB (114,271 to 0) and cuts `w_await` 55% and queue depth 39%, with **total bytes written unchanged (+1.4%) and throughput down 3%**. It is a correctness and latency fix for solid-state vdevs, not a speedup, and I will file it that way.

PR #1463, opened 2026-07-31
virtio-net: implement multiqueue (VIRTIO_NET_F_MQ)

gburd:
## What

OSv's virtio-net driver hardcodes a single Rx/Tx virtqueue pair drained by one `receiver()` poll thread. Every guest's inbound network I/O funnels through that one queue on one CPU. `VIRTIO_NET_F_MQ` exists as a feature constant but was never negotiated.

This implements multiqueue:
- Negotiate `VIRTIO_NET_F_MQ` + `VIRTIO_NET_F_CTRL_VQ`; read `max_virtqueue_pairs` from device config; use `nqp = min(advertised, num_vCPUs, MSI-X budget)`.
- `_rxq`/`_txq` single structs become N-element vectors; N Rx poll threads, one per pair, pinned across distinct CPUs.
- Per-queue MSI-X vectors; Tx queue selected by CPU id; `VIRTIO_NET_CTRL_MQ_VQ_PAIRS_SET` issued on the control queue.
- The `nqp == 1` path (device without MQ, e.g. QEMU SLIRP) is unchanged.

## Prerequisite fix: MSI-X vector budget (also fixes a boot crash)

> This commit is also proposed on its own as #1465, which is independently
> reviewable and applies directly to master. If #1465 lands first, this PR
> reduces to the virtio-net multiqueue commit alone.

Enabling multiqueue on a many-vCPU guest exposed a pre-existing crash: OSv `abort()`s in `interrupt_descriptor_table::register_interrupt_handler` when the 224 usable IDT vectors (32..255) are exhausted. With several multiqueue virtio-blk devices at high `-smp` this already crashed at boot before any net change. This PR adds a process-wide MSI-X vector budget: `register_interrupt_handler` returns 0 (no vector) instead of aborting when the IDT is full, and virtio-blk and virtio-net cap their interrupt-bearing queues against the shared pool.

## Validation

- **Multiqueue activates:** with a vhost NIC advertising 8 pairs (mq=on), the driver negotiates F_MQ, spins up 8 Rx threads pinned to 8 distinct CPUs, and the device ACKs `CTRL_MQ_VQ_PAIRS_SET(8)` reaching DRIVER_OK. A single-queue device falls back to nqp=1.
- **Boot stability:** native-example boots to ready 10/10 at each of `-smp 4`, `-smp 16`, and `-smp 48`.
- **MSI-X budget A/B:** identical config (8x virtio-blk num-queues=48 + virtio-net mq=on, `-smp 48`): stock master aborts 3/3 in `register_interrupt_handler`; patched boots 5/5.

## Note on throughput

The motivation is concurrent-workload scaling (single-queue net serializes many concurrent connections' RX through one thread). I verified the MQ *mechanism* (8 active queues/threads) and boot stability here; the end-to-end concurrent-throughput curve is best shown with a many-connection server workload and I have not included those numbers in this PR. Bulk single-stream TCP is not a faithful measure (vhost GRO coalesces it so one Rx thread suffices). The change is a strict superset of the current single-queue behavior (nqp=1 unchanged), so it is safe independent of the throughput delta.

### DIFF
--- a/arch/x64/exceptions.cc
+++ b/arch/x64/exceptions.cc
@@ -118,7 +118,12 @@ unsigned interrupt_descriptor_table::register_interrupt_handler(
             }
         }
     }
-    abort();
+    // No free IDT vector.  Vectors 0..31 are CPU exceptions and are never
+    // returned here, so 0 is an unambiguous "exhausted" sentinel.  Returning it
+    // rather than aborting lets MSI-X allocation (msix_vector) fail gracefully
+    // and lets a device fall back to fewer queues instead of crashing the boot
+    // when many multiqueue devices exhaust the 224 usable vectors.
+    return 0;
 }
 
 void interrupt_descriptor_table::unregister_handler(unsigned vector)
--- a/drivers/msi.cc
+++ b/drivers/msi.cc
@@ -25,7 +25,11 @@ msix_vector::msix_vector(pci::function* dev)
 
 msix_vector::~msix_vector()
 {
-    idt.unregister_handler(_vector);
+    // _vector == 0 means IDT vector allocation failed (see the ctor); there is
+    // nothing to unregister in that case.
+    if (_vector) {
+        idt.unregister_handler(_vector);
+    }
 }
 
 pci::function* msix_vector::get_pci_function(void)
@@ -198,7 +202,16 @@ std::vector<msix_vector*> interrupt_manager::request_vectors(unsigned num_vector
     auto num = std::min(num_vectors, num_entries);
 
     for (unsigned i = 0; i < num; ++i) {
-        results.push_back(new msix_vector(_dev));
+        auto* vec = new msix_vector(_dev);
+        // A zero vector means the global IDT vector pool is exhausted.  Stop
+        // here and return however many we did get; easy_register() treats a
+        // short result as failure so the caller can fall back to fewer queues
+        // rather than crash.
+        if (!vec->get_vector()) {
+            delete vec;
+            break;
+        }
+        results.push_back(vec);
     }
 
     return (results);
--- a/drivers/virtio-blk.cc
+++ b/drivers/virtio-blk.cc
@@ -137,6 +137,20 @@ blk::blk(virtio_device& virtio_dev)
     // base class did not set up.
     _num_queues = virtio_driver::_num_queues;
 
+    // Cap the number of queues we arm with an interrupt against the shared
+    // MSI-X vector budget.  Each queue uses one vector, and the x86-64 IDT has
+    // only 224 usable vectors shared by every device; a high-vCPU guest with
+    // several multiqueue virtio devices can otherwise exhaust them at boot.
+    // Extra probed virtqueues beyond this cap are simply left unused - I/O is
+    // directed only to queues 0.._num_queues-1, all of which have a serviced
+    // interrupt.
+    unsigned granted = reserve_msix_vectors(_num_queues);
+    if (granted < (unsigned)_num_queues) {
+        virtio_i("virtio-blk: capping active queues from %d to %d (MSI-X budget)\n",
+                 _num_queues, granted);
+        _num_queues = granted;
+    }
+
     // Resize per-queue lock vector now that _num_queues is known.
     _queue_locks = std::vector<mutex>(_num_queues);
 
--- a/drivers/virtio-net.cc
+++ b/drivers/virtio-net.cc
@@ -19,6 +19,7 @@
 #include <string>
 #include <string.h>
 #include <map>
+#include <algorithm>
 #include <errno.h>
 #include <osv/debug.h>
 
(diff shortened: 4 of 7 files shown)


gburd:
## Update: benchmark results and a regression to fix before merge

I ran a direct A/B (multiqueue nqp>=8 vs single-queue nqp=1, same image, same NIC, an external client) with PostgreSQL/pgbench across concurrency levels, and the results change what I can claim for this PR. Converting to draft.

**Two honest findings:**

**1. The multiqueue feature does not improve concurrent throughput on this workload.** Single-queue (nqp=1) equals or beats nqp>=8 at every concurrency level (e.g. read-only tps at 16 connections: nqp=1 gave ~66k vs nqp=16 ~25k, reproduced). The throughput ceiling this workload hits is a per-connection CPU/scheduler cost and a storage write path, not RX-queue depth. So multiqueue is not the scaling win I expected; it is correct and it activates (N queues, N Rx threads, device ACKs CTRL_MQ_VQ_PAIRS_SET), but it does not move the numbers here.

**2. The MSI-X vector budget introduces a virtio-blk regression under load.** When virtio-net claims 2*nqp vectors at high vCPU count, it starves the several virtio-blk devices; the soft-fail path (register_interrupt_handler returning 0 instead of aborting) then leaves a blk queue with no interrupt vector, so completions are never delivered and reads fail with EIO under sustained write. The pre-MQ image was clean. This needs fixing: a queue that cannot get its own vector must fall back to a shared/polled vector rather than run interrupt-less, and net must not be allowed to starve blk.

**What stands:** the vector-budget change does fix a real pre-existing boot crash (many-vCPU guests with several multiqueue virtio-blk devices previously aborted in register_interrupt_handler on IDT vector exhaustion; they now boot). That part is worth keeping.

**Plan:** I will either (a) fix the vector arbitration so blk always keeps a working (shared/polled) vector and re-take the A/B, or (b) split the boot-crash MSI-X fix into its own PR and hold the multiqueue feature until it demonstrates a real gain. Marking draft until then.

wkozaczuk:
I am not surprised by your performance findings. I am seeing similar ones with the virtio-blk multi-queue multi-consumer -threads changes. Single queue, single consumer thread seems to be beating all other setups.

gburd:
That lines up with what I found, and it points at a shared root cause rather than anything queue-specific.

I profiled the concurrency collapse separately (a many-connection server workload over the network) and the guest was ~90% idle while throughput fell: threads were piling onto a few CPUs while the rest sat idle. The scheduler places a new thread on its creator's CPU and re-wakes a blocked thread on its last CPU, and the load balancer only migrated one thread per 100ms, far too slow to disperse a fan-out of many worker threads. That is why more queues (net or blk) did not help: adding consumers does not change where the woken threads land, so they still serialize on the same few CPUs.

I opened #1464 for the scheduler side (run the balancer more often and drain the whole imbalance per pass); it recovered +53% to +73% at 32 to 64 connections in an A/B. Your virtio-blk observation is consistent with the same effect: single queue/single consumer wins because it avoids paying migration/placement overhead for consumers that the balancer cannot spread anyway.

Given all that, I am holding this virtio-net MQ PR as draft. The multiqueue feature does not earn its keep on its own, and the vector-budget change in it also needs the virtio-blk starvation fix I noted above. The genuinely useful part is the MSI-X vector-budget boot fix (many-vCPU guests with several multiqueue virtio-blk devices previously aborted at boot); if useful I can split that out into its own small PR and drop the multiqueue machinery.

gburd:
Reworked: the MSI-X vector-budget logic this originally carried is factored out into #1465 (so blk can never be starved of interrupt vectors), and this PR now stacks on it with the multiqueue mechanism alone (N Rx virtqueues + N receiver threads pinned across CPUs, N Tx queues selected by CPU, CTRL_MQ_VQ_PAIRS_SET).

Honest performance note: on a many-connection PostgreSQL/pgbench workload I measured MQ vs single-queue on an otherwise-fixed multi-CPU guest and saw no reliable throughput change (gains within run-to-run noise; that workload is bottlenecked on wake/round-trip latency, not RX-queue depth, so more queues don't help it). So I am not claiming a benchmark win here.

The case for it is architectural, not workload-specific: single-queue serializes all inbound traffic through one receiver thread on one CPU, which caps any RX-bound concurrent workload regardless of core count. Multiqueue removes that serialization for the workloads where RX genuinely is the bottleneck. I am fine keeping it as draft pending an RX-bound benchmark that demonstrates the gain, or landing it as a general capability with the MSI-X budget (#1465) that makes it safe. Reviewers' call on that basis.

gburd:
## Follow-up: an RX-bound benchmark now demonstrates the gain

Earlier I held this draft because an isolated MQ-vs-single-queue A/B showed no reliable throughput change: that workload was bottlenecked on wake/dispatch latency downstream of RX, so more RX queues could not help it, and the effect of multiqueue was masked.

With that downstream bottleneck fixed as well (the idle-CPU pull dispatch work in #1472), the picture separates cleanly and multiqueue earns its keep. On a workload with many concurrent network connections driven over a real NIC from an external client:

- **Single-queue funnels all inbound traffic through one receiver thread on one CPU.** Under rising concurrency the aggregate guest CPU actually *drops* while throughput falls, and sampling shows CPUs idle-spinning waiting for the one receiver to dispatch: a classic single-serializer collapse on an otherwise-idle guest.
- **With multiqueue on, inbound work spreads across the queues' receiver threads.** Under the same concurrent load 7 of 8 CPUs go busy (versus the single receiver funnel), and mid-to-tail throughput rises by roughly 20-25% across the higher concurrency levels, 0-failed.

So the architectural claim I made before is now backed by a measured RX-bound result: N concurrent connections spread across N CPUs instead of serializing through a single receiver. Multiqueue is necessary but not sufficient on its own; paired with the dispatch fix it is a real, repeatable win.

**Dependency:** this still stacks on the MSI-X vector-budget fix (#1465, so many-queue guests keep working interrupt vectors for every device); #1465 should land first. Marking ready for review on that basis.

Issue #396, opened 2014-07-16
Load-balancer inefficiency with intermittent thread


nyh:
In OSv, migration of threads between CPUs works like this: Each CPU has a load-balancer thread which runs 10 times a second and sees if there are other CPUs with fewer runnable threads, and if so, it sends one of its threads to the less-loaded CPU.

One of the implications of this approach is that when a CPU becomes idle, it has to wait until some other CPU's load balancer wakes up and gives it work to do. In the time that passes - up to 1/10th of a second if we're unlucky - this CPU remains idle.

This problem is easily seen with test/misc-loadbalance's "intermittent thread" tests. In that test, we have two cpus running two main busy-looping threads, plus one "intermittent thread" - a thread which calculates for 1ms and then goes to sleep for 10ms - and so on. The problem is that sometimes, when the intermittent thread is running,
the load balancer decides to move the main thread to the other CPU; When the intermittent thread is finished, in a mere 1ms, there's no more work for this CPU to do, because the two main threads are on the other CPU, and it takes up to 100ms until the other CPU's load balancer gives us back a thread to run. Without this patch, "top" on the host shows that the guest uses around 180%-190% CPU. With this patch, it shows 200%.

I think we may be seeing the same problem with the page-access-scanner. In a 2-cpu SpecJVM compress benchmark, I see the page-access-scanner takes only 0.7-1% of the time one one CPU, yet the performance drop compared to no-scanner is 3%. I suspect that it might be related to the intermittent thread problem. However I'm not sure the intermittent thread problem really explains it - I can understand a 1% performance drop (1% of one cpu doing scanner, 1% of one cpu wasted), but not 3%.

In December, I sent a patch titled "RFC: Idle cores "pull" work" where a CPU becoming idle chooses a highly-loaded CPU, and wake()s up its load-balancer thread. Hopefully, this load-balancer will decide to send some work to this idle CPU.  Without this patch, "top" on the host shows that the misc-loadbalance uses around 180%-190% CPU. With this patch, it shows 200%. Unfortunately this patch also had a negative performance impact on tst-ctxsw "apart" case (too many unnecessary wakeups of the load balancer) so it still needs some improvement.

nyh:
The patch I mentioned 12 years ago: https://groups.google.com/g/osv-dev/c/iVO55bXrCIo/m/mQ-T_UKElMUJ
As that thread mentions, it helped one benchmark but hurt other benchmarks :-(

[...]

PR #1464, opened 2026-08-01
sched: disperse threads faster in the load balancer

gburd:
## What

`cpu::load_balance()` is the only mechanism in the scheduler that spreads threads across CPUs: `thread::start()` places a new thread on its creator's CPU, and `thread::wake()` re-wakes a blocked thread on the CPU it last ran on. Neither disperses work on its own. The balancer runs **once every 100ms and migrates a single thread per pass**.

That is far too slow for a workload that fans many threads out from one parent (a server that forks or spawns a worker per connection). The workers pile onto the few CPUs they were started or woken on while the rest of the machine sits idle, and a one-thread-per-100ms drip cannot catch up under a high request rate.

This changes the balancer to run every **10ms** and, on each pass, keep migrating the most-migratable queued thread to the least-loaded CPU until this CPU is no longer meaningfully more loaded, draining the whole imbalance in one pass. Work per pass is bounded by the CPU count so a transient spike cannot become a migration storm. It reuses the existing, proven migration path unchanged.

## Why (measured)

A many-connection server benchmark on a 16-vCPU guest, driven over the network, was collapsing past 8 connections. Profiling showed the guest was **~93% idle** at 32 connections while throughput fell: the sampler put 93.5% of CPU time in `cpu::idle()`, thread dumps showed ~37 worker threads crammed onto **4 of 16 vCPUs** with 12 idle, and the host saw the VM using about 6 of 16 cores. The bottleneck was thread placement, not lock contention, CPU saturation, or I/O.

## Before / after (same guest, fresh boot each, connections = c)

| c | before tps | after tps | delta |
|---|---|---|---|
| 8 | 68543 | 64294 | -6% |
| 16 | 28113 | 31892 | +13% |
| 32 | 17015 | 26075 | +53% |
| 48 | 14341 | 24692 | +72% |
| 64 | 13139 | 22698 | +73% |

It trades ~6% at low concurrency (extra migration traffic) for the tail no longer collapsing. Measured effect on placement: idle time 93.5% -> 89.7%, worker spread 4 -> 6 CPUs.

## Scope / honest note

General scheduler change, no application specifics. It does not fully saturate the machine: even after the fix the guest is still substantially idle at high concurrency, because the remaining limit is per-request round-trip latency through the single network receive path and per-packet wakeup IPIs, which grows with concurrency. That is a separate, larger change (batching wakeups) and is not attempted here; this PR removes the placement collapse, which is the dominant effect.

### DIFF
--- a/core/sched.cc
+++ b/core/sched.cc
@@ -25,6 +25,7 @@
 #include <osv/wait_record.hh>
 #include <osv/preempt-lock.hh>
 #include <osv/app.hh>
+#include <osv/kernel_config_sched_load_balance_ms.h>
 #include <osv/symbols.hh>
 #include <osv/stubbing.hh>
 #include <algorithm>
@@ -792,54 +793,89 @@ void thread::unpin()
     helper->join();
 }
 
+// The load balancer is the only mechanism that spreads threads across CPUs:
+// thread::start() places a new thread on its creator's CPU and thread::wake()
+// re-wakes a blocked thread on the CPU it last ran on, so neither disperses
+// work by itself.  Running once every 100ms and migrating a single thread per
+// wakeup is far too slow for a workload that fans out many threads from one
+// parent (e.g. a server that forks or spawns a worker per connection): the
+// workers pile onto the few CPUs they were started/woken on while the rest of
+// the machine sits idle, and a 1-thread-per-100ms drip cannot catch up under a
+// high request rate.  So this runs more often (the interval is configurable via
+// CONF_sched_load_balance_ms), and on each pass keeps migrating the
+// most-migratable queued thread to the least-loaded CPU until this CPU is no
+// longer meaningfully more loaded than that CPU -- the load difference itself
+// is the terminator, so a pass drains the whole imbalance (which can be larger
+// than the CPU count) rather than a single thread of it.
 void cpu::load_balance()
 {
     notifier::fire();
     timer tmr(*thread::current());
     while (true) {
-        tmr.set(osv::clock::uptime::now() + 100_ms);
+        tmr.set(osv::clock::uptime::now() + std::chrono::milliseconds(CONF_sched_load_balance_ms));
         thread::wait_until([&] { return tmr.expired(); });
         if (runqueue.empty()) {
             continue;
         }
-        auto min = *std::min_element(cpus.begin(), cpus.end(),
-                [](cpu* c1, cpu* c2) { return c1->load() < c2->load(); });
-        if (min == this) {
-            continue;
-        }
-        // This CPU is temporarily running one extra thread (this thread),
-        // so don't migrate a thread away if the difference is only 1.
-        if (min->load() >= (load() - 1)) {
-            continue;
-        }
-#if CONF_lazy_stack_invariant
-        assert(!thread::current()->is_app());
-#endif
-        WITH_LOCK(irq_lock) {
-            auto i = std::find_if(runqueue.rbegin(), runqueue.rend(),
-                    [](thread& t) { return t._migration_lock_counter == 0; });
-            if (i == runqueue.rend()) {
-                continue;
+        // Each thread we migrate this pass is queued on its destination CPU but
+        // has not yet been dequeued there, so cpu::load() (runqueue.size()) does
+        // not yet reflect it.  Track the migrations we have queued per CPU so
+        // both the least-loaded selection and the stop condition use an
+        // effective load (current load + pending arrivals); otherwise the same
+        // destination would keep looking least-loaded and we would pile the
+        // whole imbalance onto one CPU instead of spreading it.
+        std::vector<unsigned> pending(cpus.size(), 0);
+        // The load difference is the real terminator (see below); this only
+        // caps a pathological storm if loads never converge.  It is generous
+        // (many times the CPU count) so a genuinely large imbalance still
+        // drains in one pass.
+        unsigned migrated = 0;
+        const unsigned max_migrations_per_pass = cpus.size() * 8;
+        while (migrated < max_migrations_per_pass) {
+            auto min = *std::min_element(cpus.begin(), cpus.end(),
+                    [&pending](cpu* c1, cpu* c2) {
+                        return c1->load() + pending[c1->id] <
+                               c2->load() + pending[c2->id];
+                    });
+            if (min == this) {
+                break;
             }
-            auto& mig = *i;
-            trace_sched_migrate(&mig, min->id);
-            runqueue.erase(std::prev(i.base()));  // i.base() returns off-by-one
-            // we won't race with wake(), since we're not thread::waiting
-            assert(mig._detached_state->st.load() == thread::status::queued);
-            mig._detached_state->st.store(thread::status::waking);
-            mig.suspend_timers();
-            mig._detached_state->_cpu = min;
-            // Convert the CPU-local runtime measure to a globally meaningful
-            // measure
-            mig._runtime.export_runtime();
-            mig.remote_thread_local_var(::percpu_base) = min->percpu_base;
-            mig.remote_thread_local_var(current_cpu) = min;
-            mig.stat_migrations.incr();
-            min->incoming_wakeups[id].push_back(mig);
-            min->incoming_wakeups_mask.set(id);
-            // FIXME: avoid if the cpu is alive and if the priority does not
-            // FIXME: warrant an interruption
-            min->send_wakeup_ipi();
+            // This CPU is temporarily running one extra thread (this thread),
+            // so don't migrate a thread away if the difference is only 1.
+            if (min->load() + pending[min->id] >= (load() - 1)) {
+                break;
+            }
+#if CONF_lazy_stack_invariant
+            assert(!thread::current()->is_app());
+#endif
+            WITH_LOCK(irq_lock) {
+                auto i = std::find_if(runqueue.rbegin(), runqueue.rend(),
+                        [](thread& t) { return t._migration_lock_counter == 0; });
+                if (i == runqueue.rend()) {
+                    break;
+                }
+                auto& mig = *i;
+                trace_sched_migrate(&mig, min->id);
+                runqueue.erase(std::prev(i.base()));  // i.base() returns off-by-one
+                // we won't race with wake(), since we're not thread::waiting
+                assert(mig._detached_state->st.load() == thread::status::queued);
+                mig._detached_state->st.store(thread::status::waking);
+                mig.suspend_timers();
+                mig._detached_state->_cpu = min;
+                // Convert the CPU-local runtime measure to a globally meaningful
+                // measure
+                mig._runtime.export_runtime();
+                mig.remote_thread_local_var(::percpu_base) = min->percpu_base;
+                mig.remote_thread_local_var(current_cpu) = min;
+                mig.stat_migrations.incr();
+                min->incoming_wakeups[id].push_back(mig);
+                min->incoming_wakeups_mask.set(id);
+                // FIXME: avoid if the cpu is alive and if the priority does not
+                // FIXME: warrant an interruption
+                min->send_wakeup_ipi();
+                pending[min->id]++;
+            }
+            migrated++;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


core/sched.cc
```diff
@@ -797,20 +797,38 @@ void cpu::load_balance()
...
+        // in one pass instead of one thread of it.
+        tmr.set(osv::clock::uptime::now() + 10_ms);
         thread::wait_until([&] { return tmr.expired(); });
         if (runqueue.empty()) {
             continue;
         }
+        // Bound the work per pass so a transient spike cannot become an
+        // unbounded migration storm.
+        unsigned migrated = 0;
+        const unsigned max_migrations_per_pass = cpus.size();
```
wkozaczuk: Does it mean that we are capped to migrate a maximum equal to the number of cpus? If so, why the number of cpus? The comment above claims something different: "draining the whole imbalance in one pass instead of one thread of it". And I think the imbalance may be more than as many threads as there are cpus?

core/sched.cc
```diff
@@ -797,20 +797,38 @@ void cpu::load_balance()
...
+        while (migrated < max_migrations_per_pass) {
```
wkozaczuk, line 834: I think there is an indentation issue.

core/sched.cc
```diff
@@ -797,20 +797,38 @@ void cpu::load_balance()
...
+        // migrating a single thread per wakeup is far too slow for a workload
+        // that fans out many threads from one parent (e.g. a server that forks
+        // or spawns a worker per connection): the workers pile onto the few
+        // CPUs they were started/woken on while the rest of the machine sits
+        // idle, and a 1-thread-per-100ms drip cannot catch up under a high
+        // request rate.  Run more often, and on each pass keep migrating the
+        // most-migratable queued thread to the least-loaded CPU until this CPU
+        // is no longer meaningfully more loaded -- draining the whole imbalance
+        // in one pass instead of one thread of it.
+        tmr.set(osv::clock::uptime::now() + 10_ms);
```
wkozaczuk: Should this become a parameter in the `conf/kconfig`?

wkozaczuk (commented):
I have left some comments. Asked @nyh to review as I think he is more familiar with the scheduler.

core/sched.cc
gburd: Good catch, the comment and the cap were saying different things. I have reworked this so the loop terminator is the load *difference* itself, not a count: the pass keeps migrating the most-migratable queued thread to the least-loaded CPU while this CPU is meaningfully more loaded than that CPU. So a pass now genuinely drains the whole imbalance in one go, even when the imbalance is larger than the number of CPUs. The per-pass bound remains only as a backstop against a pathological case where loads never converge, and I made it generous (a multiple of the CPU count) so it is not the normal stopping condition. The comment is rewritten to describe this honestly.

core/sched.cc
```diff
@@ -797,20 +797,38 @@ void cpu::load_balance()
...
             continue;
         }
+        // Bound the work per pass so a transient spike cannot become an
+        // unbounded migration storm.
+        unsigned migrated = 0;
+        const unsigned max_migrations_per_pass = cpus.size();
+        while (migrated < max_migrations_per_pass) {
         auto min = *std::min_element(cpus.begin(), cpus.end(),
                 [](cpu* c1, cpu* c2) { return c1->load() < c2->load(); });
         if (min == this) {
```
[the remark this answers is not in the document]
gburd: Addressed. A thread we migrate is queued on its destination CPU but not yet dequeued there, so cpu::load() (runqueue.size()) does not yet reflect it, exactly as you note. The pass now tracks the migrations it has already queued per CPU and both the min_element selection and the stop condition use an effective load (current load + pending arrivals). That stops one CPU from repeatedly looking least-loaded and absorbing the whole imbalance.

core/sched.cc
gburd: Fixed the indentation of the inner loop body in the latest revision.

core/sched.cc
gburd: Done. The interval is now a kconfig option, CONF_sched_load_balance_ms (int, default 10ms), with help text describing the tradeoff (smaller disperses a fan-out faster, larger reduces balancing overhead on workloads that do not fan out). core/sched.cc uses std::chrono::milliseconds(CONF_sched_load_balance_ms) via the generated per-option header.

core/sched.cc
```diff
@@ -797,20 +797,38 @@ void cpu::load_balance()
     notifier::fire();
     timer tmr(*thread::current());
     while (true) {
-        tmr.set(osv::clock::uptime::now() + 100_ms);
+        // This balancer is the only mechanism that spreads threads across CPUs:
```
nyh: I think if you want a long comment explaining what cpu::load_balance() is, it should be next to the function (in sched.cc or sched.hh), not in the middle of its implementation.

gburd: Good point - moved the explanatory block up to be a header comment right before void cpu::load_balance(), so it describes the function as a whole rather than sitting in the middle of the loop. The two short comments left inline (the per-CPU pending[] tracking and the migration-storm cap) are about specific local mechanics at those points, so I kept them where they are. sched.o compiles clean.

core/sched.cc
```diff
@@ -792,54 +793,89 @@ void thread::unpin()
...
+        // not yet reflect it.  Track the migrations we have queued per CPU so
+        // both the least-loaded selection and the stop condition use an
+        // effective load (current load + pending arrivals); otherwise the same
+        // destination would keep looking least-loaded and we would pile the
+        // whole imbalance onto one CPU instead of spreading it.
+        std::vector<unsigned> pending(cpus.size(), 0);
+        // The load difference is the real terminator (see below); this only
+        // caps a pathological storm if loads never converge.  It is generous
+        // (many times the CPU count) so a genuinely large imbalance still
+        // drains in one pass.
```
nyh, line 831: I'm not sure this situation (and the code and comments needed to excuse it) is worth defending against. If someone has a ridiculous loop spawning 10000 threads, do we really care if the load balancer moves all of them in one loop? What's the downside?

nyh (approved):
I'm tentatively approving, but I also a bit worried:

My experience is that scheduling policy changes always have cases where they are perfect, while in other cases, they actually make things worse.
I definitely agree that in the case you mentioned - an (old style...) threaded server that always listen()s on one thread and then creates a brand new thread to handle the request, you can have 100 new threads in a 100ms period, and shouldn't move just one thread.

However, I'm not convinced that this patch won't make things worse for **other** workloads, which do not involve the (inherently inefficient!) workload of creating many new threads from a single parent threads:

Consider for example a hypothetical workload, where you have 10 CPUs, and at some unlucky moment, 9 of them have 10 running threads, but in one of them, none of the threads are running (maybe they are waiting for IO or something). After this patch, it is possible that **each** of the 9 CPUs will send half its running threads to that same target CPU, and the result will be 9 CPUs with 5 running threads each plus one CPU with 45 runnings thread! :-(

The old code "solved" this problem by just being slow to react - you couldn't get 50 threads being shipped over in 10 milliseconds because the code could only ship 10 threads (one per CPU) in a period of 100 milliseconds.

In general, a controller can suffer from big fluctuations if you let it react too quickly and you often see things like "momentum" or "hysteresis" trying to avoid violent jumps of dozens of threads back and forth between CPUs.

In https://github.com/cloudius-systems/osv/issues/396 I mentioned another idea I tried in he past. I noticed that the worst situation is to have idle CPUs while other CPUs have threads waiting to run. I solved this by adding to the idle thread code which finds a highly loaded CPU and wakes up its load balancer thread - hoping it will send some work to this idle CPU. It improved my tests, and will also help in your case (the idle CPUs will cause the listener thread's CPU to run its load balancer frequently) but might not be as efficient as your patch.

gburd:
Thanks, and this is the right worry to raise. You are describing the classic controller-overshoot failure, and I agree it is the real risk with reacting faster.

Two things are already in the patch to damp exactly your "9 CPUs each ship half their threads to the one idle CPU" case, and one thing is deliberately deferred to a separate PR:

1. Intra-pass damping (in this patch). Within a single balancer pass this CPU does not keep choosing the same victim target. Each thread it queues to a destination increments a per-destination `pending[]` counter, and both the least-loaded selection and the stop condition use effective load = `load() + pending[dest]`. So a single overloaded CPU cannot dump its whole runqueue onto one idle CPU: after it sends a few, that target stops looking least-loaded and the remainder spreads across the next-least-loaded CPUs. The stop condition is the load difference itself (`min->load()+pending >= load()-1`), so a pass converges rather than emptying blindly; the `cpus.size()*8` cap is only a pathological-storm backstop, not the normal terminator.

2. The cross-balancer race you describe is the residual. `pending[]` is per-pass and CPU-local, so it does not stop nine *different* balancer threads from independently eyeing the same idle CPU inside the same time window. What bounds that today is that each balancer re-reads the live `load()` (which does reflect migrations other balancers have already dequeued onto that target) and the `load()-1` guard, so the target fills up across passes rather than all-at-once; but I agree it is weaker than the intra-pass guard, and a hostile phase alignment could still overshoot for one interval before the next pass corrects it. If you would prefer, the cheap hardening is a small hysteresis / per-target global reservation (an atomic pending count on the destination CPU that every balancer consults), which turns the cross-balancer case into the same effective-load logic as the intra-pass case. I am happy to add that here if you want it before merge.

3. The idle-side half from #396 is #1472. Your idea in #396 (have an idle CPU wake the load balancer of a loaded CPU) is implemented as a separate PR, #1472 (idle-CPU pull work-stealing). That is the complementary half: instead of only pushing from the loaded side, an idle CPU pulls. Running them together is what actually dispersed a many-connection network workload in my testing, and the pull side naturally avoids the overshoot you describe because the idle CPU takes work for itself only.

So: the single-parent-fanout case is handled by this PR, the idle-CPU case by #1472, and if you want the cross-balancer overshoot closed defensively rather than relying on it self-correcting over a pass, say the word and I will add the per-target reservation here.

gburd:
Fair question, and it is the right one to ask directly rather than defend the machinery.

Answer to "what's the downside of just moving all of them in one loop": for the pathological 10000-thread case, honestly nothing that matters, and I do not think it is worth writing code to defend against that case specifically. The `cpus.size()*8` cap is not there to protect the 10000-thread loop; it is only a loop backstop so a single balancer pass has a bounded upper iteration count and cannot spin arbitrarily long while other CPUs are concurrently mutating their runqueues underneath it. It is a "this pass terminates" guard, not a "this workload is abusive" policy.

The real terminator is the load-difference stop condition (`min->load()+pending >= load()-1`): a pass naturally stops when this CPU is no longer meaningfully more loaded than the least-loaded target, which for the honest fan-out case (one parent spawns N children) is exactly the behavior you want and is reached long before the cap.

So I agree with the spirit of your comment. The cap and its explanatory comment are the part carrying their weight poorly. I can drop the `cpus.size()*8` cap and the paragraph excusing it, and rely solely on the load-difference stop condition, which is simpler and is the thing actually doing the work. That removes the code-and-comment-to-excuse-it that you flagged. If you would rather keep a trivial iteration bound purely as a runaway-loop safety (not a policy), I can leave a bare `runqueue.size()`-derived bound with a one-line comment instead of the current justification. Tell me which you prefer and I will push it here.

core/sched.cc
gburd: Answering in-thread, since my reply to this landed only as a top-level comment
(https://github.com/cloudius-systems/osv/pull/1464#issuecomment-5466541442) and
so never showed up as a reply here.

Short version: I agree with you, and nothing is being defended against.

The `cpus.size() * 8` cap is not there for the 10000-thread loop. It is only a
"this pass terminates" bound so one balancer pass has a finite iteration count
while other CPUs concurrently mutate their runqueues underneath it. It is not a
policy about abusive workloads, and for the honest fan-out case it is never
reached: the real terminator is the load-difference condition
`min->load() + pending >= load() - 1`, which stops the pass once this CPU is no
longer meaningfully more loaded than the least-loaded target.

So the part carrying its weight poorly is the cap plus the paragraph excusing
it, which is exactly what you flagged. I am happy to drop both and rely solely
on the load-difference stop condition, or to keep a bare iteration bound with a
one-line "runaway-loop safety, not policy" comment. Since you have already
approved, I did not push that change unasked; say which you prefer and I will
push it here.

PR #1468, opened 2026-08-04
virtio-net: keep the RX mbuf refcount inside the receive buffer

gburd:
The virtio-net receiver allocated a separate `unsigned` refcount per packet (`new unsigned` in `packet_to_mbuf`, freed via the EXT_EXTREF cleanup) for every mbuf built from an RX buffer. That is a heap alloc/free pair on the receive hot path.

This reserves `sizeof(unsigned)` at the tail of each RX buffer (space never advertised to the device in the available ring) and points the mbuf's `ref_cnt` at it, so the refcount rides in the buffer itself. No per-packet allocation, no new allocator machinery. The buffer is freed as before once the refcount drops.

2 files (drivers/virtio-net.cc/.hh). A straightforward hot-path allocation removal.

### DIFF
--- a/drivers/virtio-net.cc
+++ b/drivers/virtio-net.cc
@@ -561,13 +561,18 @@ void net::receiver()
     }
 }
 
+// The mbuf external-storage reference count lives inside the receive buffer
+// itself: fill_rx_ring() reserves sizeof(unsigned) bytes at the tail of each
+// buffer (see rx_buffer_refcnt()) and never advertises them to the device, so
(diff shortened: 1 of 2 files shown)


wkozaczuk:
This looks pretty clever. Is it applicable to other drivers like vmxnet or ena?

Do you see any performance improvement?

wkozaczuk:
I have run all unit tests and stress-tested golang http server.

PR #1469, opened 2026-08-07
tests: union-merge the tests module Makefile to avoid list conflicts

gburd:
The tests module Makefile keeps a single append-only `tests` list that many independent feature branches extend, each adding its own test. Because every addition edits the same contiguous region, concurrent PRs conflict on it: as soon as one merges, the others need a manual resolution (and an extra merge commit) for what is really an independent append.

This marks the file `merge=union` in `.gitattributes` so git resolves these additions by keeping both sides instead of raising a conflict.

Combined with one-line `tests += tst-foo.so` additions (rather than splicing into the list literal), concurrent test-adding PRs merge cleanly in any order with no manual work. Verified: three branches each appending a test line merge clean in sequence and `make` still expands `tests` to the full list.

Suggested to merge this first; I will then convert my open test-adding PRs to the one-line `+=` form so they stop conflicting on this file.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+# The tests module Makefile is an append-only list that many independent
+# feature branches extend (each adds its own test). Union-merge it so those
+# additions never conflict: git keeps both sides instead of raising a conflict
+# that would otherwise need a manual resolution (and an extra merge commit) for
+# every concurrent PR. Pair with one-line "tests += tst-foo.so" additions so the
+# merged result stays valid.
+modules/tests/Makefile merge=union


gburd:
Hi Waldek,

Sorry about this, and thanks for flagging it. The conflicts are my fault: almost
all of my open PRs add their test by splicing a line into the single `tests :=`
list in `modules/tests/Makefile`, so every one of them edits the same region.
As soon as one merges, the rest conflict there, and resolving by hand adds the
extra merge commit you saw. I will do all the mechanical work so this costs you
no time.

Two-part fix, tested:

1. I opened #1469, a one-line change marking `modules/tests/Makefile` as
   `merge=union` in `.gitattributes`. With that, git auto-resolves these
   append conflicts by keeping both sides: no conflict markers, no manual
   resolution, no extra commit. If you merge #1469 first, the rest stop
   conflicting on that file.

2. I am converting each of my open test-adding PRs to append its test on its own
   line (`tests += tst-foo.so`) instead of splicing into the list literal, and
   rebasing them onto current master. Self-contained `+=` lines merge cleanly
   under union and keep the Makefile valid (verified: several branches merging
   in sequence stay clean and `make` still expands `tests` to the full list).

After that you should be able to merge them in any order without touching the
Makefile or adding merge commits. I will reply on each PR once it is rebased and
clean so you can see at a glance which are ready.

If you would rather a different convention (one `.mk` fragment per test, an
alphabetized list, etc.), say the word and I will redo them to match, but I did
not want to make you wait, so I am proceeding with the union + one-line-`+=`
approach unless you tell me otherwise.

Greg

gburd:
Quick heads-up on the test-adding PRs (follow-up to #1469): GitHub may show some
of them with a red "conflicting" label on `modules/tests/Makefile` even though
they merge cleanly. GitHub's mergeability preview does not apply the
`.gitattributes merge=union` rule from #1469, so its preview still sees a textual
conflict on the `tests` list, but the actual merge (which does honor merge=union)
resolves it automatically, keeping both test entries, no markers, no extra
commit.

I verified this by test-merging each of my open test-adding PRs onto current
master with the union rule in place: all merge clean and `make` still expands the
full `tests` list. So please don't let the red label make you skip them, they
should merge as smoothly as #1431/#1433/#1469 did. If any actually conflict on
merge, tell me and I'll rebase it right away.

Thanks again for merging these.

wkozaczuk:
I have merged one of them locally and pushed. But the history is
now convoluted with my extra merge commits. I am not sure how to fix this.

PR #1470, opened 2026-08-08
virtio-net: implement RSS (VIRTIO_NET_F_RSS) receive flow steering

gburd:
Stacked on #1463 (multiqueue). Adds RSS receive flow steering so incoming connections are hashed across the multiqueue RX queues implemented in #1463, letting a high-connection-count workload spread receive processing over multiple CPUs.

Measured (stock PostgreSQL, read-only OLTP benchmark, on a bare-metal KVM host): with the multiqueue + RSS substrate active, mid-range concurrency throughput improves (c16 ~+34%, c32 ~+16%, c48 ~+12%) versus a single RX queue. This is a generic net-path improvement that benefits any high-connection-count server; PostgreSQL is just the measurement workload.

Pending #1463 review, since RSS builds on the multiqueue infrastructure there.

### DIFF
--- a/arch/x64/exceptions.cc
+++ b/arch/x64/exceptions.cc
@@ -118,7 +118,12 @@ unsigned interrupt_descriptor_table::register_interrupt_handler(
             }
         }
     }
-    abort();
+    // No free IDT vector.  Vectors 0..31 are CPU exceptions and are never
+    // returned here, so 0 is an unambiguous "exhausted" sentinel.  Returning it
+    // rather than aborting lets MSI-X allocation (msix_vector) fail gracefully
+    // and lets a device fall back to fewer queues instead of crashing the boot
+    // when many multiqueue devices exhaust the 224 usable vectors.
+    return 0;
 }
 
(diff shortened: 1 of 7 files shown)


gburd:
## Follow-up: RSS validated alongside multiqueue

With the downstream dispatch bottleneck fixed (idle-CPU pull, #1472) the multiqueue receive path (#1463) plus RSS flow steering now shows a measured, repeatable gain on a many-connection network workload driven over a real NIC from an external client.

RSS hashes inbound flows across the receive queues so that distinct connections land on distinct receiver threads rather than all funneling through one. Under concurrent load the receive work spreads (7 of 8 CPUs busy versus the single-receiver funnel), and mid-to-tail throughput rises ~20-25% across the higher concurrency levels, 0-failed. Without RSS the flows can pile onto one queue even when several exist; with it, N concurrent connections spread across N CPUs.

**Dependency:** this stacks on multiqueue (#1463) which stacks on the MSI-X vector-budget fix (#1465); they should land in that order. Marking ready for review on that basis.

PR #1475, opened 2026-08-17
mm: resolve fork COW write faults lock-free when nested (fork-stack follow-on to #1456)

gburd:
**DRAFT / fork-stack follow-on -- not a standalone master change.**

This is a one-commit follow-on to the opt-in `fork()` copy-on-write work
(#1456), and belongs on that stack, tracked under #1458. It is **not**
proposable against stock master on its own: the code it touches
(`handle_cow_write_fault` and the nested COW branch of `vm_fault`) only
exists under `#if CONF_fork`, which lands with #1455/#1456. On a
`conf_fork=0` build the change compiles out entirely and the binary is
byte-identical.

The reviewable unit here is the single top commit `mm: resolve fork COW
write faults lock-free when nested ...` touching `core/mmu.cc` only. The
other commits shown in the diff are the #1456 COW base this stacks on; they
appear because master does not yet carry that base. Review just the top
commit.

### What it fixes

A copy-on-write write fault can be taken while another exception is already
in flight -- e.g. a device/timer IRQ lands a fault while a COW fault is
already being serviced during a `fork()` -- driving the fault handler to
`exception_depth >= 2`. At that depth no reschedule is allowed:
`reschedule_from_interrupt` asserts `exception_depth <= 1` (the per-CPU
nested-exception stacks are finite).

The normal COW path reschedules in two places, either of which trips that
assert deep in the fork path under concurrency:

- it takes the AS-wide `vmas_mutex->for_write()` -- a contended rwlock that
  blocks (and therefore reschedules), and
- `handle_cow_write_fault` calls `flush_tlb_all()`, which takes
  `tlb_flush_mutex` and `wait_until()`s every other CPU (also a reschedule).

### The change

When the COW write fault is itself nested (`exception_depth >= 2`), resolve
it lock-free: `handle_cow_write_fault` already walks only this address
space's own page tables and installs the private page with a single PTE
write to this AS, so a **local** TLB flush suffices (the page is private to
this AS; other CPUs run a different CR3 and reload on switch-in). Nothing
blocks, so no reschedule can happen at `exception_depth >= 2`. A concurrent
COW copy of the same page by another CPU at most duplicates a private page
(the loser leaks one page -- rare, only in this nested race, and
correctness-preserving).

The non-nested (`exception_depth <= 1`) path is byte-for-byte unchanged.

### Why it matters

Beyond the crash-safety (it removes an assert that fires deep in the fork
path under concurrent load), it also removes a latency serialization for
forked-process servers: a backend blocking on the AS-wide `vmas_mutex` and a
global cross-CPU TLB flush on the nested COW write-fault path serializes
concurrent forked-process workloads. Resolving the nested fault lock-free
with a local flush lets those backends make progress independently.

### Scope / stacking

- Depends on the `CONF_fork` COW machinery from #1456 (and the fork gate
  #1455); tracked under #1458.
- Entirely under `#if CONF_fork`; `conf_fork=0` byte-identical.
- Single commit, `core/mmu.cc` only (+44 / -3).

Draft while the fork stack is gated on #1455 -- opened so the fix is banked
reviewably on the right stack, not to request near-term merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1057,6 +1057,9 @@ objects += arch/$(arch)/firmware.o
 objects += arch/$(arch)/hypervisor.o
 objects += arch/$(arch)/interrupt.o
 objects += arch/$(arch)/clone.o
+ifeq ($(conf_fork),1)
+objects += arch/$(arch)/fork.o
+endif
 ifeq ($(conf_drivers_pci),1)
 objects += arch/$(arch)/pci.o
 objects += arch/$(arch)/msi.o
@@ -1660,11 +1663,20 @@ musl += prng/srand48.o
 libc += random.o
 
 libc += process/execve.o
+ifeq ($(conf_fork),1)
+libc += process/fork.o
+endif
 musl += process/execle.o
 musl += process/execv.o
 musl += process/execl.o
 libc += process/waitpid.o
+# When fork() is enabled, our waitpid.cc provides wait()/waitpid()/wait4()
+# backed by the fork() child registry, superseding musl's process/wait.o.
+# When fork() is disabled, waitpid.cc provides only the ECHILD waitpid() stub,
+# so keep musl's wait.o for wait()/wait4().
+ifneq ($(conf_fork),1)
 musl += process/wait.o
+endif
 
 musl += setjmp/$(musl_arch)/setjmp.o
 musl += setjmp/$(musl_arch)/longjmp.o
--- a/arch/aarch64/fork.cc
+++ b/arch/aarch64/fork.cc
@@ -22,8 +22,9 @@
 extern "C" void __osv_run_atfork_child();
 
 sched::thread *fork_thread(void *caller_ret, void *caller_sp,
-                           void **out_stack_to_free)
+                           void *resume_ctx, void **out_stack_to_free)
 {
+    (void)resume_ctx;   // aarch64 uses caller_sp bias-copy (same-VA is x86-64)
     auto parent = sched::thread::current();
     auto parent_pinned_cpu = parent->pinned() ? sched::cpu::current() : nullptr;
 
--- a/arch/aarch64/fork.cc
+++ b/arch/aarch64/fork.cc
@@ -69,7 +69,15 @@ sched::thread *fork_thread(void *caller_ret, void *caller_sp,
            "br %1 \n\t"         // resume in fork()'s caller
            : : "r"(resume_sp), "r"(resume_pc) : "x0", "memory");
     }, sched::thread::attr().
-        stack(4096 * 4),
+        stack(4096 * 4).
+        // Detached: nobody join()s the fork child (the parent reaps it via the
+        // pid registry / waitpid, not sched::thread::join).  A detached thread
+        // is handed to the reaper on completion, which runs our set_cleanup()
+        // (freeing the copied stack and disposing the thread object, releasing
+        // its application_runtime reference).  Without this the thread object
+        // (and its app_runtime shared_ptr) would leak and OSv would hang at
+        // shutdown -- see the cleanup comment in libc/process/fork.cc.
+        detached(),
         false,
         true);
     t->set_app_tcb(parent->get_app_tcb());
--- a/arch/aarch64/fork.cc
+++ b/arch/aarch64/fork.cc
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2026 Greg Burd
+ *
+ * This work is open source software, licensed under the terms of the
+ * BSD license as described in the LICENSE file in the top-level directory.
+ *
+ * fork_thread() for aarch64 -- mirrors arch/x64/fork.cc.  fork() (fork.cc)
+ * passes the caller's return address and stack pointer; we copy the parent's
+ * user stack, bias the SP into the copy, and start a child thread that installs
+ * the copied stack, sets x0=0 (fork()'s return value in the child), and returns
+ * to fork()'s caller.
+ */
+
+#include "arch.hh"
+#include <errno.h>
+#include <string.h>
+#include <cstdlib>
+#include <osv/sched.hh>
+
+// pthread_atfork child-handler chain (defined in libc/pthread.cc), run in the
+// child's context before it resumes user code.
+extern "C" void __osv_run_atfork_child();
+
+sched::thread *fork_thread(void *caller_ret, void *caller_sp,
+                           void **out_stack_to_free)
+{
+    auto parent = sched::thread::current();
+    auto parent_pinned_cpu = parent->pinned() ? sched::cpu::current() : nullptr;
+
+    auto si = parent->get_stack_info();
+    char *stack_base = static_cast<char*>(si.begin) + si.size;
+    char *sp = static_cast<char*>(caller_sp);
+    if (sp < static_cast<char*>(si.begin) || sp > stack_base) {
+        return nullptr;
+    }
+    size_t stack_size = si.size;
+
+    char *child_stack_mem = static_cast<char*>(malloc(stack_size));
+    if (!child_stack_mem) {
+        return nullptr;
+    }
+    // Copy ONLY the live top [caller_sp .. stack_base) into the top of the
+    // child buffer (app stacks are demand-paged; copying from si.begin faults).
+    char *child_base = child_stack_mem + stack_size;
+    ptrdiff_t bias = child_base - stack_base;
+    size_t live = static_cast<size_t>(stack_base - sp);
+    memcpy(child_base - live, sp, live);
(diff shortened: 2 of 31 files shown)


gburd:
## Soundness question on the nested local-flush fast path

Reviewing this draft against its own stated justification, I think the `exception_depth >= 2` local-flush branch is unsound for the **parent** address space, and I would like to check my reading before you act on it. The problem is not the reschedule-avoidance goal, which is clearly right, but the specific premise the comment uses to license a local-only flush.

All line numbers below are this PR's head `0b72ee0c6581b653d5024cc7e55942c7536b4aed`.

### The premise

`core/mmu.cc:2022-2026`:

```c
    // is not needed for correctness here: the page was just made PRIVATE to
    // THIS address space (a fresh phys page in this AS's own tables); only the
    // faulting CPU can hold a stale write-protected TLB entry for addr in this
    // AS, and other CPUs run a different CR3 and reload the full context on
    // switch-in.  So when nested do the non-blocking LOCAL flush; otherwise
```

guarding `core/mmu.cc:2029-2033`:

```c
    if (sched::exception_depth >= 2) {
        mmu::flush_tlb_local();
    } else {
        mmu::flush_tlb_all();
    }
```

The premise has two parts: (1) only the faulting CPU can hold a stale entry for `addr` in this AS, and (2) other CPUs run a different CR3.

### Why part (2) does not hold for the parent

`clone_pt_level0` write-protects the **parent's** PTE, not only the child's, at `core/mmu.cc:320-325`:

```c
        } else if (ppte.writable() && !pte_is_cow(ppte)) {
            // Private writable page: make it COW (write-protect + cow bit) in
            // BOTH parent and child.
            pt_element<0> cow = pte_mark_cow(ppte, true);
            parent_pt[i] = cow;
            child_pt[i] = cow;
```

So after any `fork()`, the parent's own private writable anonymous pages are write-protected + COW-tagged in the parent's page tables. The parent then takes COW write faults on its own pages, and those faults reach the same `handle_cow_write_fault` through `core/mmu.cc:2084-2089`, which resolves against `t->address_space()` -- the parent's AS.

The parent AS is AS0 (`kernel_as`, `core/mmu.cc:187-192`), and AS0 is shared by every thread that has not been reassigned by `fork()`. `thread::thread` inherits the creator's AS (`core/sched.cc:1094`), so in a multi-threaded process every thread is in AS0. Concretely: N application threads spread across N CPUs are all running with the **same** CR3. "Other CPUs run a different CR3" is false for AS0 by construction, and AS0 is exactly the AS in which the parent's post-fork COW faults are resolved.

That also breaks part (1): sibling threads on other CPUs in the same AS can and do touch the same VAs, so they can hold cached entries for `addr`.

### Why the stale entry is not self-correcting

This is the point I expect pushback on, so to be explicit about the failure mode.

The stale entry left on another CPU is **present and read-only**, not not-present. That distinction is the whole problem:

- A stale **not-present** entry would be benign here. The next access faults, the fault handler walks the current tables, finds the new PTE, and the entry self-heals. Losing an invalidation would cost one spurious fault, nothing more.
- A stale **present read-only** entry does not fault on **read**. The CPU has a valid translation and uses it. Nothing in the hardware path consults the page tables again, so nothing notices that the PTE now points at a different physical page.

So the sequence is a silent stale read, not a fault:

1. Thread A on CPU0 takes a nested COW write fault on VA `X` in AS0 (parent), `exception_depth >= 2`.
2. A allocates a fresh physical page `P_new`, `memcpy`s `P_old` into it (`core/mmu.cc:2004-2006`), and installs `P_new` writable in AS0's tables (`core/mmu.cc:2009-2013`).
3. A calls `flush_tlb_local()` (`core/mmu.cc:2030`), invalidating only CPU0.
4. Thread B on CPU1 is in the **same** AS0 and still has `X -> P_old` cached read-only.
5. A writes to `X`. The write lands in `P_new`.
6. B reads `X`. The read is satisfied from the cached translation, so B reads `P_old` and does not observe A's write. B also reads a page that AS0's page tables no longer reference at that VA.

The result is not a crash but silent divergence: two threads in one address space disagreeing about the contents of one VA, for as long as the stale entry survives.

### Is there any path that masks it

I looked for something that would guarantee a timely CR3 reload on CPU1, since that is the obvious "it self-heals in practice" objection. I could not find one, and I think the fast path specifically removes the mechanism that would have provided it:

- **Same-AS context switch does not touch CR3.** `arch/x64/arch-switch.hh:114` computes `switch_as = (_current_as != old->_current_as)`, and the `mov %2, %%cr3` at line 148 is only in the `switch_as` branch. For two threads both in AS0, `switch_as` is false and the non-CR3 asm at lines 160-174 runs. So CPU1 can context-switch freely, many times, and never reload CR3.
- **The lazy-flush mechanism is only armed by `flush_tlb_all`.** `arch/x64/mmu.cc:94` is the only place that sets `c->lazy_flush_tlb`, and it is inside `flush_tlb_all`. The consumer at `core/sched.cc:402-404` therefore only fires for a CPU that a global flush marked. Taking the local branch skips the mark entirely, so the switch-in flush at `core/sched.cc:403` never runs for this invalidation.
- **No other CR3 write applies.** The only other `write_cr3` calls are inside `flush_tlb_local` itself (`arch/x64/mmu.cc:54`) and `switch_to_runtime_page_tables` (`arch/x64/mmu.cc:144`, boot only).

So the surviving window is bounded only by whatever unrelated event happens to reload CR3 on CPU1 or evict the TLB line: an unrelated `flush_tlb_all` from some other subsystem, a switch to a thread in a genuinely different AS, or natural TLB pressure. None of those is ordered against the write in step 5, and none is guaranteed to occur before B's read. The window is indefinite, and the failure is timing-dependent and silent, which is the hard kind to attribute later.

### Scope, and what I am not claiming

- **The `depth <= 1` path is correct.** `flush_tlb_all` at `core/mmu.cc:2032` is conservative and sound, including for AS0. This is only about the new branch.
- **`conf_fork=0` is unaffected**, as the PR says.
- **A genuinely single-threaded child AS is fine.** If an AS is provably live on at most one CPU, the local flush is sound for it. The bug is that the branch does not test that condition, and the parent AS violates it whenever the process has more than one thread.
- **I have not run this.** This is a source-level argument, not an observed corruption, and I have not built a reproducer. The mechanism is concrete enough to state as above, but "hazard demonstrated by construction, not yet demonstrated at runtime" is the honest label. A reproducer would want a multi-threaded parent, a `fork()`, and a nested COW write fault on a page a sibling thread reads, which is awkward to force deterministically precisely because it needs `exception_depth >= 2`.

### Options, your call

The guard needs one more condition: the flush may be local only when this AS cannot be live on another CPU. Some ways to get there:

1. **Only take the local branch for an AS known to be single-CPU.** A child AS created by `fork()` starts single-threaded, so a flag on `address_space` set at `clone_address_space` time and cleared the moment a second thread joins that AS would let children keep the fast path while AS0 always takes the safe path. Narrow, and preserves the benefit for the case the PR cares about.
2. **Refcount CPUs per AS.** More precise, more bookkeeping, and it has to be exactly right on the migration path to be worth it.
3. **Find a non-blocking broadcast.** The reason for the local flush is that `flush_tlb_all` reschedules; a send-IPI-without-waiting variant would invalidate remote CPUs without blocking. Correctness then depends on whether the fault can return before the IPIs are acknowledged, which needs its own argument.
4. **Drop the local-flush half and keep only the lock-free half.** The `vmas_mutex` avoidance at `core/mmu.cc:2085-2089` is the other half of this PR and does not depend on the flush change. If the reschedule at depth >= 2 comes mainly from the contended rwlock, that half may buy most of the benefit without the TLB question.

Option 1 looks like the smallest change that keeps the stated goal, but you know this stack better than I do.

### Secondary, and not on this PR's head

Worth flagging because it breaks the same "one AS, one CPU" assumption by an independent route, but **it is not present on this PR's head** -- I checked, and `libc/signal.cc` here has no `as_for_pid` and no `set_address_space`. On the integration branch (#1458, `integ/pg-fork-zfs`) `kill()` assigns a freshly created signal-handler thread into a *live fork child's* address space, `libc/signal.cc:718`:

```c
        t->set_address_space(target_as ? target_as : mmu::kernel_address_space());
```

with `target_as` coming from `osv::fork::as_for_pid(pid)` at `libc/signal.cc:559`. That deliberately makes a second thread, on a possibly different CPU, run in a child's AS, so on that branch even a fork child's AS is not reliably single-CPU. It is PostgreSQL's cross-backend latch wakeup path, so it runs in normal operation rather than only in a corner case. If the fix here takes the "AS is single-CPU" route (option 1), that route needs to account for this injection point before the integration branch adopts it. Filing it here only as context for choosing the fix, since the code is not in this PR.

PR #1483, opened 2026-08-27
pagecache: report page-writeback errors from sync() by return value

gburd:
## Summary

`pagecache::sync()` threw a C++ error when a page writeback (`cached_page_write::writeback()`) failed. Its only caller is the munmap / msync path: `vfs_file::sync()`, invoked from `mmu::file_vma::sync()`, which already converts the outcome to a return value and surfaces any error through its own `sys_fsync()` tail. The `fsync(2)` syscall uses a separate `VOP_FSYNC` path and never relied on this throw, so no caller depended on the exception.

Throwing from the munmap path is both unnecessary and fragile: raising an exception during `munmap()` can abort when the C++ unwinder cannot complete in the calling context, turning a recoverable page-writeback error into a fatal `abort()` that takes down the whole instance.

## Change

- `pagecache::sync()` returns `int` (0 on success, or the writeback errno) instead of throwing.
- `vfs_file::sync()` consumes the return value; `mmu::file_vma::sync()` continues to report the error via `sys_fsync()`.
- The declaration in `include/osv/pagecache.hh` is updated to match.

Behavior is unchanged on success. On a writeback error, `sync()` now returns the errno rather than throwing, and the page is left marked dirty for retry exactly as before.

### DIFF
--- a/core/pagecache.cc
+++ b/core/pagecache.cc
@@ -957,7 +957,15 @@ bool release(vfs_file* fp, void *addr, off_t offset, mmu::hw_ptep<0> ptep)
     return addr != zero_page;
 }
 
-void sync(vfs_file* fp, off_t start, off_t end)
+// Flush dirty MAP_SHARED pages for [start,end) back to the filesystem.
+// Returns 0 on success or the writeback errno.  Historically this threw a C++
+// error on writeback failure, but it is reached from munmap() (via
+// mmu::file_vma::sync) as well as from an explicit fsync path, and throwing on
+// the munmap path is both unnecessary (the caller converts the result to a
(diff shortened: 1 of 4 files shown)


gburd:
Added a companion commit: the base `file::sync()` (the virtual default in `include/osv/file.h`) also threw `make_error(ENOSYS)` and is reached from the same munmap page-writeback path (`mmu::file_vma::sync -> _file->sync()`). Unmapping a file-backed mapping whose file type does not override `sync()` therefore raised an exception during `munmap()` that cannot be unwound from that path and bottoms out in `abort()`. A file type with no cached writable pages has nothing to flush, so the default is now a no-op. Together with the pagecache change this removes all C++ throws from the munmap writeback path.

PR #1486, opened 2026-08-28
bsd/callout: treat a null callout in callout_reset_on as a no-op

gburd:
`callout_reset_on()` dereferences its callout pointer immediately. Under concurrency the FreeBSD ARP/link-layer path can reach `callout_reset()` on a link-layer entry (`la->la_timer`) that has raced with its own teardown, passing a null callout. The dereference then page-faults at address 0 deep in the network output path (`arpresolve` -> `ether_output` -> `tcp_output`), aborting the instance instead of harmlessly doing nothing.

Guard the entry: if the callout pointer is null there is nothing to arm, so return 0 (no pending callout deactivated), matching the contract that resetting an already-inactive callout is a no-op.

Observed under a high-connection-count network workload; the abort is intermittent and race-dependent. Standalone change, no dependencies.

### DIFF
--- a/bsd/porting/callout.cc
+++ b/bsd/porting/callout.cc
@@ -122,6 +122,15 @@ static void _callout_thread(void)
         // get the first callout with the earliest time
         callout *c = callouts::get_one();
 
+        // Defensive: get_one() can return null if the set drained after the
+        // wait_until predicate woke us (a concurrent stop/drain removed the
+        // last callout).  Guard against a null head rather than page-faulting
+        // at address 0 in the dispatcher and aborting the guest under heavy
+        // concurrency (ZFS txg timers + many TCP timers).
+        if (!c) {
+            continue;
+        }
+
         assert(c->c_flags & (CALLOUT_ACTIVE | CALLOUT_PENDING));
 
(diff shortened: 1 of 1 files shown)


gburd:
Added a second commit that fixes the **root cause** behind this crash, not just the symptom.

The dispatcher (`_callout_thread`) ran a callout's handler while the callout was still linked in the ordered set that `callout_compare` keys on. Handlers on the TCP and link-layer teardown paths free their callout, leaving the set holding a dangling pointer; the next `add_callout` insert then dereferenced freed memory and page-faulted at address 0 (the intermittent `arpresolve`/`tcp_output` crash). The first commit's null-guard only masked one manifestation.

The fix removes the callout from the set before dropping the lock to run the handler; if the handler reschedules via `callout_reset`, `add_callout` re-links it, so the post-handler `have_callout` check cleanly means rescheduled. `callout_drain` semantics are preserved (a drain waiter is always completed and woken, whether or not the handler re-armed).

Validated: a large-working-set PostgreSQL schema build whose index phase previously aborted here now completes and survives reboots.

PR #1487, opened 2026-08-28
zfs(openzfs): cap the default ARC at half of RAM for app coexistence

gburd:
**Draft / follow-on to #1423 and #1478.**

On OSv the ARC and the application share one physical address space. `arc_default_max()` for a guest with >= 1 GiB of RAM returns `MAX(allmem*5/8, allmem-1GiB)`, which on a large guest is nearly all of RAM (about 31 GiB of a 32 GiB guest). A memory-hungry application such as PostgreSQL (multi-GiB shared_buffers plus many concurrent backends) then has almost no headroom, and a checkpoint burst stacking dirty ARC buffers on top of the backends' working set drives the guest OOM and aborts it. This showed up as an OOM at high concurrency on a 32 GiB guest.

Since `ZFS_MODULE_PARAM` is a no-op on OSv (`zfs_arc_max` cannot be set at boot), this caps the computed default at half of RAM for guests with more than 2 GiB, so the application always keeps at least half the guest's memory. Small guests are unchanged.

Draft because the proper long-term fix is to wire `zfs_arc_max` to the boot cmdline so this becomes an overridable default rather than a hard cap; kept as a draft PR so the tuning is not lost while the maintainer reviews the ZFS follow-on stack. Edits the tracked `modules/open_zfs/osv/module/os/osv/zfs/arc_os.c` directly (the OSv platform layer is now source files on master, not a numbered patch).

### DIFF
--- a/modules/open_zfs/osv/module/os/osv/zfs/arc_os.c
+++ b/modules/open_zfs/osv/module/os/osv/zfs/arc_os.c
@@ -106,6 +106,18 @@ arc_default_max(uint64_t min, uint64_t allmem)
 	else
 		size = allmem / 8;
 
+	/*
+	 * OSv: the ARC and the application share one physical address space.
+	 * Cap the default at half of RAM for guests > 2 GiB so a large app
+	 * (e.g. PostgreSQL with multi-GiB shared_buffers plus many backends)
+	 * always keeps at least half the guest's memory and a checkpoint burst
+	 * cannot drive the guest OOM.  ZFS_MODULE_PARAM is a no-op on OSv, so
+	 * zfs_arc_max cannot yet override this at boot; a future cmdline wiring
+	 * could.
+	 */
+	if (allmem > (2ULL << 30))
+		size = MIN(size, allmem / 2);
+
 	return (MAX(size, min));
 }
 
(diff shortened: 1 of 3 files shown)


gburd:
Added a second commit: a synchronous hard-mode ARC shrinker, the companion to the half-RAM cap. Even with the ARC capped, at high concurrency (many backends + a large checkpoint + autovacuum worker forks) OSv's reclaimer could abort out-of-memory while the ARC still held evictable pages, because the hard-mode shrinker only lowered the target and woke the async evict thread, returning before any page was freed. The new arc_reduce_target_size_shrink_sync() lowers the target and then waits in arc_wait_for_eviction() until the requested memory is actually evicted, so the reclaimer gets real freed memory before deciding it cannot reclaim; soft mode stays async.

Measured: a virtual-user level that previously aborted with 'could not reclaim any further' now completes, riding through a large mid-run checkpoint, with no reclaimer OOM, no deadlock, and no throughput regression at the level below. Kept as part of this draft since both commits are ARC memory-pressure coexistence fixes for the same app-shares-RAM situation.

gburd:
Marking this ready for review.

It was held as a draft on the reasoning that review attention was the scarce resource and adding to the queue would not help. That reasoning has not aged well: a draft does not concentrate attention, it just hides work, and this is the cheapest thing I have to review - 2 commits, 3 files, merge-clean against master, both commits signed, base current.

One thing I checked rather than assumed, because it could have made this a trap: the second commit edits an inner hunk of `modules/open_zfs/patches/0001-osv-openzfs-upstream-file-edits.patch`, which is the kind of change that silently stops applying when the vendored tree moves. It applies cleanly at the current pin, and the same content builds from scratch in both `conf_zfs=bsd` and `conf_zfs=openzfs` in the integration branch I test from.

No urgency implied. If the queue is the constraint, #1514 and #1516 are both smaller.

PR #1489, opened 2026-09-05
libext: bound ext_readlink and ext_readdir against a crafted ext4 image (rebase of #1449)

gburd:
Rebase of #1449 onto current master. #1449 cannot be updated in place: three commits landed on `modules/libext/ext_vnops.cc` after it was branched (bf16817c9 inode deletion time, af27ba4b7 real fsync + page-cache bridge, c08544641 inode refcounting), and force-push to that head branch is not available to me, so this is a fresh branch. **#1449 can be closed in favour of this one.** Same two bugs, re-verified against the current code, with two corrections to the original patch described at the bottom.

Both overflows are still present verbatim in `master` as of 0e34e4dc2. The upstream refactor moved the functions but did not touch either unbounded length.

## 1. `ext_readlink()` heap overflow (critical)

For a slow symlink (target not inline in the inode):

```c
uint32_t block_size = ext4_sb_get_block_size(&fs->sb);
void *buf = malloc(block_size);                        // e.g. 4096
ext_internal_read(fs, ..., uio->uio_offset, buf, fsize, &read_count);
//                                                ^^^^^ up to 2^64-1
```

`fsize` is `ext4_inode_get_size()`, the raw 64-bit `i_size` off disk, and it is passed as the read *size* into a `block_size` allocation. `ext_internal_read()`'s only clamp is

```c
size = ((uint64_t)size > (fsize - offset)) ? ((size_t)(fsize - offset)) : size;
```

which with `size == fsize` and `offset == 0` is a no-op. So it writes ~`fsize` bytes into a `block_size` heap buffer.

A symlink inode with `i_size = 0x100000` and `i_blocks != 0` (so the slow branch is taken) turns any `readlink()`, or any path traversal through that link, into a ~1 MiB write into a 4 KiB allocation. In a unikernel that is kernel heap corruption, i.e. RCE-class. Reaching it needs only that a crafted ext4 image be mounted: an attached volume, or a multi-tenant host.

Working the arithmetic through for the guarded and unguarded cases:

| `i_size` on disk | buffer | bytes written, unpatched | patched |
|---|---|---|---|
| 64 (normal slow symlink) | 4096 | 64, in bounds | 64, in bounds |
| 1 MiB (the PoC) | 4096 | 1048576 | rejected `EIO` |
| 2^32 | 4096 | 4294967296 | rejected `EIO` |
| 2^64-1 | 4096 | 2^64-1 (~16 EiB) | rejected `EIO` |
| 1023 on a 1 KiB-block fs | 1024 | 1023, in bounds | 1023, in bounds |

**Fix:** reject `fsize > block_size`. That is not a heuristic bound: `ext_fsymlink_set()`, the only creator, already refuses `size > block_size`, and a path is bounded by `PATH_MAX` regardless, so a slow symlink's target always occupies exactly one block. Anything larger is a malformed inode. Also reject an offset at or past EOF, where the `(fsize - offset)` clamp underflows; after the first guard that is no longer an overflow, but it still reads blocks the symlink does not own. Both in-tree callers (`sys_readlink`, `read_link`) pass offset 0, so that guard is defence in depth. And check `malloc`.

## 2. `ext_readdir()` `d_name` overflow (high, old-rev image)

```c
uint16_t name_length = ext4_dir_en_get_name_len(&fs->sb, it.curr);
memcpy(dir->d_name, it.curr->name, name_length);   // d_name is 256 bytes
```

On a rev-0 (`rev_level == 0 && minor_rev_level < 5`) image, `ext4_dir_en_get_name_len()` folds in `name_length_high` and can return up to `entry_len - 8`, roughly a full block. lwext4's `ext4_dir_iterator_set()` validates the name against the *entry* length only, never against the caller's buffer, so a crafted rev-0 entry overflows the fixed 256-byte `d_name` by ~4 KiB. **Fix:** clamp to `sizeof(d_name) - 1`.

## Two corrections versus #1449

Reviewing the fix against the new code turned up two things worth changing, so this is not a plain re-apply:

1. **`EIO` instead of `EINVAL`** for the malformed-inode rejection, to match how `rofs_readlink()` reports corrupt on-disk symlink metadata (07f61154a). `EINVAL` on a path OSv reaches through lookup reads as "bad argument" when the argument is fine and the disk is not.
2. **The regression test in #1449 did not reach the patched code.** It asserted `readlink() == "realfile"`, 9 bytes, which is under `sizeof(inode->blocks)` (60) and therefore served by the *fast* inline path; the guarded `else` branch never executed. It was also added to the generic `tests +=` list, so it entered `usr.manifest` for every `fs_type` and `scripts/test.py` would have run it on zfs/rofs images, where `/data/mylink` does not exist, failing the run.

   The test here covers the legitimate side of both guards, which is what a guest can actually construct: an inline symlink, a slow symlink just over the 60-byte inline limit, a slow symlink at the largest target a 1 KiB block can hold (the boundary the `fsize <= block_size` guard turns on), and a 255-byte directory entry round-tripped through `readdir()`. That pins down that the bounds reject only malformed images and truncate nothing a well-formed one can contain. It creates its own fixtures on a writable ext root and is registered in `ext-only-tests` next to `tst-ext4-rw`, so it is built and run only for `fs_type=ext`.

   Crafting `i_size > block_size` needs debugfs to build the image and cannot be done from inside the guest, so the overflow case itself is argued from the arithmetic above rather than asserted in the test.

## Verification

- `modules/libext/ext_vnops.cc` compiles clean under the module's own flags (`-std=gnu++11 -Wall -fno-exceptions -fno-rtti -O2`) against current OSv and lwext4 headers; no new warnings, including under `-Wextra -Wsign-compare` for the two patched functions (the `uint16_t` vs `sizeof` comparison in the clamp is warning-free).
- `tst-ext-readlink` compiles `-Wall -Wextra` clean and all four cases pass when run against a real ext4 filesystem, including the 1023-byte slow symlink and the 255-byte directory entry.
- The length arithmetic in the table above was reproduced from `ext_internal_read()`'s clamp rather than estimated.

Both bugs are in shipped `master`, not in new code.

### DIFF
--- a/modules/libext/ext_vnops.cc
+++ b/modules/libext/ext_vnops.cc
@@ -677,6 +677,15 @@ ext_readdir(struct vnode *dvp, struct file *fp, struct dirent *dir)
         if (ext4_dir_en_get_inode(it.curr) != 0) {
             memset(dir->d_name, 0, sizeof(dir->d_name));
             uint16_t name_length = ext4_dir_en_get_name_len(&fs->sb, it.curr);
+            // On an old-rev (rev0, minor<5) image ext4_dir_en_get_name_len()
+            // folds in name_length_high, so it can return up to entry_len-8
+            // (~block_size, ~4 KiB) while d_name is only 256 bytes.  lwext4's
+            // ext4_dir_iterator_set() only bounds the name against the entry
+            // length, not against d_name, so a crafted rev0 directory entry
+            // memcpy'd here would overflow the fixed-size d_name.  Clamp it.
+            if (name_length >= sizeof(dir->d_name)) {
+                name_length = sizeof(dir->d_name) - 1;
+            }
             memcpy(dir->d_name, it.curr->name, name_length);
             ext_debug("readdir directory with i-node=%ld at offset:%ld => entry name:%s\n", dvp->v_ino, file_offset(fp), dir->d_name);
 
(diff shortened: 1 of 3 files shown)


gburd:
Closing: this was opened on a mistaken premise. I believed #1449 conflicted with master and needed a rebase. It does not. `git merge-tree --write-tree` against current master exits 0 for #1449, with `modules/libext/ext_vnops.cc` auto-merging against bf16817c9/af27ba4b7/c08544641. My earlier check was run from a worktree predating d7c3cb81a, so it lacked the `.gitattributes` that activates the `modules/tests/Makefile merge=union` driver, and produced a false conflict. Full retraction on #1449.

#1449 merges clean and remains the canonical PR for this fix. The two test improvements found while porting (the regression test not reaching the guarded slow-symlink branch, and its placement in the generic test list instead of `ext-only-tests`) have been applied directly to #1449 as 9340ab1f9 and ba63c5725, along with an AddressSanitizer demonstration that the guard is what prevents the overflow.

Nothing here needs review. Apologies for the noise.

PR #1490, opened 2026-09-05
aarch64: cover all of physical memory in the temporary phys map

gburd:
## Summary

On aarch64, an OSv guest with more than about 3 GB of RAM hangs during early
boot with no output past the banner. The temporary physical-memory map only
covers the first 4 GB of physical addresses, so the first access above that
faults before any exception handler is installed.

## The bug

`setup_temporary_phys_map()` aliased each `phys_mem` window onto the boot
identity L1 table:

```c
pt_ttbr0[mmu::pt_index(base, 3)] = pt_ttbr0[0];
```

That boot table (built in `boot.S`) has only four valid entries,
`ident_pt_l2_0..3_ttbr0`, one 1 GB block each, so it describes physical
addresses 0-4 GB and leaves the remaining 508 entries zero. Every `phys_mem`
window therefore resolved only within the first 4 GB.

`free_initial_memory_range()` -> `page_range_allocator::insert()` writes a
back-pointer into the last 8 bytes of each free range. For memory above 4 GB
that store takes a translation fault while the exception vectors are not yet
usable, so the guest wedges with no diagnostic.

The `virt` machine bases RAM at 1 GB, which is why the boundary looks like
"3 GB works, 4 GB hangs": 3 GB of RAM ends exactly at the 4 GB limit, and 4 GB
of RAM ends at 5 GB, past it.

Backtrace at `-m 8G` (before the fix):

```
#0  exception_vectors () at arch/aarch64/entry.S:37
#1  arch_setup_free_memory () at arch/aarch64/arch-setup.cc:96
    x1 = 0x400240000000   (phys_mem + 0x240000000, the end of 8 GB at a 1 GB base)
```

## The fix

Give the `phys_mem` windows their own L1 table populated with 1 GB block
descriptors covering all of physical memory, instead of aliasing the
four-entry boot identity table. 512 entries describe up to 512 GB; beyond that
the code now aborts with a clear message rather than hanging.

`pt_ttbr0[0]` is left untouched, so the low identity mapping and the existing
kernel mapping are unaffected.

## Validation

Booted under KVM on aarch64 hardware at 2, 4, 8, 16 and 32 GB. All five now
reach the scheduler. Before the fix, 4 GB and above printed nothing after the
banner; 2 GB was unaffected.

One file, 39 insertions, 2 deletions. No dependency on any other pending change.

### DIFF
--- a/arch/aarch64/arch-setup.cc
+++ b/arch/aarch64/arch-setup.cc
@@ -39,13 +39,50 @@
 
 #include <osv/kernel_config_networking_stack.h>
 
+// Temporary level-1 table for the phys_mem windows, filled with 1 GB block
+// descriptors so the windows cover all of physical memory during early boot.
+// One table = 512 entries x 1 GB = 512 GB of coverage.
+// ponytail: 512 GB ceiling; chain a second L1 table if a guest ever exceeds it
+// (setup_temporary_phys_map aborts with a clear message rather than hanging).
+static constexpr unsigned temp_phys_map_gbs = 512;
+static u64 temp_phys_map_l1[temp_phys_map_gbs] __attribute__((aligned(4096)));
+
 void setup_temporary_phys_map()
 {
-    // duplicate 1:1 mapping into the lower part of phys_mem
+    // Map the phys_mem windows over all of physical memory.
+    //
+    // This used to simply alias each window's top-level entry onto the boot
+    // identity entry (pt_ttbr0[pt_index(base,3)] = pt_ttbr0[0]).  That shares
+    // the boot identity level-1 table from boot.S, which only has valid
+    // entries for the first 4 GB of physical addresses (ident_pt_l2_0..3).
+    // Any RAM above 4 GB was therefore absent from the window, and
+    // free_initial_memory_range() -> page_range_allocator::insert() - which
+    // writes a back-pointer into the last 8 bytes of the free range - faulted
+    // at phys_mem + <end of RAM>.  With no exception handler installed this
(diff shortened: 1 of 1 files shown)


gburd:
## Additional evidence: the >3 GB hang reproduces at smp=1, isolating it from SMP

Preparing an aarch64 benchmark base (Graviton bare-metal host, qemu 8.1.3 KVM,
guest = OSv + PostgreSQL 18) on a branch carrying neither this fix nor #1491
produced these three boot outcomes:

    smp=1  mem=4G  ->  silent hang immediately after the OSv banner.  No assert,
                       no panic, no further console output whatsoever.
    smp=32 mem=2G  ->  Assertion failed: smp_idx < max_sgi_cpus
                       (arch/aarch64/gic-v3.cc: send_sgi: 671)   <- that one is #1491
    smp=2  mem=2G  ->  boots normally, "Booted up in 124.41 ms"

The `smp=1 mem=4G` case is the useful datum for this PR: with a single vCPU there
is no IPI, no load balancer and no SMP interaction of any kind, so the hang is
purely a function of RAM size crossing the 4 GB temporary-phys-map limit. It also
confirms the diagnostic character described above, that the failure happens before
the exception vectors are usable, so there is no output to go on: the console shows
the banner and then nothing.

Practical consequence: #1490 and #1491 together are prerequisites for booting an
aarch64 OSv guest at any modern instance geometry (>~3 GB RAM or >16 vCPU). Without
them there is no bootable large aarch64 guest at all, so neither is incremental
hardening.

PR #1502, opened 2026-09-09
vfs: fix AB-BA deadlock between dentry_hash_lock and the vnode lock

gburd:
The VFS has an implicit lock order that `namei()` establishes and `drele()`
violated. This series states the rule, adds an assertion that enforces it, and
makes `drele()` conform.

## The rule

`fs/vfs/vfs_lookup.cc` holds the directory vnode lock across the dentry cache
lookup, at two sites (`:178` and `:284`):

    vn_lock(dvp);
    dp = dentry_lookup(mp, node);          // takes dentry_hash_lock
    if (dp == nullptr) {
        error = VOP_LOOKUP(dvp, name, &vp);
        dp = dentry_alloc(ddp, vp, node);  // takes dentry_hash_lock
    }
    vn_unlock(dvp);

So the order is **vnode lock, then `dentry_hash_lock`**. Nothing may take a
vnode lock while holding `dentry_hash_lock`.

`drele()` did exactly that. It took `dentry_hash_lock` and then called
`vn_del_name()`, which takes the vnode lock internally
(`fs/vfs/vfs_vnode.cc:531`):

    mutex_lock(&dentry_hash_lock);
    if (--dp->d_refcnt) { ... }
    LIST_REMOVE(dp, d_link);
    vn_del_name(dp->d_vnode, dp);          // vn_lock(vp) in here
    mutex_unlock(&dentry_hash_lock);

A `namei()` holding `vn_lock` and waiting for `dentry_hash_lock`, against a
`drele()` holding `dentry_hash_lock` and waiting for `vn_lock`, is an AB-BA
deadlock in which neither side can proceed.

## Evidence

The first commit adds `ASSERT(!vfs_dentry_hash_lock_held())` to `vn_lock()`
under `DEBUG_VFS`, so the rule is checked rather than asserted in prose. On
unfixed code it fires immediately, during `vfs_init`, with this stack:

    0x00000000403bc1b8 <vn_lock+168>
    0x00000000403bcc71 <vn_del_name+17>
    0x00000000403c1167 <drele+103>
    0x00000000403c0be7 <vfs_file::close()+87>
    0x00000000403b8090 <fdrop+80>
    0x00000000403b8216 <fdclose(int)+118>
    0x00000000403b1267 <close+23>
    0x00000000403b6832 <unpack_bootfs()+258>
    0x00000000403b69b6 <vfs_init+182>

On fixed code, same test and same configuration, the assertion is silent and
the test passes:

    VFSFLAG ... unset, default 1
    Cmdline: /tests/tst-dentry-lock.so 8 40
    tst-dentry-lock: 8 threads x 40 iters
    tst-dentry-lock: 2880 ops, 0 failures
    OK

Note what the stack shows: `fdrop()` is the ordinary `close()` path
(`fs/vfs/main.cc:154`, `:159`, `:194`). The illegal order is therefore taken by
every `close()` that drops a last dentry reference. It is not a rare
interleaving; the lock cycle is constructed constantly, and the only missing
ingredient for the deadlock is a concurrent `namei()` on the same vnode.

That ingredient was supplied in production. On a 32-vCPU guest running a
forking server with 184 concurrent backends, the system wedged with
`dentry_hash_lock` held, every vCPU halted in `do_idle` with an empty wakeup
mask, and over a thousand threads parked, while ~27 GiB of memory was free.

**The included stress test passes on unfixed code.** 48 threads x 400
iterations, 172,800 operations, 16 vCPUs under TCG, and it does not wedge:
OSv's mutex hold times here are short, and closing the cycle needs genuine
parallelism plus a same-vnode collision. A reviewer who runs it on a laptop and
sees it pass should not conclude the defect is fictional; the assertion above
is the reliable signal, and the hang itself was observed on 32 real vCPUs. The
test is included because it exercises the paths (concurrent open/close in one
directory, plus rename and unlink churn) and would catch a regression that
reintroduces a blocking call under the lock.

## The fix

Release `dentry_hash_lock` before calling `vn_del_name()`. Unlinking from the
hash chain stays inside the critical section, and that is what makes the early
release safe: once the dentry is off the chain, `dentry_lookup()` cannot find
it, so no thread can take a new reference. The releasing thread is its sole
owner and the refcount is 0 and stays 0.

`dentry_hash_lock` becomes a leaf lock: it calls out to no other subsystem.
Every acquisition is routed through accessors so a new call site cannot quietly
bypass the rule, and the per-thread ownership tracking that the assertion needs
costs nothing when `DEBUG_VFS` is off.

## Commits

1. `vfs: fix AB-BA deadlock between dentry_hash_lock and the vnode lock` - the
   ordering fix, the documented rule, and the assertion.
2. `vfs: do not allocate while holding dentry_hash_lock in dentry_move()` - an
   independent latent defect found while auditing the other call sites.
   `dentry_move()` called `strdup()` inside the critical section; that
   allocation can block in the page allocator while every lookup in the system
   waits. Hoisted above the lock. Not the deadlock, and on the `rename()` path
   rather than the `open()` path, but the same class of mistake.
3. `tests: concurrency stress for the dentry cache lock order` - the test,
   with the caveat above recorded in its header comment.

## Notes

Not fork-specific. `fork` supplies the concurrency that closes the cycle, but
the inversion is present regardless: there are no `CONF_fork` references in
this series, and `vfs_dentry.cc` and `vfs_vnode.cc` compile clean with
`conf_fork=0`.

A separate change splitting the global `dentry_hash_lock` into per-bucket locks
is deliberately **not** in this PR. The 32 hash buckets have always shared one
mutex taken at 6 sites, which is a scalability defect, but it is independent of
this correctness fix and a reviewer may reasonably want one without the other.
It also has to come after this fix, because a per-bucket lock inverts against
the vnode lock exactly as the global one does. Its throughput effect is
currently unmeasured and no claim is made about it.

Tested on x86_64: 12 of 12 relevant tests pass (`tst-vfs`, `tst-symlink` 141
assertions, `tst-remove` 38, `tst-pipe` 87, `tst-mmap`, `tst-condvar`,
`tst-yield`, `tst-dentry-lock`, and three fork tests). `tst-openat` reports one
pre-existing failure that reproduces identically on unpatched master.

---

**Update after review (c7ab5929e).** The commit messages made three claims they had not earned, all corrected in a rewritten series:

- The reachability argument in the first commit was wrong (a thread holding `vn_lock` also holds a reference on the dentry it found, so a concurrent `drele()` cannot reach zero). Replaced with the `sys_rename()` path: it holds `vn_lock(vp1)` across `lookup()` and `dentry_move()`/`dentry_remove()`, `vp1` is any vnode type, and two hardlinks to one file are enough.
- The production hang we observed was **not** this inversion. Its captured state shows the lock's owner blocked in `memory::page_pool::l1::alloc_page` while holding the lock, which is the `dentry_move()` allocation fixed by the second commit. That evidence now sits with that commit, and the first commit says the inversion was found by inspection.
- "Leaf lock" was overstated: `dentry_hash_lock` no longer calls into the vnode layer, but it still nests `dentry->d_lock` via `dentry_children_remove()`, consistently and without inversion.

The test header claimed it hangs on unfixed code, contradicting the description below. The description was right. The test runs on ramfs, where `ramfs_link()` is `vop_eperm`, so it cannot build the two-hardlinks-to-one-file aliasing the race requires and was never going to reproduce it. The header now states that, and what the test does cover.

### DIFF
--- a/fs/vfs/vfs_dentry.cc
+++ b/fs/vfs/vfs_dentry.cc
@@ -46,7 +46,64 @@
 
 static LIST_HEAD(dentry_hash_head, dentry) dentry_hash_table[DENTRY_BUCKETS];
 static LIST_HEAD(fake, dentry) fake;
+
+/*
+ * LOCK ORDERING RULE (whole VFS):
+ *
+ *     vnode lock (vn_lock)  ->  dentry_hash_lock
+ *
+ * and NEVER the reverse.  namei() establishes this direction: it holds
+ * vn_lock(dvp) across dentry_lookup()/dentry_alloc() (vfs_lookup.cc), both of
+ * which take dentry_hash_lock inside.  Therefore no code may call anything
+ * that takes a vnode lock while holding dentry_hash_lock.
+ *
+ * drele() used to violate exactly that: it took dentry_hash_lock and then
+ * called vn_del_name(), which does vn_lock() internally.  A namei() on one
+ * thread (vn_lock held, waiting for dentry_hash_lock) against a drele() on
+ * another (dentry_hash_lock held, waiting for vn_lock) is an AB-BA deadlock
+ * with nothing left runnable -- observed as all vCPUs halted in do_idle with
+ * an empty wakeup mask and >1000 threads parked, while memory was plentiful.
+ *
+ * dentry_hash_lock is a leaf lock now.  Keep it that way: do not call out to
+ * vnode code, and do not allocate, while holding it.
+ */
 static mutex dentry_hash_lock;
+
+#ifdef DEBUG_VFS
+/*
+ * Teeth for the ordering rule above.  When DEBUG_VFS is on, every thread
+ * tracks whether it currently holds dentry_hash_lock; vn_lock() asserts that
+ * it does not.  This fires on the unfixed drele() and is silent once
+ * vn_del_name() is moved out of the critical section.
+ */
+extern "C" bool vfs_dentry_hash_lock_held(void);   /* declared in vfs.h */
+static __thread int dentry_hash_lock_depth;
+bool vfs_dentry_hash_lock_held(void)
+{
+    return dentry_hash_lock_depth != 0;
+}
+static void dentry_hash_lock_enter(void) { dentry_hash_lock_depth++; }
+static void dentry_hash_lock_exit(void)  { dentry_hash_lock_depth--; }
+#else
+static inline void dentry_hash_lock_enter(void) {}
+static inline void dentry_hash_lock_exit(void)  {}
+#endif
+
+/*
+ * dentry_hash_lock accessors.  Use these, not mutex_lock/unlock directly, so
+ * the ordering instrumentation cannot be bypassed by a new call site.
+ */
+static void dentry_hash_lock_acquire(void)
+{
+    mutex_lock(&dentry_hash_lock);
+    dentry_hash_lock_enter();
+}
+
+static void dentry_hash_lock_release(void)
+{
+    dentry_hash_lock_exit();
+    mutex_unlock(&dentry_hash_lock);
+}
 
 /*
  * Get the hash value from the mount point and path name.
@@ -95,9 +152,9 @@ dentry_alloc(struct dentry *parent_dp, struct vnode *vp, const char *path)
 
     vn_add_name(vp, dp);
 
-    mutex_lock(&dentry_hash_lock);
+    dentry_hash_lock_acquire();
     LIST_INSERT_HEAD(&dentry_hash_table[dentry_hash(mp, path)], dp, d_link);
-    mutex_unlock(&dentry_hash_lock);
+    dentry_hash_lock_release();
     return dp;
 };
 
@@ -106,15 +163,15 @@ dentry_lookup(struct mount *mp, char *path)
 {
     struct dentry *dp;
 
-    mutex_lock(&dentry_hash_lock);
+    dentry_hash_lock_acquire();
     LIST_FOREACH(dp, &dentry_hash_table[dentry_hash(mp, path)], d_link) {
         if (dp->d_mount == mp && !strncmp(dp->d_path, path, PATH_MAX)) {
             dp->d_refcnt++;
-            mutex_unlock(&dentry_hash_lock);
+            dentry_hash_lock_release();
             return dp;
         }
     }
-    mutex_unlock(&dentry_hash_lock);
+    dentry_hash_lock_release();
     return nullptr;                /* not found */
 }
 
@@ -153,6 +210,7 @@ dentry_move(struct dentry *dp, struct dentry *parent_dp, char *path)
     }
 
     WITH_LOCK(dentry_hash_lock) {
+        dentry_hash_lock_enter();
         // Remove all dp's child dentries from the hashtable.
         dentry_children_remove(dp);
         // Remove dp with outdated hash info from the hashtable.
@@ -163,6 +221,7 @@ dentry_move(struct dentry *dp, struct dentry *parent_dp, char *path)
         // Insert dp updated hash info into the hashtable.
         LIST_INSERT_HEAD(&dentry_hash_table[dentry_hash(dp->d_mount, path)],
             dp, d_link);
+        dentry_hash_lock_exit();
     }
 
     if (old_pdp) {
@@ -175,11 +234,11 @@ dentry_move(struct dentry *dp, struct dentry *parent_dp, char *path)
 void
 dentry_remove(struct dentry *dp)
 {
-    mutex_lock(&dentry_hash_lock);
+    dentry_hash_lock_acquire();
     LIST_REMOVE(dp, d_link);
     /* put it on a fake list for drele() to work*/
     LIST_INSERT_HEAD(&fake, dp, d_link);
-    mutex_unlock(&dentry_hash_lock);
+    dentry_hash_lock_release();
 }
 
 void
@@ -188,9 +247,9 @@ dref(struct dentry *dp)
     ASSERT(dp);
     ASSERT(dp->d_refcnt > 0);
 
-    mutex_lock(&dentry_hash_lock);
+    dentry_hash_lock_acquire();
     dp->d_refcnt++;
-    mutex_unlock(&dentry_hash_lock);
+    dentry_hash_lock_release();
 }
 
 void
@@ -199,15 +258,27 @@ drele(struct dentry *dp)
     ASSERT(dp);
     ASSERT(dp->d_refcnt > 0);
 
-    mutex_lock(&dentry_hash_lock);
+    dentry_hash_lock_acquire();
     if (--dp->d_refcnt) {
-        mutex_unlock(&dentry_hash_lock);
+        dentry_hash_lock_release();
         return;
     }
+    /*
+     * Last reference.  Unlink from the hash chain while still holding the
+     * lock -- that is what makes the drop below safe: once dp is off the
+     * chain, dentry_lookup() can no longer find it, so no other thread can
+     * resurrect it by taking a new reference.  This thread is the sole owner
+     * of dp from here on, and d_refcnt is 0 and stays 0.
+     *
+     * vn_del_name() must NOT be called under dentry_hash_lock: it takes the
+     * vnode lock, which inverts the vn_lock -> dentry_hash_lock order that
+     * namei() establishes (see the ordering note at the top of this file).
+     * Release first, then touch the vnode.
+     */
     LIST_REMOVE(dp, d_link);
-    vn_del_name(dp->d_vnode, dp);
+    dentry_hash_lock_release();
 
-    mutex_unlock(&dentry_hash_lock);
+    vn_del_name(dp->d_vnode, dp);
 
     if (dp->d_parent) {
         WITH_LOCK(dp->d_parent->d_lock) {
--- a/fs/vfs/vfs_dentry.cc
+++ b/fs/vfs/vfs_dentry.cc
@@ -193,6 +193,11 @@ dentry_move(struct dentry *dp, struct dentry *parent_dp, char *path)
 {
     struct dentry *old_pdp = dp->d_parent;
     char *old_path = dp->d_path;
+    // Duplicate the new path BEFORE taking dentry_hash_lock.  strdup() can
+    // block in the page allocator, and dentry_hash_lock is a leaf lock held
+    // by every lookup in the system; sleeping under it stalls all VFS name
+    // resolution for the duration.  Nothing here needs the lock.
+    char *new_path = strdup(path);
 
     if (old_pdp) {
         WITH_LOCK(old_pdp->d_lock) {
@@ -215,8 +220,8 @@ dentry_move(struct dentry *dp, struct dentry *parent_dp, char *path)
         dentry_children_remove(dp);
         // Remove dp with outdated hash info from the hashtable.
         LIST_REMOVE(dp, d_link);
-        // Update dp.
-        dp->d_path = strdup(path);
+        // Update dp with the path duplicated above, outside the lock.
+        dp->d_path = new_path;
         dp->d_parent = parent_dp;
         // Insert dp updated hash info into the hashtable.
         LIST_INSERT_HEAD(&dentry_hash_table[dentry_hash(dp->d_mount, path)],
--- a/fs/vfs/vfs.h
+++ b/fs/vfs/vfs.h
@@ -160,6 +160,14 @@ void drele(struct dentry *dp);
 void dentry_init(void);
 
 #ifdef DEBUG_VFS
+/*
+ * True iff the calling thread holds dentry_hash_lock.  Used by vn_lock() to
+ * assert the VFS lock order (vnode lock -> dentry_hash_lock, never reverse).
+ */
+bool vfs_dentry_hash_lock_held(void);
+#endif
+
+#ifdef DEBUG_VFS
 void	 vnode_dump(void);
 void	 mount_dump(void);
 #endif
--- a/fs/vfs/vfs_vnode.cc
+++ b/fs/vfs/vfs_vnode.cc
@@ -143,6 +143,16 @@ vn_lock(struct vnode *vp)
 	ASSERT(vp);
 	ASSERT(vp->v_refcnt > 0);
 
+#ifdef DEBUG_VFS
+	/*
+	 * Lock-order check with teeth (see the ordering note in vfs_dentry.cc):
+	 * the VFS order is vnode lock -> dentry_hash_lock, never the reverse.
+	 * Taking a vnode lock while already holding dentry_hash_lock is the
+	 * AB-BA inversion that deadlocked drele() against namei().
+	 */
+	ASSERT(!vfs_dentry_hash_lock_held());
+#endif
+
 	mutex_lock(&vp->v_lock);
 	vp->v_nrlocks++;
 	DPRINTF(VFSDB_VNODE, ("vn_lock:   %s\n", vn_path(vp)));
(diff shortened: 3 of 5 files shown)


gburd:
End-to-end confirmation on a real workload, in addition to the lock-order assertion already in the PR.

A PostgreSQL-on-ZFS HammerDB run on a base WITHOUT this fix wedged the guest hard at 32 virtual users: `top -H` on the host showed all 35 qemu vCPU threads asleep, zero runnable, with ~27 GiB free. That is the AB-BA deadlock this PR fixes-a true all-asleep hang, not a slow path (a lost wake or memory stall would make progress with that much memory free).

Rebuilding the identical image with this fix applied and re-running the same 32-VU cells: no wedge, three cells completed at ~62,400 NOPM, with ~18 of 32 vCPUs busy (vs 0 in the wedge). So the fix removes the deadlock end-to-end under a real database workload, not just under the lock-order assertion.

To be clear about scope: clearing the deadlock lets the workload run at 32 VU, but throughput there is still well below Linux on the same hardware-that ceiling is a separate WAL-commit-completion serialization issue, not this deadlock, and is not something this PR claims to address. This PR's claim is narrower and now has both forms of proof: the lock-order assertion fires on the unfixed path, and the workload that deadlocked without it completes with it.

fs/vfs/vfs_dentry.cc
```diff
@@ -46,8 +46,65 @@
...
 static mutex dentry_hash_lock;
 
+#ifdef DEBUG_VFS
+/*
+ * Teeth for the ordering rule above.  When DEBUG_VFS is on, every thread
+ * tracks whether it currently holds dentry_hash_lock; vn_lock() asserts that
+ * it does not.  This fires on the unfixed drele() and is silent once
+ * vn_del_name() is moved out of the critical section.
+ */
+extern "C" bool vfs_dentry_hash_lock_held(void);   /* declared in vfs.h */
```
nyh: Nitpick: if it's already declared in vfs.h, and this file includes that header file, why do we need to declare it again here? I think it's harmless, but also useless.

fs/vfs/vfs_dentry.cc
```diff
@@ -46,8 +46,65 @@
...
+static __thread int dentry_hash_lock_depth;
+bool vfs_dentry_hash_lock_held(void)
+{
+    return dentry_hash_lock_depth != 0;
+}
+static void dentry_hash_lock_enter(void) { dentry_hash_lock_depth++; }
```
nyh: nitpick: why is these extra depth variables needed? Don't our locks already have a depth that we can use?

fs/vfs/vfs_dentry.cc
```diff
@@ -153,6 +210,7 @@ dentry_move(struct dentry *dp, struct dentry *parent_dp, char *path)
     }
 
     WITH_LOCK(dentry_hash_lock) {
+        dentry_hash_lock_enter();
```
nyh: Nitpick: can't we fix WITH_LOCK(dentry_hash_lock) to do those dentry_hash_lock_enter()/exit() if they are really needed?

nyh (changes requested):
Thanks. In general, the fix looked good to me, but I feel really "rusty" in this code, so I asked my AI to critically review it. It agreed that the deadlock you found is real, but had a bunch of complaints about the details - please look into them. The following are edited quotes from my AI :

**The bug is real and the fix is correct**. But three of the claims the PR uses to justify it are wrong or overstated, including one in a commit message and one in the test's header comment, and those should be corrected before this goes into history. There's also a worse latent bug sitting right next to the one commit 2 fixes.

**Claim 1 that's wrong** — the reachability argument (commit 1 message):
"It is reachable whenever the dentry being released is a directory, which namei() itself does on every path component, so it is not theoretical."

That doesn't follow, and a skeptical reviewer can shoot it down. namei() holds a dref on ddp while it holds vn_lock(ddp->d_vnode), so a concurrent drele() on that dentry cannot reach 0. And two threads walking the same path get the same dentry from dentry_lookup(). **You need two distinct dentries aliasing one vnode**, which the commit message never says.

There is a much stronger argument available, and I'd ask for it to replace the current one: [Nadav: however, you said you saw this bug in practice, what use case did it use?]

sys_rename() holds vn_lock(vp1) — the renamed object's own vnode, any type — from vfs_syscalls.cc:730 all the way across lookup(dest,...) at :746 (which is namei → dentry_lookup → dentry_hash_lock) and dentry_move()/dentry_remove() at :827-829, with dvp1/dvp2/vp2 locked too. So the A-side of the cycle is not restricted to directory vnodes at all.

Combine that with hard links and you get a two-line recipe: thread A does rename("/a", "/c"), thread B does the last close() of /b, where /a and /b are links to the same file. A holds vn_lock(F) waiting for dentry_hash_lock; B holds dentry_hash_lock waiting for vn_lock(F). On ramfs this particular recipe is unavailable (ramfs_link is vop_eperm, ramfs_vnops.cc:746) — but ZFS supports it, and the forking-server report was presumably not on ramfs. On ramfs the aliasing instead comes from dentry_children_remove() dropping still-referenced children off the hash so a later lookup allocates a second dentry for the same vnode, and from dentry_remove()'s fake list.

Worth also noting in passing: the obvious /tmp/x/. aliasing route does not exist, because path_conv() strips . and .. before namei() ever runs (vfs_task.cc:68).

**Claim 2 that's wrong — the test (commit 3)**
The commit message and the test's header comment both say:

"On an unfixed kernel this wedges with nothing runnable and the harness kills it by timeout."

The PR description says the opposite, in bold: "The included stress test passes on unfixed code.", and explains why.

Both of these comments can't be true at the same time. [Nadav: Moreover, if this test really does **not** reproduce this bug, maybe it demonstrates my AI's claims that the scenario where you thought the hang could occur is incorrect - a hang can't occur then]. Also, how long does this new test run?

**Claim 3 that's overstated — "leaf lock" / "every acquisition routed through accessors"**
Both sentences appear in the new file comment and the commit message, and neither is quite true of the code as submitted:

dentry_hash_lock is not a leaf lock: vfs_dentry.cc:155-157 still takes dp->d_lock inside it via dentry_children_remove(). There's no inversion (d_lock is never held while acquiring the hash lock — I checked all four sites), so it's safe, but the comment claims more than it delivers.
dentry_move() still uses WITH_LOCK(dentry_hash_lock) with hand-written dentry_hash_lock_enter()/_exit() inside the body, bypassing the very accessors the commit says cannot be bypassed. A small RAII/lockable wrapper so WITH_LOCK goes through the same path would make the claim true.

Also: the instrumentation and the assertion are #ifdef DEBUG_VFS only, and DEBUG_VFS has been commented out at vfs.h:50 essentially forever and isn't built by CI — so the "teeth" never bite in any build anyone runs. At minimum, confirm DEBUG_VFS still compiles at all. Minor: the new decl adds a second adjacent #ifdef DEBUG_VFS block in vfs.h that should just merge with the existing one, and it puts a bool inside __BEGIN_DECLS.

(The PR text's "taken at 14 sites" is also off — there are 6 acquisition sites. Cosmetic, description only.)

**Commit 2 (strdup hoist)**
**Correct and safe** — nothing in the copy needs the lock, and old_path is still freed at the end. One nit, pre-existing but now trivial to handle since the allocation moved up: strdup() is unchecked, so on OOM dp->d_path becomes NULL and dentry_lookup()'s strncmp(dp->d_path, ...) will fault. Bailing before the list surgery is now a two-line change.

But since commit 2 claims to have audited dentry_move()'s call sites, this belongs in the same review:

dentry_children_remove() does LIST_REMOVE(entry, d_link) on each child but does not put them on the fake list — unlike dentry_remove(), which does exactly that, with the comment "put it on a fake list for drele() to work". So after renaming a directory, any child dentry that is still referenced will, on its final drele(), run LIST_REMOVE a second time on its hash link. LIST_REMOVE (bsd/sys/sys/queue.h:430) does `*elm->le_prev = le_next` unconditionally, so that's a write through a stale (possibly freed) pointer. That's a worse latent defect than the strdup() one, in the same function. Not a blocker for this PR, but I'd ask for it to be at least acknowledged as a follow-up.

My AI also suggests there's a second similar deadlock which may have been the one you saw in the field, not this one:

there is at least one other AB-BA of the same class in this code (vnode_lock vs v_lock: vfs_vnode.cc:111-118) takes v_lock under VNODE_LOCK, while vfs_vnode.cc:234 vput() takes VNODE_LOCK under v_lock; mostly masked by the parent-directory lock serializing lookups). So the fix is right, but I wouldn't assume it's necessarily the wedge they saw.

gburd:
All three were right, and the result is smaller than what you reviewed. Pushed as `d6166ed3a`.

**The duplicate declaration** - deleted. `vfs.h:167` already declares it and this file includes `vfs.h`, so it was pure noise.

**The depth variables** - you are right that the lock already has what I needed, and it is better than what I wrote. `dentry_hash_lock` is a `mutex`, and `mutex::owned()` is exactly the predicate:

```c
bool mutex::owned() const {
    return owner.load(std::memory_order_relaxed) == sched::thread::current();
}
```

So `vfs_dentry_hash_lock_held()` is now just `return dentry_hash_lock.owned();`. That deleted the thread-local counter, both `enter`/`exit` helpers, the non-debug no-op stubs, and every paired call site: 18 lines out, 6 in. It is also more robust than the counter was, since a future call site that forgot the pairing would have silently desynchronized my counter from the real lock state, and `owned()` cannot drift that way.

**`WITH_LOCK` doing the enter/exit** - agreed that the hook belonged in one place rather than at every site, and with `owned()` there is no hook to place at all, so `WITH_LOCK` is untouched.

The assert semantics are unchanged: both the definition and the `vn_lock()` call site remain under `DEBUG_VFS`, and `owned()` is per-thread just as the counter was, so the AB-BA check still fires on the unfixed `drele()` path and is silent after the fix.

gburd:
You are right on all three, and the answer to your question is worse than you guessed. Series rewritten and force-pushed as `c7ab5929e`; the only code change from what you reviewed is the test's header comment, the rest is commit messages that were making claims they had not earned.

**"What use case did it use?"** - the honest answer is that the AB-BA inversion was **found by reading the code, not by hitting it**, and the commit message had no business attaching the production observation to it. What we actually observed on a 32-vCPU guest (stock PostgreSQL, 184 concurrent backends under HammerDB, every vCPU halted in `do_idle` with an empty wakeup mask and memory still plentiful) was a *different* defect on the same lock. From the captured state:

```
mutex: {count = 8, depth = 1, owner = 0x40032f208040, sequence = 92}
owner = live thread 220584 (postgres), stack:
  #0 sched::thread::switch_to  ...  #3 memory::page_pool::l1::alloc_page  core/mempool.cc:1464
```

The lock's owner was blocked **inside the page allocator while holding the global VFS lock** - which is the `dentry_move()`/`strdup()` defect fixed by the second commit in this series, not the inversion fixed by the first. I have moved that evidence to where it belongs and said explicitly in commit 1 that the inversion was found by inspection, so neither commit claims the other's evidence.

**The reachability argument was wrong.** You are right that `namei()` holds a reference on the dentry it found while holding `vn_lock`, so a concurrent `drele()` of that dentry cannot reach zero, and two threads walking one path get the same dentry. The race needs two *distinct* dentries aliasing one vnode, which my message never said. I have taken your `sys_rename()` argument instead: it holds `vn_lock(vp1)` across `lookup()` of the destination and the following `dentry_move()`/`dentry_remove()` (`vfs_syscalls.cc:730`, `:746`, `:827-829`), `vp1` is any vnode type so the A side is not directory-restricted, and two hardlinks to one file are enough - rename one while the other's last `close()` runs.

**The test claim contradicted my own PR description, and your inference was correct.** The header said it hangs on unfixed code; the description said it passes. The description was the true one and I never went back and fixed the header. Your suspicion that the scenario might be wrong turns out to be half right, and your hardlink observation explains it: the test runs on ramfs, and `ramfs_link()` is `vop_eperm`, so it **cannot construct the two-hardlinks-to-one-file aliasing the race requires**. It was never going to reproduce, on any timeout. The header now says that plainly, says what the test does cover (dentry-cache paths under contention, plus the `DEBUG_VFS` order assertion, which catches a reintroduced inversion without needing to win a race), and states that it passes before and after the fix. Runtime: 48 threads x 400 iterations, about 172,800 operations, well under a minute on 16 vCPUs under TCG.

**"Leaf lock" was overstated.** `dentry_children_remove()` takes `dentry->d_lock` inside the `dentry_hash_lock` critical section, so the lock does nest another. The accurate claim, which is what the fix actually achieved, is that it no longer calls into the *vnode* layer; it still nests `d_lock`, consistently, and as you found there is no inversion at any of the sites. Corrected to say that.

The three inline nitpicks are also addressed in the same series (`owned()` replacing the depth counter, the duplicate declaration gone, `WITH_LOCK` untouched).

gburd:
Three more from your review that I had not answered yet. All three check out.

**`DEBUG_VFS` never builds.** `fs/vfs/vfs.h:50` is `/* #define DEBUG_VFS 1 */` and nothing in `conf/` or the Makefile defines it, so the `vn_lock()` assertion I described as the teeth of the ordering rule does not compile in any normal build, and my claim that it "fires on the unfixed drele()" is not something a reader can check. Now that the predicate is just `dentry_hash_lock.owned()` (a relaxed load and a compare) the check is cheap enough to be unconditional, which makes the rule actually enforceable instead of aspirational. I would rather do that than keep a claim about code that never builds. Say the word if you would prefer it simply deleted instead.

**"14 sites" is really 7.** Miscounted in my own description; corrected.

**The `fake` list double-remove you flagged is real, and it is pre-existing rather than mine.** `dentry_remove()` unlinks from the hash chain and then puts the dentry on `fake` using the same `d_link` field (`vfs_dentry.cc:48`, `:181`), while `dentry_children_remove()` does `LIST_REMOVE(entry, d_link)` for every child. A child already on `fake` gets unlinked from `fake` by that loop, and the eventual `drele()` does `LIST_REMOVE(dp, d_link)` a second time, writing through stale pointers. I checked whether I had introduced it: `dentry_children_remove()` is byte-identical between `upstream/master` and this branch, and the `fake`/`d_link` sharing is on master too, so this is an existing defect your reading surfaced rather than a regression from this series.

I would rather not fold that fix into this PR, because it is a separate bug with a separate argument and this series is already carrying three commits plus a test. Two ways to go and I am happy either way: I file it as its own PR against master (the minimal fix looks like giving the fake list its own link field, so `d_link` membership stops being ambiguous, rather than teaching the loop to guess), or I add it here if you would rather see the dentry-cache work land as one unit. Which do you prefer?

For the record on the earlier points, the reachability argument, the observation attribution, the leaf-lock wording and the test header are all corrected in `c7ab5929e` with details in the comment above.

gburd:
Correction to my earlier comment: when I said those three claims were fixed, I had only fixed them in the commit messages and the test header. The same claims were still sitting in the source comments in `fs/vfs/vfs_dentry.cc`, which are lines this diff adds and which are what actually persists in the tree. That was a bad answer and you would have found it on your next look. Fixed now in `18c9cc675`, comments only, no code change:

- **"dentry_hash_lock is a leaf lock now"** at line 67, and the same claim again at line 190 in `dentry_move()`. Both replaced with what is actually true: it no longer calls into the vnode layer, which is the point of the ordering rule, but it does nest `dentry->d_lock` through `dentry_children_remove()`, and that nesting is consistent everywhere with `d_lock` never held while taking `dentry_hash_lock`.
- **The "all vCPUs halted in do_idle with an empty wakeup mask and >1000 threads parked" observation** was still in the AB-BA header block, which is the misattribution I had just told you belonged to the allocation commit. Removed from there.
- **"This fires on the unfixed drele()"** replaced with a note that `DEBUG_VFS` is not enabled in a default build, so the assertion only has effect when it is turned on deliberately.

I also applied your `physio()` comment note here, since you said it was worth applying repo-wide: the ordering block now describes what the rule prevents and what it takes to reach the deadlock (two dentries aliasing one vnode, which rename can produce), instead of narrating what the previous code did wrong.

One thing I have deliberately not touched yet: #1503 inherits the old versions of these commits and therefore still carries two of the stale comments. I would rather rebase it onto this branch once #1502 settles, so it picks the corrections up rather than me hand-resolving the six conflicts between your per-chain rewrite and these comment edits. Flagging it so it is on the record rather than quietly pending.

The two questions from my previous comment are still open whenever you get to them: whether the pre-existing `fake` list double-remove should be its own PR, and for #1508 whether to delete `physio()` outright.

PR #1504, opened 2026-09-10
x64: cover the psABI stack and pointer guards in the thread control block

gburd:
`thread_control_block` on x86-64 is three words:

```c
struct thread_control_block {
    thread_control_block* self;
    void* tls_base;
    unsigned long app_tcb;
};
```

The x86-64 psABI, however, places the `-fstack-protector` canary at `%fs:0x28` and the pointer-mangling guard at `%fs:0x30`. Both offsets fall past the end of that structure, and past the end of its allocation: `setup_tcb()` sizes the block as `total_tls_size + sizeof(*_tcb)` and places the TCB at the very end. So every canary check made by an application built with the stack protector is an out-of-bounds read of whatever the allocator happened to place next.

The values that come back are not random-looking, which is what makes the diagnosis clear. Reading `%fs:0x28` in a freshly created thread returns things like

```
0x802000006812d065
0x802000007592d065
0x802000007c32d065
```

that is, page-table entries: NX set, present, user, accessed, dirty. The read is landing in neighbouring allocator memory.

This has not been noticed because the kernel is built with `-fno-stack-protector`, so nothing in OSv itself consults those slots. It affects only applications, and only those compiled with the protector enabled, where the effect is a silently wrong read rather than a visible failure. `__stack_chk_guard` is defined in `runtime.cc`, but GCC does not read the global on this architecture, it reads `%fs:0x28`.

It turns into a hard failure as soon as one execution context resumes on a stack frame into which another spilled its canary, because the two then disagree and the epilogue check fails:

```
__stack_chk_fail(): Stack overflow detected. Aborting.
```

The message is misleading in that situation: there is no stack overflow, only two disagreeing canary values, one of them read out of bounds.

This change extends the block to cover `0x28` and `0x30` with named fields, adds `static_assert`s on both offsets and on the minimum size so the layout cannot drift out of agreement with the ABI, and initialises both guards in `setup_tcb()` as well as for the pre-scheduler TCB, so every stack-protected function runs under an initialised value rather than whatever memory follows.

Notes for review:

* The guard value is deliberately uniform rather than per-thread. A per-thread random canary is stronger in principle, but it is only sound if no execution context ever resumes on a frame another created; a uniform value is the conservative choice and is what makes the check consistent. This is a correctness fix, not a hardening claim.
* `aarch64` should be checked separately. Its TCB layout and the offset its toolchain uses for the canary are different, and I have not verified whether the same shortfall exists there.
* The compile-time asserts are the part worth keeping regardless of the rest: they turn any future divergence between this structure and the psABI into a build failure instead of an out-of-bounds read.

### DIFF
--- a/arch/x64/arch-setup.cc
+++ b/arch/x64/arch-setup.cc
@@ -226,6 +226,12 @@ void arch_setup_tls(void *tls, const elf::tls_data& info)
     memset(tls + info.filesize, 0, info.size - info.filesize);
     tcb = (struct thread_control_block *)(tls + info.size);
     tcb->self = tcb;
+    // tcb0 lives in .bss (see tcb0 in arch/x64/loader.ld) so it is already zero,
+    // but set the psABI guards explicitly: this is the TCB every stack-protected
+    // function runs under until the first sched::thread installs its own, and a
+    // zero canary here would disagree with the ones setup_tcb() installs later.
+    tcb->stack_guard = tcb_guard_value();
+    tcb->pointer_guard = tcb_guard_value();
     processor::wrmsr(msr::IA32_FS_BASE, reinterpret_cast<uint64_t>(tcb));
 }
 
--- a/arch/x64/arch-switch.hh
+++ b/arch/x64/arch-switch.hh
@@ -285,6 +285,16 @@ void thread::setup_tcb()
     _tcb->tls_base = p + user_tls_size;
 
     _tcb->app_tcb = 0;
+    _tcb->reserved0 = 0;
+    _tcb->reserved1 = 0;
+    // Install the psABI guards.  Applications compiled with -fstack-protector
+    // read the canary straight from %fs:0x28, so an uninitialised slot here is
+    // read as a canary by the very first stack-protected function the app calls.
+    // The value is process-wide on purpose: fork() children resume mid-frame on
+    // the parent's stack and must see the value the parent spilled (see
+    // arch-tls.hh and arch/x64/fork.cc).
+    _tcb->stack_guard = tcb_guard_value();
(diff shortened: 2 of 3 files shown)


gburd:
Answering the aarch64 question I left open in the description, so it is not a loose end for a reviewer.

**The defect does not apply to aarch64, for a structural reason rather than by luck.** There is no AArch64 equivalent of `%fs:0x28`. GCC's `-mstack-protector-guard` defaults to `global` on that target, so the canary is read from the `__stack_chk_guard` symbol, which this tree already defines in `runtime.cc`:

```c
void* __stack_chk_guard = reinterpret_cast<void*>(0x12345678abcdefull);
```

No `-mstack-protector-guard=` flag is passed anywhere in the build, so the default applies. For contrast, `arch/aarch64/arch-tls.hh`'s block is two words and does not need to be any longer:

```c
struct thread_control_block {
    void *tls_base; // Address of the per-thread static TLS block
    void *dtv;      // Address of the DTV (Dynamic Thread Vector)
};
```

There is no fixed ABI offset past those fields that the compiler reads, so nothing falls off the end. This is why the failure was only ever observed on x86_64.

Two consequences worth noting for review:

The change is therefore correctly confined to `arch/x64/`, and no matching aarch64 patch is needed or included.

It also means the canary on aarch64 is a fixed compile-time constant today. That is a separate observation rather than something this change addresses, and it is the same reason the value installed here is uniform rather than per-thread: OSv resumes execution contexts on stacks that other contexts created, and a per-thread canary is only sound if that never happens. Making the guard genuinely random, on either architecture, is a hardening change that should stand on its own and be argued separately from this correctness fix.

PR #1505, opened 2026-09-11
zfs: honour sub-tick timeouts in the SPL cv_timedwait_hires

gburd:
The OSv SPL's `cv_timedwait_hires()` converts its nanosecond deadline into `hz`-granular ticks and forwards it to `cv_timedwait()`. That is wrong in two independent ways, and the second one silently disables ZIL commit batching.

**The unit.** With `CALLOUT_FLAG_ABSOLUTE` the caller passes an absolute deadline, so the code correctly skips adding `gethrtime()`, then divides the absolute nanosecond timestamp by `1e6` and hands the quotient to `cv_timedwait()`, which treats its argument as a *relative* tick count. A deadline a quarter of a millisecond away becomes a relative wait of over twenty thousand days:

```
OLD, ABSOLUTE 250us window -> cv_timedwait(ticks=1789146780736)
     = 1789146780.7 seconds = 20707.7 DAYS
```

So the timeout never fires. `zil_commit_waiter()` depends on it firing to discharge what upstream's comment calls responsibility (2), issuing an lwb whose batch window has expired; instead that work waits for whichever other thread happens along.

**The resolution.** Even on the relative path, `hz` is 1000, so a tick is a millisecond and integer division truncates anything shorter to zero:

```
asks   10 us -> 0 ticks    asks  100 us -> 0 ticks
asks   25 us -> 0 ticks    asks  250 us -> 0 ticks
asks   50 us -> 0 ticks    asks  999 us -> 0 ticks
```

The ZIL sizes its commit-batch window at `zfs_commit_timeout_pct`, 10% of the last log-write latency (`zil.c`), which is tens to hundreds of microseconds. Every such window collapses to no wait at all, so a commit waiter never accumulates a second itx and **each commit pays its own log write and device cache flush instead of coalescing with its peers.**

The patch resolves the deadline against `gethrtime()` and waits the remaining nanoseconds via a new `osv_cv_timedwait_ns()`. Verified with the fixed body compiled standalone:

```
ABSOLUTE, 250us window -> requested 250000 ns
RELATIVE, 250us window -> requested 250000 ns
past deadline           -> rc -1
```

One subtlety worth flagging for review, because it is easy to get wrong: `now` must come from `gethrtime()`, the same clock the caller used to build the deadline. `gethrtime()` here resolves to `CLOCK_REALTIME` (`compat/opensolaris/sys/time.h` defines `CLOCK_UPTIME` to `CLOCK_REALTIME`), so subtracting a monotonic reading instead would reintroduce an error the size of the gap between the two clocks. An earlier version of this fix that I did not submit used a monotonic clock and was broken in the same direction as the bug it was meant to repair.

**On the performance side I want to be careful about what is and is not measured here.** Earlier measurement on this port showed `synchronous_commit=on` at 878 tps / 18.0 ms versus `off` at 5528 tps / 2.9 ms for the same workload, i.e. the per-commit synchronous flush accounted for essentially the whole gap to Linux on sync writes, and sync throughput barely scaled with concurrency (c1 364, c8 656, c32 845 tps) which is the signature of no batching. That is consistent with this defect and is what sent me looking here. **I have not yet re-run that benchmark with this patch applied**, so I am claiming the defect and the mechanism, not a specific speedup. Happy to post numbers once I have them on a matched base.

Applies cleanly to master (`git merge-tree` reports no conflicts).

### DIFF
--- a/bsd/porting/netport1.cc
+++ b/bsd/porting/netport1.cc
@@ -90,3 +90,19 @@ int openzfs_cv_timedwait(kcondvar_t *cv, mutex_t *mutex, clock_t abstime)
     auto ret = cv->wait(mutex, std::chrono::nanoseconds(ticks2ns(delta)));
     return ret == ETIMEDOUT ? -1 : 0;
 }
+
+// Nanosecond-granular timed wait, for OpenZFS callers whose delays are shorter
+// than a tick.  cv_timedwait_hires() in the OSv SPL header resolves its
+// absolute/relative deadline against gethrtime() and passes the remaining delay
+// here, so no clock conversion happens on this side.  Rounding these delays to
+// hz-granular ticks would truncate every sub-millisecond wait to zero, which
+// defeats ZIL commit-batch coalescing (see the comment at the call site).
+OSV_LIBSOLARIS_API
+int osv_cv_timedwait_ns(kcondvar_t *cv, mutex_t *mutex, long long delta_ns)
+{
+    if (delta_ns <= 0) {
+        return -1;
+    }
+    auto ret = cv->wait(mutex, std::chrono::nanoseconds(delta_ns));
+    return ret == ETIMEDOUT ? -1 : 0;
+}
--- a/bsd/sys/cddl/compat/opensolaris/sys/kcondvar.h
+++ b/bsd/sys/cddl/compat/opensolaris/sys/kcondvar.h
@@ -50,6 +50,9 @@ typedef enum {
 int cv_timedwait(kcondvar_t *cv, mutex_t *mutex, clock_t tmo);
 // OpenZFS variant: absolute deadline (see bsd/porting/netport1.cc).
 int openzfs_cv_timedwait(kcondvar_t *cv, mutex_t *mutex, clock_t abstime);
+// Nanosecond-granular relative wait, for delays shorter than a tick
+// (see cv_timedwait_hires() in the OSv SPL condvar header).
+int osv_cv_timedwait_ns(kcondvar_t *cv, mutex_t *mutex, long long delta_ns);
 
 #ifdef __cplusplus
 }
--- a/exported_symbols/osv_libsolaris.so.symbols
+++ b/exported_symbols/osv_libsolaris.so.symbols
@@ -9,6 +9,7 @@ copyinstr
 copyout
 cv_timedwait
 openzfs_cv_timedwait
+osv_cv_timedwait_ns
 debug
 destroy_bio
 device_close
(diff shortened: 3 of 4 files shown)


gburd:
A note on scope, to be careful about what this does and does not claim.

Master's `modules/open_zfs/osv/` SPL header still tick-converts the deadline, so master has this defect. This patch is correct against it and applies cleanly.

I want to be precise about performance, though. My integration branches already carry an equivalent fix (by a different route, in `netport1.cc` rather than the header), and on the one base where I have HammerDB numbers that fix was present and the workload was still well behind Linux at the same concurrency. So I am **not** claiming this patch moves throughput. What I am claiming is narrower and I am confident of it: the tick conversion is wrong on both the unit and the resolution axes, master has that wrong code, and this repairs it. The earlier synchronous-commit on/off measurement (6x) is what motivated looking here; it is consistent with this defect but does not prove this defect is the whole of it.

I am running a same-binary A/B now (a runtime switch toggling the fixed vs the old path on one image, dataset, and pool) to measure whether honouring the sub-millisecond ZIL window actually changes throughput. I will post the numbers here either way. If it turns out the window does not matter for this workload, the patch still stands on correctness grounds, but I would not want the commit message read as a performance claim I have not backed.

gburd:
A/B results, as promised. Same image, pool, dataset, and GUCs; the only difference between arms is a boot env var toggling the fixed nanosecond path against the old tick-truncating one. The console confirmed the flag bound with different values per arm, so this is a real A/B and not a dropped flag.

```
VU    fixed median   old median   delta
 8       48,758        47,144      +3.4%
32       37,012        36,957      +0.1%
64       33,864        32,910      +2.9%
```

Every delta is within the 3.0-3.6% run-to-run spread. **For this TPROC-C workload the fix does not change throughput.**

So I want to be explicit: this patch should be judged on correctness, not performance. Rounding a requested sub-millisecond deadline down to zero is wrong, and master does that; this repairs it. But I have no evidence it speeds up this workload, and I would rather say so than let the earlier synchronous-commit numbers imply a benefit I could not reproduce here.

The instrumentation also explains why the flag can't matter: at VU32 both arms leave the NVMe ~10-15% busy and the guest using ~13-15% of its vCPUs. Device idle and CPU idle while throughput is capped, which means the limit is a serialization point, not the ZIL batch-window timing. That is a separate investigation and not something this PR needs to resolve.

Happy to drop the patch if you would rather not carry a correctness-only fix here, but I think honouring the deadline the caller asked for is the right behaviour regardless.

PR #1508, opened 2026-09-13
vfs: issue one bio per iovec in physio(), with the correct length

gburd:
`physio()` builds one bio per loop iteration but set `bio_bcount` to `uio_resid` - the total remaining length of the whole uio - while `bio_data` points at a single iovec's base. When the uio carries more than one iovec, the bio describes `uio_resid` bytes starting at `iov[0].iov_base` even though only `iov[0].iov_len` bytes are valid there. A multi-iovec write therefore hands adjacent memory to the device as data.

A driver that validates each block server-side rejects the write; a plain disk silently writes garbage. This surfaces the moment ZFS submits a multi-iov write through `physio()`.

The fix sets `bio_bcount` to `iov->iov_len`, the length actually described by `bio_data`.

While here, a zero-length leading iovec was mishandled: the loop did `continue` without advancing `uio_iov`/`uio_iovcnt`, so it spun forever on that iovec. This advances past a zero-length iovec instead.

The sibling function earlier in the same file already iterates one bio per iovec with `iov->iov_len`; this brings `physio()` in line with it.

Provenance: found while bringing up a block driver that checksums each block (#1424), where the mismatch showed up immediately as a per-block hash-validation failure on multi-iov ZFS writes. It is a general `physio()` bug independent of that driver, so it is split out here to land on its own rather than being gated behind the driver PR.

Applies cleanly to master (`git merge-tree` reports no conflicts). Single file, +27/-7 (the code change is small; most of the addition is the explanatory comment and the tab/space cleanup of the loop tail).


fs/vfs/vfs_bdev.cc
```diff
@@ -188,12 +188,32 @@ physio(struct device *dev, struct uio *uio, int ioflags)
...
+	 *    uio carried more than one iovec the bio described uio_resid bytes
+	 *    starting at iov[0].iov_base while only iov[0].iov_len bytes are
+	 *    valid there, so a multi-iovec write sent adjacent stack/heap memory
+	 *    to the device as data.  A driver that checksums each block (for
+	 *    example an over-the-wire block protocol) rejects it; a plain disk
+	 *    silently writes garbage.  Use iov->iov_len.
+	 *
+	 *  - a zero-length leading iovec hit `continue` without advancing
+	 *    uio_iov/uio_iovcnt, so the loop spun forever on it.  Skip it
+	 *    properly instead.
```
nyh: Please ask your AI to avoid writing history lessons or changelogs in comments. The comment, if needed, should explain why some surprising code is correct - it shouldn't explain why one random mistake (which happens to be a mistake that existed in this code previously) is indeed a mistake. That's not only unhelpful, it's also confusing.

fs/vfs/vfs_bdev.cc
```diff
@@ -216,10 +236,10 @@ physio(struct device *dev, struct uio *uio, int ioflags)
...
 			return ret;
 
-	        uio->uio_iov++;
-        	uio->uio_iovcnt--;
-        	uio->uio_resid -= iov->iov_len;
-        	uio->uio_offset += iov->iov_len;
+		uio->uio_iov++;
+		uio->uio_iovcnt--;
+		uio->uio_resid -= iov->iov_len;
+		uio->uio_offset += iov->iov_len;
```
nyh: After the fix, physio() is character-for-character `bdev_strategy_rw()` with `bio_cmd` derived from `uio_rw.` So the whole function can be:
```c++
int
physio(struct device *dev, struct uio *uio, int ioflags)
{
	if (uio->uio_offset < 0)
		return EINVAL;
	if (uio->uio_resid == 0)
		return 0;

	return bdev_strategy_rw(dev, uio,
	    uio->uio_rw == UIO_READ ? BIO_READ : BIO_WRITE);
}
```

Moreover, using `bdev_strategy_rw()` will fix another bug - this version is missing a `uio_iovcnt > 0` check that bdev_strategy_rw has `uio->uio_resid > 0 && uio->uio_iovcnt > 0`; the PR keeps physio()'s bare uio_resid > 0. That matters because of the new continue: previously a zero-length iovec hung, now it advances, so if uio_resid and the iovec list ever disagree the loop walks off the end of the array and dereferences garbage. Trading a hang for an OOB read is a step sideways.

fs/vfs/vfs_bdev.cc
```diff
@@ -188,12 +188,32 @@ physio(struct device *dev, struct uio *uio, int ioflags)
...
+	 *
+	 *  - a zero-length leading iovec hit `continue` without advancing
+	 *    uio_iov/uio_iovcnt, so the loop spun forever on it.  Skip it
+	 *    properly instead.
+	 */
 	while (uio->uio_resid > 0) {
 		struct iovec *iov = uio->uio_iov;
 
-		if (!iov->iov_len)
+		if (!iov->iov_len) {
```
nyh: Pre-existing problem that my AI complained about:
No bounds check. bdev_read/bdev_write both reject uio_offset + uio_resid > dev->size; physio() doesn't. #1424 makes it a real device read/write entry point, so out-of-range I/O would go straight at the driver. Worth adding deliberately or noting why not.

nyh (changes requested):
I left some specific comments, but my AI also asked this, regarding whether we need this function at all:

#1424's comment says it "MUST go through physio() rather than bdev_read/bdev_write — those route through the OSv buf cache which is hard-coded to BSIZE=512". That's stale since his own c8c15304: bdev_read/bdev_write now take the bdev_strategy_rw fast path for any 512-aligned uio and never touch the buffer cache. For a 4096-block Crucible volume, a 512-aligned-but-not-4096-aligned request takes that fast path and gets the same EINVAL from the driver either way. So it's worth asking whether physio() needs to be revived at all, or whether the Crucible driver should just use bdev_read/bdev_write and physio() should be deleted — 13 years dead is a decent argument for removal rather than repair.

gburd:
Both points taken. Pushed as `1b9b5add8`; the change is now +2/-37 against master.

**The function collapses to `bdev_strategy_rw()`** - correct, and I should have seen it. `bdev_strategy_rw()` already walks the uio issuing one bio per iovec with `bio_bcount = iov->iov_len`, already skips a zero-length iovec while advancing `uio_iov`, and `bdev_read()`/`bdev_write()` already delegate to it. Both bugs I was fixing are simply absent there, so the fix is to call it rather than to repeat it:

```c
int
physio(struct device *dev, struct uio *uio, int ioflags)
{
	if (uio->uio_offset < 0)
		return EINVAL;
	if (uio->uio_resid == 0)
		return 0;

	return bdev_strategy_rw(dev, uio,
	    uio->uio_rw == UIO_READ ? BIO_READ : BIO_WRITE);
}
```

Its loop condition also carries `uio_iovcnt > 0`, which mine did not, so this is a small correctness gain on top of the deletion.

**The comment** - fair, and the criticism generalizes beyond this hunk. A comment explaining why the code that is no longer there was wrong is a changelog, and the commit message is where that belongs; a reader of the function does not need the archaeology, and it invites confusion about which behavior is current. The 18-line block is gone with the loop it described, and nothing replaced it because the delegating call needs no explanation. I have taken the general note, not just the instance.

gburd:
Third point addressed too, in `75bb43bf6` - I had missed it when I replied above, apologies.

You are right that `physio()` was the only one of the three entry points without the range check. `bdev_read()` and `bdev_write()` both do `uio_offset + uio_resid > dev->size -> EIO` first, so `physio()` now does the same, in the same order:

```c
	if (uio->uio_offset + uio->uio_resid > dev->size)
		return EIO;
	if (uio->uio_offset < 0)
		return EINVAL;
	if (uio->uio_resid == 0)
		return 0;
```

Adding it deliberately rather than noting why not, since as you say #1424 turns this into a real device I/O entry point and an out-of-range transfer would otherwise reach the driver.

Also worth recording that your `uio_iovcnt > 0` observation was the sharper half of the earlier point: my loop kept `physio()`'s bare `uio_resid > 0`, so once the zero-length iovec stopped hanging, a `uio_resid`/iovec-list disagreement would have walked off the end of the array instead. Delegating to `bdev_strategy_rw()` picks up its `uio_resid > 0 && uio_iovcnt > 0` condition, so that is fixed by construction rather than by me adding a second guard. Trading a hang for an out-of-bounds read would indeed have been sideways.

gburd:
On the larger question of whether `physio()` should exist at all: I checked, and you are right to ask.

Our 3-argument `physio(struct device *, struct uio *, int)` has **no in-tree caller**. The hits that look like callers are not:

- `bsd/.../zfs/zvol.c:1429` and `:1476` call the old BSD six-argument form (`physio(zvol_strategy, NULL, dev, B_READ, zvol_minphys, uio)`), which does not match our prototype - and that file is not in the build anyway. The `zvol.o` we compile is `$(OPENZFS)/module/zfs/zvol.o` from the OpenZFS tree, not this one.
- `vdev_disk_physio()` is a different function that happens to share the substring.

So the only consumer is the out-of-tree driver in #1424, which uses it in two lines, and the justification I gave in this PR is stale anyway now that `c8c153047` reworked the device I/O paths.

Given that, I think deleting it is the better change and I am happy to do that instead: drop `physio()` and its declaration in `include/osv/device.h`, and have #1424's driver call `bdev_strategy_rw()` directly, which is what my "fix" ended up delegating to. That removes a public entry point nothing calls rather than carefully repairing one, and it means the OOB-read hazard you identified cannot be reintroduced by a future caller.

What is on this branch now is the conservative version: `physio()` delegates to `bdev_strategy_rw()` and picks up its `uio_iovcnt > 0` guard, plus the range check to match `bdev_read()`/`bdev_write()`. That is correct as far as it goes, but it is repairing something with no users.

Tell me which you want and I will push it: delete the function, or keep the minimal repair. I lean toward deletion for the reason above.

PR #1509, opened 2026-09-14
virtio-blk: optional per-queue completion threads (OSV_BLK_MQ_COMPLETE)

gburd:
The virtio-blk driver used a single completion thread that drained every virtqueue's used ring and ran the whole bio completion inline (`biodone`->`vdev_disk_bio_done`->`zio_interrupt`->the ZIL `cv_broadcast` that wakes the waiting writers). On a multiqueue device that funnels every queue's completion through one thread, and under concurrent durable I/O that thread becomes the ceiling: completions queue behind it while the device and the other CPUs sit idle.

## Measurement

A concurrent 8k `pwrite`+`fdatasync` microbenchmark on ZFS-on-NVMe, in a 32-vCPU guest, one writer thread per concurrency level (each writing its own file), aggregate throughput:

| concurrent writers | single thread (default) | per-queue (`OSV_BLK_MQ_COMPLETE=1`) | delta |
|---:|---:|---:|---:|
| 8  | 21,302 op/s | 24,024 op/s | +13% |
| 16 | 21,749 op/s | 34,444 op/s | **+58%** |
| 32 | 27,772 op/s | ~41,000 op/s | **+48%** |

With the single thread, throughput flatlines from 8 to 16 writers (the completion thread saturates) while the NVMe is only ~40% busy and most vCPUs idle. With one completion thread per virtqueue the NVMe reaches ~75% busy and per-op p50 roughly halves (1026->648 us at 32 writers). This was found while chasing why PostgreSQL-on-ZFS does not scale past a handful of connections on OSv; the single completion thread is the WAL-commit-completion serialization point.

## The change

`OSV_BLK_MQ_COMPLETE=1` spawns one completion thread per virtqueue, each woken by its own queue's MSI-X vector and draining only its own queue. A single-queue guest behaves identically to before, so it is off by default and enabled per boot (the resolved mode and queue count print once at boot).

The per-queue drain deliberately does **not** take the per-queue submission lock, matching the existing single-thread `req_done()`: `make_request()` holds that lock across a sleeping `add_buf_wait()` when the ring is full, so a completion thread that took it would deadlock against a producer waiting on that same thread to free ring space. Because each queue is drained by exactly one thread, the drain races the producer only on the used-ring head index, which is lock-free.

Generic: parallel durable-write completion helps any fsync-heavy workload on a multiqueue virtio-blk device, not just PostgreSQL. Applies cleanly to master (`git merge-tree` reports no conflicts); touches only `drivers/virtio-blk.{cc,hh}`.


gburd:
End-to-end result, for honest scope. I ran a same-image A/B under HammerDB TPROC-C (200 warehouses) toggling this fix (single completion thread vs per-queue), 32 queues on the pool disk both arms, WAL on that device, matched config.

The fix does **not** move steady-state TPROC-C throughput: 67,055->70,152 NOPM at 16 VU (+4.6%, within run-to-run noise) and it still declines to 61,361 at 32 VU. At settled load the two arms are near-identical (about 12 of 32 vCPUs busy, NVMe ~25% utilized, ~74% idle in both)-the per-queue completion threads have nothing extra to parallelize because that workload never fills the device queues.

Where the fix clearly works is the condition the microbenchmark exercised and this workload does not: during cold/backlog drain the per-queue arm drove the NVMe to 56-96% util at aqu-sz 5 / ~580 MB/s, while the single-thread arm never exceeded 28% / aqu-sz 1.0. So the completion path genuinely parallelizes and drives the device 2-3x harder when the queues are actually full; a steady-state fsync-per-commit database workload just does not fill them.

So I am **not** claiming a database-throughput win here. The fix stands on the microbenchmark (+58% aggregate durable-write throughput at 16 concurrent writers, device 40->75% busy) and on correctness (parallel completion with the drain kept lock-free so it cannot deadlock against a producer sleeping in add_buf_wait). It is a generic improvement to the completion path for any workload that keeps the block queues full; it is not the fix for the steady-state ceiling of a per-commit-fsync workload, which sits upstream in the post-commit wakeup path.

One implementation note confirmed under load: the lock-free `req_done_q` drain held-no run wedged with zero runnable threads across the whole A/B.

gburd:
Flipped this to enabled-by-default, with a cleaner measurement than the earlier HammerDB null.

A matched durable-write A/B (concurrent 8k pwrite+fdatasync on ZFS-on-NVMe, 32-vCPU guest, virtio-blk with 4 queues, single completion thread vs per-queue), measuring the ZIL log-flush rate directly:

```
thr  single op/s  single flush/s  per-queue op/s  per-queue flush/s  gain
  8       11,652         3,222         12,763            3,495        +10%
 16       15,992         3,001         21,996            4,291        +38% op/s, +43% flush/s
 32       24,079         2,579         32,097            3,621        +33% op/s, +40% flush/s
```

p99 roughly halved with per-queue (5828->2593 us at t32). So the single completion thread was a real serialization on the flush-completion path, and per-queue completion lifts it. This is the direct-measurement version of what the earlier HammerDB A/B could not see: TPROC-C at 200 warehouses did not push the flush-completion path to saturation, so the switch read as a null there, but a durable-write workload that does saturate it shows a clear win.

Honest scope, unchanged: this narrows but does not close the gap to Linux (per-queue 3621 flush/s vs Linux 7077 on the same device at t8). A second, larger constraint remains in the ZIL block-WRITE round-trip itself (about 270us of the 380us per-lwb latency, versus a plain 8k write completing in ~6us on the same device), which is a separate follow-up. But per-queue completion is a real, generic, measured win on the completion path and there is no reason to keep it behind an env var: a single-queue guest is unaffected, and the only cost otherwise is one completion thread per queue. `OSV_BLK_MQ_COMPLETE=0` remains as an off switch.

`git merge-tree` clean against master.

gburd:
## Relationship to #1466, and a correction to how I framed this PR

@wkozaczuk's #1466 changes virtio-blk to one completion thread per virtqueue **as the default**, and reverts the single-completion-thread machinery this PR builds on. I had recorded that internally as a design collision. Having now read his measurements properly, **that framing was too simple, and part of it was wrong.**

His fio numbers (means of his own table) are **0.83x read / 0.80x write** for 4 queues + 4 completion threads versus 1 queue + 1 thread - fan-out is clearly *worse* on his benchmark. His 4q/4ct write runs also spread 343 / 505 / 365 MiB/s (**1.47x**, with an 85.7-126 MiB/s per-job range inside a single run) against 1.04x for single-queue, which looks like load imbalance across completion threads and is a p99 concern on its own.

I do not think that contradicts the +58% in this PR. The workloads differ in the way that matters:

- **His:** `fio`, streaming read/write on a raw block device, no `fsync`, deep queues. One completion thread **batches many completions per wakeup**; splitting it adds IPIs and cross-CPU traffic for no gain.
- **This PR:** `pwrite` + **`fdatasync` per op**, one outstanding completion per writer, **every writer sleeping on it**. Nothing to batch, so the single thread's wakeup latency is on the critical path of every commit.

**The honest general statement is narrower than what this PR's description implies:** completion-thread fan-out is a *latency* optimisation for blocking-durable-write workloads, and a *throughput regression* for deep-queue streaming ones. I described this PR as "helps any fsync-heavy workload on a multiqueue device", which is defensible, but I should also have said that it **hurts** streaming multiqueue workloads - his data shows that clearly and mine did not cover it. That asymmetry is the reason the opt-in default in this PR is the right shape, and I should have arrived at it from evidence rather than caution.

Two further things worth recording here:

1. **His `run.py` fix is orthogonal and matters to everyone's measurements.** `--virtio-blk-queues` advertises `default: 1`, but the old code only passed `num-queues` to QEMU when the value was >1, so at the default QEMU chose its own count - and OSv's driver takes its queue count from the probed device config (`virtio-blk.cc:146`), not from the flag. Any multiqueue virtio-blk measurement taken through `run.py` at the default had an uncontrolled queue count. **That includes potentially some of ours.**
2. **I have not run his benchmark against this PR.** These numbers are from a QEMU/KVM guest on ZFS-on-NVMe with an fsync-per-op microbenchmark; I have not reproduced his fio streaming configuration. Until I do, the claim here should be read as scoped to blocking durable writes.

I have asked him how he wants to resolve it rather than assuming - including offering to rebase this PR on top of his default change as a pure opt-in addition, which I think may be the cleanest outcome for both.

